PR #118, opened 2015-01-30
Encoded Redirect URLs

fhanik:
First proposed solution to https://www.pivotaltracker.com/story/show/87310482

### DIFF
--- a/common/src/main/java/org/cloudfoundry/identity/uaa/authorization/UaaAuthorizationEndpoint.java
+++ b/common/src/main/java/org/cloudfoundry/identity/uaa/authorization/UaaAuthorizationEndpoint.java
@@ -13,6 +13,8 @@
 
 package org.cloudfoundry.identity.uaa.authorization;
 
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
 import java.security.Principal;
 import java.util.Collections;
 import java.util.Date;
@@ -72,6 +74,7 @@ import org.springframework.web.servlet.View;
 import org.springframework.web.servlet.view.RedirectView;
 import org.springframework.web.util.UriComponentsBuilder;
 import org.springframework.web.util.UriTemplate;
+import org.springframework.web.util.UriUtils;
 
 /**
  * Authorization endpoint that returns id_token's if requested.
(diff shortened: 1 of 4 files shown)


common/src/main/java/org/cloudfoundry/identity/uaa/authorization/UaaAuthorizationEndpoint.java
```diff
@@ -399,14 +409,16 @@ private String getSuccessfulRedirect(AuthorizationRequest authorizationRequest,
         }
 
         UriComponentsBuilder template = UriComponentsBuilder.fromUriString(authorizationRequest.getRedirectUri());
-        template.queryParam("code", authorizationCode);
+        //decode already encoded values
+        template = UriComponentsBuilder.fromHttpUrl(template.build(true).toUriString());
+        template.queryParam("code", encode(authorizationCode));
```
fhanik: Here we are proceeding to assume that the input URL is always encoded. Thus using the UriComponentsBuilder, we can only add already encoded parameters. Each parameter that is added, must be encoded before it is added.

common/src/main/java/org/cloudfoundry/identity/uaa/authorization/UaaAuthorizationEndpoint.java
```diff
@@ -399,14 +409,16 @@ private String getSuccessfulRedirect(AuthorizationRequest authorizationRequest,
...
 
         String state = authorizationRequest.getState();
         if (state != null) {
-            template.queryParam("state", state);
+            template.queryParam("state", encode(state));
         }
 
-        return template.build().encode().toUriString();
+        return template.build(true).toUriString();
```
fhanik: Since we assume the URI to be encoded, do NOT encode again.

common/src/main/java/org/cloudfoundry/identity/uaa/authorization/UaaAuthorizationEndpoint.java
```diff
@@ -418,24 +430,20 @@ private String getUnsuccessfulRedirect(AuthorizationRequest authorizationRequest
         }
 
         UriComponentsBuilder template = UriComponentsBuilder.fromUriString(authorizationRequest.getRedirectUri());
-        Map<String, String> query = new LinkedHashMap<String, String>();
```
fhanik: Based on the assumption that the URI is encoded, we can't expand variables. In order to expand variables in UriComponentsBuilder, the URI must be decoded.

common/src/main/java/org/cloudfoundry/identity/uaa/authorization/UaaAuthorizationEndpoint.java
```diff
@@ -418,24 +430,20 @@ private String getUnsuccessfulRedirect(AuthorizationRequest authorizationRequest
...
+        template = UriComponentsBuilder.fromHttpUrl(template.build(true).toUriString());
         StringBuilder values = new StringBuilder();
 
-        values.append("error={error}");
-        query.put("error", failure.getOAuth2ErrorCode());
-
-        values.append("&error_description={error_description}");
-        query.put("error_description", failure.getMessage());
+        values.append("error="+encode(failure.getOAuth2ErrorCode()));
+        values.append("&error_description="+encode(failure.getMessage()));
```
fhanik: Each variable that we add, we must encode, here we know that the parameter names do not need encoding

common/src/main/java/org/cloudfoundry/identity/uaa/authorization/UaaAuthorizationEndpoint.java
```diff
@@ -418,24 +430,20 @@ private String getUnsuccessfulRedirect(AuthorizationRequest authorizationRequest
...
-            values.append("&state={state}");
-            query.put("state", authorizationRequest.getState());
+            values.append("&state="+encode(authorizationRequest.getState()));
         }
 
         if (failure.getAdditionalInformation() != null) {
             for (Map.Entry<String, String> additionalInfo : failure.getAdditionalInformation().entrySet()) {
-                values.append("&" + additionalInfo.getKey() + "={" + additionalInfo.getKey() + "}");
-                query.put(additionalInfo.getKey(), additionalInfo.getValue());
+                values.append("&" + encode(additionalInfo.getKey()) + "=" + encode(additionalInfo.getValue()));
```
fhanik: we must ensure that even the parameter name is encoded

common/src/main/java/org/cloudfoundry/identity/uaa/authorization/UaaAuthorizationEndpoint.java
```diff
@@ -445,7 +453,7 @@ private String getUnsuccessfulRedirect(AuthorizationRequest authorizationRequest
             template.query(values.toString());
         }
 
-        return template.build().expand(query).encode().toUriString();
+        return template.build(true).toUriString();
```
fhanik: and we no longer expand variables, everything that the UriComponentsBuilder is already encoded

fhanik:
@fraenkel Thanks for your feedback. We're in a bit of a pickle here, cause we are dealing with contradictions. The Oauth2 specification demands that query parameters be encoded using application/x-www-form-encoded. The URI spec itself, defines plus sign as a delimiter, while the new HTTP spec RFC 7230 (that came after Oauth 2) points back to RFC 3896 URI spec for exact syntax. So they are contradicting each other.

It's all about timing of writing specs :) So in theory, we couldn't use UriComponentsBuilder as encoder because of OAuth 2 specification. But if we now started using form encoding like we are supposed to (form encoded in the URI) I'm sure all hell would break lose. I think for now, we'll stick with what we are doing, and just solve the double encoding issue.

PR #119, opened 2015-01-30
Encoded Redirect URLs

fhanik:
Alternative proposed solution to https://www.pivotaltracker.com/story/show/87310482

### DIFF
--- a/common/src/main/java/org/cloudfoundry/identity/uaa/authorization/UaaAuthorizationEndpoint.java
+++ b/common/src/main/java/org/cloudfoundry/identity/uaa/authorization/UaaAuthorizationEndpoint.java
@@ -13,6 +13,9 @@
 
 package org.cloudfoundry.identity.uaa.authorization;
 
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+import java.net.URI;
 import java.security.Principal;
 import java.util.Collections;
 import java.util.Date;
@@ -22,6 +25,7 @@ import java.util.Map;
 import java.util.Set;
 
 import org.cloudfoundry.identity.uaa.oauth.token.OpenIdToken;
+import org.cloudfoundry.identity.uaa.util.UaaUrlUtils;
 import org.springframework.http.ResponseEntity;
 import org.springframework.security.access.AccessDeniedException;
 import org.springframework.security.authentication.InsufficientAuthenticationException;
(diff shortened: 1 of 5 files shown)


common/src/main/java/org/cloudfoundry/identity/uaa/authorization/UaaAuthorizationEndpoint.java
```diff
@@ -392,13 +402,15 @@ private String generateCode(AuthorizationRequest authorizationRequest, Authentic
...
+
+
     private String getSuccessfulRedirect(AuthorizationRequest authorizationRequest, String authorizationCode) {
 
         if (authorizationCode == null) {
             throw new IllegalStateException("No authorization code found in the current request scope.");
         }
 
-        UriComponentsBuilder template = UriComponentsBuilder.fromUriString(authorizationRequest.getRedirectUri());
+        UriComponentsBuilder template = uaaUrlUtils.parseAndDecodeUrl(authorizationRequest.getRedirectUri());
```
fhanik: We assume that the URI is encoded, so we properly decode it
this allows us to create a UriComponentsBuilder that can handle variable expansion

common/src/main/java/org/cloudfoundry/identity/uaa/authorization/UaaAuthorizationEndpoint.java
```diff
@@ -417,8 +429,8 @@ private String getUnsuccessfulRedirect(AuthorizationRequest authorizationRequest
             throw new UnapprovedClientAuthenticationException("Authorization failure, and no redirect URI.", failure);
         }
 
-        UriComponentsBuilder template = UriComponentsBuilder.fromUriString(authorizationRequest.getRedirectUri());
-        Map<String, String> query = new LinkedHashMap<String, String>();
+        UriComponentsBuilder template = uaaUrlUtils.parseAndDecodeUrl(authorizationRequest.getRedirectUri());
+        Map<String, String> query = new LinkedHashMap<>();
```
fhanik: Very little logic changes here. However, testing of this becomes a bit more complex. See below

common/src/test/java/org/cloudfoundry/identity/uaa/util/UaaUrlUtilsTest.java
```diff
@@ -35,6 +39,74 @@ public void tearDown() throws Exception {
...
+        uriEquals(new URI(url), new URI(template.build().encode().toUriString()));
+        url += "&param1="+extraValueEncoded;
+        template.queryParam("param1", extraValueDecoded);
+        uriEquals(new URI(url), new URI(template.build().encode().toUriString()));
+        url += "&param2="+extraValueEncoded;
+        template.queryParam("param2", "{param2}");
+        uriEquals(new URI(url), new URI(template.build().expand(extraValueDecoded).encode().toUriString()));
+    }
+
+    public static void uriEquals(URI uri1, URI uri2) {
```
fhanik: the Spring UriComponentsBuilder uses a very strict encoding scheme. It goes by the letter for the HTTP specification of what is allowed for different components of the URI.

However, java.net.URLEncoder is a superset of the UriComponentsBuilder encoding, thus, after we encode the string using UriComponentsBuilder, it no longer looks the same. Comparison can only be done if we use decoded values.

PR #130, opened 2015-03-05
Reinstate original empty scopes error message

mthx:
It looks like the empty scope message was unintentionally mangled as part of a rename aimed at code - this PR reinstates the original message.  It's certainly quite confusing to encouter the current one!

I've also extended the existing test to cover the text of the message.


mthx:
Corporate CLA for CoreFiling provided today via @prcjak.

fhanik:
Thank you very much. I believe we need to record your github usernames so that PRs are automatically tagged with having a CLA on file.

PR #135, opened 2015-03-19
Update UAA-Tokens.md

krishna-mk:
The link provided is expired "https://developers.google.com/wallet/digital/docs/jwtdecoder". No more helping the user.

I had to go through other ways to (http://jwt.calebb.net/) see the token body. This is a very simple method to find the body of the token.

Also very detailed writings are available here dealing with the body of the token. https://developers.google.com/accounts/docs/OpenIDConnect#validatinganidtoken


krishna-mk:
I have made my profile public and so please accept the request. Thanks.

PR #143, opened 2015-04-07
Handle UsernameNotFoundException thrown from UaaUserDatabase

prcjac:
If a token is supplied that contains a GUID for a user that isn't
present in the database, a UsernameNotFoundException is thrown. The
exception isn't handled by the calling method and since it's runtime it
propogates, triggering a HttpClientErrorException in any Spring security
app using UAA as the OAuth2 provider.
Instead the user lookup is surrounded by a try catch which throws an
appropriate InvalidTokenException, if the user isn't found, which is
handled correctly.


fhanik:
Missing test case for new code path. Will add before merging.

fhanik:
Tracked via 63d30446bbc8a21adb6b7142401e0b421be21f9b

PR #154, opened 2015-05-29
Update build version number in README.md

finitud:
@anareis


fhanik:
Thanks, can you change this to be against develop branch, or it will cause problems during our automatic merges into master

fhanik:
Fixed as part of a documentation update

PR #167, opened 2015-06-19
configurable_automatic_creation_of_SAML_shadow_users

amiri-ge:
This modification allows the administrator to configure their SAML identity provider so that shadow users are not automatically created after the user successfully authenticates with the SAML identity provider. Instead, the application that initiated the authorization code grant flow will receive an OAuth error redirect as per https://tools.ietf.org/html/rfc6749#section-4.1.2.1 indicating that the access is denied because a shadow user does not explicitly exist on the OAuth server (UAA).

For background, I discussed this feature with Sree Tummidi a few weeks ago and went over our use case for this behavior. Essentially, we want UAA to signal to the app that a shadow user has not been explicitly provisioned. The app will then put the user through a registration flow. Once the user is successfully registered the app will provision the shadow account for the user and re-initiate the OAuth authorization code grant. Since the browser's session with the SAML server should still be valid, the user is not prompted to enter their credentials a second time. Instead, the UAA server determines that the user shadow account does exist and redirects back to the app with the authorization code and the use is successfully logged in.

Here are the benefits of this approach:
1. It requires minimal changes to the existing UAA code since it leverages Spring failure handlers for SAML SSO.
2. It allows applications to defer provisioning of shadow users until they have performed the required registration steps which keeps resource usage low.
3. Once user completes registration they do not have to enter their credentials again thanks to the session they already established with the SAML server.
4.  It avoids coupling UAA with any registration logic/workflow.

Usage: the SAML idp configuration has an extra attribute "addNew" which is true by default to preserve the original behavior but can be set to false to enable the behavior described above.

I've added an integration test for this as well as verified that all existing integration tests pass with only one ignored because it was marked by annotation to be ignored.


sreetummidi:
Hi Dario,
We will have this merged in the next release cycle.

Thanks,
Sree

fhanik:
Can you clean up this PR against a fresh pull from the develop branch. That way we don't get all the junk commits in it. I have a comment in the code too.

common/src/main/java/org/cloudfoundry/identity/uaa/login/saml/IdentityProviderDefinition.java
```diff
@@ -40,6 +40,7 @@
     private String socketFactoryClassName;
     private String linkText;
     private String iconUrl;
+    private boolean addNew = true;
```
fhanik: A more descriptive name would help here. a name like addShadowAccountUponSamlLogin ensures there is no confusion.

fhanik:
Can we also split out the Gradle changes into a separate PR. I'm not sure how they relate to the feature 'add new shadow account for saml login'

fhanik:
Superseded by https://github.com/cloudfoundry/uaa/pull/171

PR #169, opened 2015-06-22
Support customization of Login and Password prompts

youngm:
In our organization we don't use Email we use ldap integration so we would prefer to prompt for a "Username".

This PR provides initial support for getting username and password prompts from configuration keeping the default the same as the previous value.  If this PR gets accepted I can submit one to support prompt configuration in cf-release.

I find it odd nobody has done this already.  So, if I'm missing something a pointer in the correct direction would be great.

### DIFF
--- a/uaa/src/main/webapp/WEB-INF/spring-servlet.xml
+++ b/uaa/src/main/webapp/WEB-INF/spring-servlet.xml
@@ -249,12 +249,12 @@
             <util:list>
                 <bean class="org.cloudfoundry.identity.uaa.authentication.login.Prompt">
                     <constructor-arg name="name" value="username" />
-                    <constructor-arg name="text" value="Email" />
+                    <constructor-arg name="text" value="${login.prompt.username.text:Email}" />
                     <constructor-arg name="type" value="text" />
                 </bean>
                 <bean class="org.cloudfoundry.identity.uaa.authentication.login.Prompt">
                     <constructor-arg name="name" value="password" />
-                    <constructor-arg name="text" value="Password" />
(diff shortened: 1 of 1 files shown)


sreetummidi:
Hi Mike,
We will have this merged for the next UAA release cycle.

Sree

fhanik:
I would suggest

    --- a/uaa/src/main/webapp/WEB-INF/spring-servlet.xml
    +++ b/uaa/src/main/webapp/WEB-INF/spring-servlet.xml
    @@ -249,12 +249,12 @@
                 <util:list>
                     <bean class="org.cloudfoundry.identity.uaa.authentication.login.Prompt">
                         <constructor-arg name="name" value="username" />
    -                    <constructor-arg name="text" value="Email" />
    +                    <constructor-arg name="text" value="${login.prompt.username.text:Email}" />
                         <constructor-arg name="type" value="text" />
                     </bean>
                     <bean class="org.cloudfoundry.identity.uaa.authentication.login.Prompt">
                         <constructor-arg name="name" value="password" />
    -                    <constructor-arg name="text" value="Password" />
    +                    <constructor-arg name="text" value="${login.prompt.password.text:Password}" />
                         <constructor-arg name="type" value="password" />
                     </bean>
                     <bean class="org.cloudfoundry.identity.uaa.authentication.login.Prompt">

This construct will let us convert into a structure like

    login:
      prompt:
        username:
          type: text
          text: username
        password:
          type: password
          text: Password

In the configuration, and dynamically generate these prompts

youngm:
@fhanik Sounds good.  I'll update the branch.

youngm:
@fhanik PR updated as requested.

PR #171, opened 2015-06-25
Modified SAML identity provider configuration to accept an addShadowU…

amiri-ge:
…serOnLogin property, which is true by default, and determines whether to add new shadow users automatically or not.

@fhanik I believe this pull request addresses your concerns.

### DIFF
--- a/uaa/src/main/webapp/WEB-INF/spring/saml-providers.xml
+++ b/uaa/src/main/webapp/WEB-INF/spring/saml-providers.xml
@@ -59,6 +59,7 @@
             <property name="providerDefinitionList">
                 <bean factory-bean="metaDataProviders" factory-method="getIdentityProviderDefinitions"/>
             </property>
+            <property name="contextProvider" ref="contextProvider" />
         </bean>
 
         <bean id="exceptionTranslationFilter" class="org.springframework.security.web.access.ExceptionTranslationFilter">
@@ -66,7 +67,9 @@
         </bean>
 
         <!-- IDP Discovery Service -->
-        <bean id="samlIDPDiscovery" class="org.cloudfoundry.identity.uaa.login.saml.LoginSamlDiscovery"/>
+        <bean id="samlIDPDiscovery" class="org.cloudfoundry.identity.uaa.login.saml.LoginSamlDiscovery">
+            <property name="contextProvider" ref="contextProvider" />
+        </bean>
 
         <bean id="samlEntityID" class="java.lang.String">
             <constructor-arg value="${login.entityID:unit-test-sp}"/>
@@ -137,6 +140,8 @@
 
         <bean id="samlUserDetails" class="org.cloudfoundry.identity.uaa.login.LoginServerSamlUserDetailsService" />
 
+        <bean id="samlLoginFailureHandler" class="org.cloudfoundry.identity.uaa.login.saml.LoginSAMLAuthenticationFailureHandler" />
+
         <!-- Provider of default SAML Context -->
         <bean id="contextProvider" class="org.springframework.security.saml.context.SAMLContextProviderImpl" />
 
@@ -144,6 +149,8 @@
         <bean id="samlWebSSOProcessingFilter" class="org.springframework.security.saml.SAMLProcessingFilter">
             <property name="authenticationManager" ref="authenticationManager" />
             <property name="authenticationSuccessHandler" ref="successRedirectHandler" />
+            <property name="authenticationFailureHandler" ref="samlLoginFailureHandler" />
+            <property name="contextProvider" ref="contextProvider" />
         </bean>
 
         <!-- Logout handler terminating local session -->
--- a/common/src/main/java/org/cloudfoundry/identity/uaa/login/saml/IdentityProviderDefinition.java
+++ b/common/src/main/java/org/cloudfoundry/identity/uaa/login/saml/IdentityProviderDefinition.java
@@ -40,6 +40,7 @@ public class IdentityProviderDefinition {
     private String socketFactoryClassName;
     private String linkText;
     private String iconUrl;
+    private boolean addShadowUserOnLogin = true;
 
     public IdentityProviderDefinition() {}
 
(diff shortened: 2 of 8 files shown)


fhanik:
@amiri-ge - please make sure that the build passes. You can run it locally as

    ./gradlew test integrationTest

and that ensures it works with HSQLDB. Travis additionally tests with MySQL and PostgreSQL

amiri-ge:
@fhanik I tested it locally with "gradle integrationTest" before I performed the pull request and all tests passed fine. Is there a way I can get a report on what tests failed?

fhanik:
https://travis-ci.org/cloudfoundry/uaa/jobs/68248689#L4880

fhanik:
Normally our builds publish a full error report to a hosted system on amazon, I'm not sure why it didn't do that here. https://github.com/cloudfoundry/uaa/blob/develop/.travis.yml#L73-L74

May it doesn't do it for pull requests?

amiri-ge:
@fhanik OK, well at least I know which test failed. I'll see if I can reproduce this. Thanks.

amiri-ge:
I could not reproduce this locally. I tried running the integration tests as both "gradle test integrationTest" and "./gradlew -Dspring.profiles.active=default cobertura" as indicated by https://travis-ci.org/cloudfoundry/uaa/builds/68248685.

With the information that I have, the best I can do is make an educated guess about why it is failing and add more assertions to the test.

fhanik:
I have restarted the build and see if it was just a build anomaly

amiri-ge:
@fhanik Is there a way to turn on logging for PhantomJSDriver? If I could collect that in conjunction with the server side logs generated during the build I might be able to figure out what is going on.

amiri-ge:
I turned on debug for PhantomJS. If there's some way I could get the results/logs for the integration tests it would really help me out.

fhanik:
I'll pull it into a branch and run it, that way it will publish the reports.

fhanik:
https://github.com/cloudfoundry/uaa/tree/feature/saml_add_shadow_account_flag

fhanik:
getCurrentUrl is the last working URL on linux - so it returns the login page of the SAML IDP, because it gets a Connection Refused error

```
[DEBUG - 2015-06-29T18:30:12.641Z] Session [ddcbe4d0-1e8c-11e5-a9c1-83991ae4457c] - page.onResourceRequested - {"headers":[{"name":"Origin","value":"http://simplesamlphp.cfapps.io"},{"name":"User-Agent","value":"Mozilla/5.0 (Unknown; Linux x86_64) AppleWebKit/534.34 (KHTML, like Gecko) PhantomJS/1.9.7 Safari/534.34"},{"name":"Accept","value":"text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8"},{"name":"Referer","value":"http://simplesamlphp.cfapps.io/module.php/core/loginuserpass.php?"}],"id":38,"method":"GET","time":"2015-06-29T18:30:12.641Z","url":"http://localhost/app?error=access_denied&error_description=SAML+user+does+not+exist.+You+can+correct+this+by+creating+a+shadow+user+for+the+SAML+user."}
[DEBUG - 2015-06-29T18:30:12.642Z] Session [ddcbe4d0-1e8c-11e5-a9c1-83991ae4457c] - page.onResourceError - {"errorCode":1,"errorString":"Connection refused","id":38,"url":"http://localhost/app?error=access_denied&error_description=SAML+user+does+not+exist.+You+can+correct+this+by+creating+a+shadow+user+for+the+SAML+user."}
[DEBUG - 2015-06-29T18:30:12.642Z] Session [ddcbe4d0-1e8c-11e5-a9c1-83991ae4457c] - page.onResourceReceived - {"contentType":null,"headers":[],"id":38,"redirectURL":null,"stage":"end","status":null,"statusText":null,"time":"2015-06-29T18:30:12.642Z","url":"http://localhost/app?error=access_denied&error_description=SAML+user+does+not+exist.+You+can+correct+this+by+creating+a+shadow+user+for+the+SAML+user."}
```

fhanik:
Should the error redirect you away to another app, instead of staying on the UAA?

fhanik:
deleteUser also must handle the use case where the user doesn't exist. What if I run this test first?

amiri-ge:
@fhanik So to summarize:
1. On my machine, instead of getting a connection refused I get a 404 response back from the server which at least allows me to check the current URL. I can think of two ways to solve this: 1. pick a better redirectURI (based on this.baseURI) 2. Figure out a way to not have WebDriver follow redirects automatically (seems unlikely) 3. Use a different web framework (risky and possibly alot of work). I'll have to think about this carefully. Any suggestions?

2. Regarding deleteUser, originally I was handling that use case. I modified that aspect of the test to determine if the issue was that the user was not being deleted after it was created by previous tests. I will revert back to it's original form.

amiri-ge:
I've checked in a change that makes the test pass but it's sort of a hack. A better solution would be to create an additional sample app to redirect to. Basically, I just wanted to verify what the issue was before working on building a sample app to support the integration test.

uaa/src/main/webapp/WEB-INF/spring/saml-providers.xml
```diff
@@ -144,6 +149,8 @@
         <bean id="samlWebSSOProcessingFilter" class="org.springframework.security.saml.SAMLProcessingFilter">
             <property name="authenticationManager" ref="authenticationManager" />
             <property name="authenticationSuccessHandler" ref="successRedirectHandler" />
+            <property name="authenticationFailureHandler" ref="samlLoginFailureHandler" />
+            <property name="contextProvider" ref="contextProvider" />
```
fhanik: Thanks for adding this, you may notice that it gets set, even without it, because of the @Autowired annotation. 

    @Autowired
    public void setContextProvider(SAMLContextProvider contextProvider) {
        Assert.notNull(contextProvider, "Context provider can't be null");
        this.contextProvider = contextProvider;
    }

amiri-ge: @fhanik I can't tell if you're seriously thanking me or being sarcastic. No offense taken either way but let me know if you want me to remove it. Sorry I didn't catch that earlier. The trouble with annotations and conventions is that you have to know ahead of time what to look for.

fhanik: @amiri-ge - not sarcastic, one shouldn't have to view source code of an external library to see what the hell is going on :)

amiri-ge: @fhanik agreed :)

amiri-ge:
@fhanik I've checked in an alternate implementation for the integration test that relies on the HTTP archive log produced by the PhantomJSDriver. Let me know which one you prefer and I'll restructure the commits to a single one.

I looked into creating an integration test in conjunction with a sample app but the benefits do not outweigh the effort involved (in my opinion).

fhanik:
Looks great, squash it, and we will merge it

Issue #173, opened 2015-06-25
CORS XHR Support


amiri-ge:
Modern browsers include a "X-Requested-With: XMLHttpRequest" header when RESTful calls are performed through the XMLHttpRequest library. Since the header is not allowed by default CORS policies, requests that would otherwise be allowed by CORS standard policy will fail across different origins, thus providing a mitigation against CSRF.

However, when using implicit grant, it is useful to allow cross-origin XHR requests for endpoints such as /logout.do and /userinfo. At the same time, one should not allow this across the board. We've developed a CORS filter that allows these requests based on a whitelist of both allowed endpoints and origins but are hesitant to submit this as a pull request because:

1. We believe it might be something you are already be working on.
2. Given that CORS is enforced at the browser, it's a difficult feature to test.

In summary, if you have plans to support such a feature please let us know. If not we would like to request this as a future feature. Finally, let us know if you would be interested in a pull request for this feature.

fhanik:
1. We are not working on such feature
2. Take a look at our integration tests for how we test browser and UI features. It may be possible to leverage Selenium to achieve your goals

It would be easier for us to see the PR to understand what you are trying to achieve.

amiri-ge:
@fhanik I submitted a pull request to provide more context for this issue.

fhanik:
https://github.com/cloudfoundry/uaa/commit/6b5cd4867a785f5e8a27621c320d77431f75a337

PR #174, opened 2015-06-25
Modified CorsFilter to support XHR for whitelisted request URIs and origins.

amiri-ge:
Please see https://github.com/cloudfoundry/uaa/issues/173 for reference. This pull request is to provide more context for that issue.

### DIFF
--- a/common/src/main/java/org/cloudfoundry/identity/uaa/web/CorsFilter.java
+++ b/common/src/main/java/org/cloudfoundry/identity/uaa/web/CorsFilter.java
@@ -14,19 +14,123 @@
 package org.cloudfoundry.identity.uaa.web;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
 
+import javax.annotation.PostConstruct;
 import javax.servlet.FilterChain;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import org.apache.commons.lang.StringUtils;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpStatus;
 import org.springframework.web.filter.OncePerRequestFilter;
 
+/**
+ * 
+ * Modern browser include the X-Requested-With header when making calls through
+ * the XMLHttpRequest API which allows the server CORS filtering to mitigate
+ * against CSRF attacks performed by XHR requests. However, in some situations
+ * XHR requests are useful. For example, when a single page JavaScript apps that
+ * implements login using implicit grant wants to: 1) log the user out by
+ * calling the /logout.do URI 2) get user information by calling the /userinfo
+ * URI.
+ * 
+ * To enable the scenarios described above, this filter allows CORS requests to
+ * include the "X-Requested-With" header for a whitelist of URIs and origins and
+ * only for the HTTP GET method.
+ * 
+ * The implementation is based on guidance from:
+ * http://www.w3.org/TR/cors/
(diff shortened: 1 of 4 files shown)


fhanik:
Thanks, testing this should be pretty straight forward using MockMvc

We define all our filters in the Spring configuration files

https://github.com/cloudfoundry/uaa/blob/feature/revert_password_policy_defaults/uaa/src/main/webapp/WEB-INF/spring-servlet.xml#L86-86

This allows us to do full web requests without ever starting up a server that requires web.xml

I would also use org.springframework.util.StringUtils instead of the HsqlDB library, where you have the method hasText and isEmpty to achieve what you need. The HsqlDB library dependency should only exist from the perspective of having an in memory DB, not as class libraries used by us.

amiri-ge:
Good catch on the StringUtil. Actually, I believe we meant to use apache commons StringUtils but got burned by the IDE's autocomplete.

As for the testing. It's not the server side that worries me so much as the browser side. Ultimately, it's the browser that enforces CORS based on the headers provided by the server. The real question is how can we create automated testing for browser side CORS behavior similar to say: http://test-cors.org ?

fhanik:
I believe it is out of scope for the UAA to test whether the browser enforces CORS or not. Our responsibility is that with given input we produce output, and we should be able to test that by providing test cases for everything that we build. Both happy day cases and error cases.

amiri-ge:
@fhanik Understood. So if we cleaned this up and added some MockMvc test cases is this something you would be interested in?

fhanik:
absolutely, the only thing I'd ask for is

1. Put filter in the Spring config
2. Provide a reference for the rules being implemented (ie, what are you basing your implementation on)
3. Mock mvc tests

after that there won't be any issue merging the PR

amiri-ge:
@fhanik I'm running into an baffling problem with the MockMvc tests. Here are the facts.
1. All integration tests pass when I run "gradle integrationTest". Except 3 tests which are annotated as Ignore and are therefore skipped.
2. The CORS filter feature seems to work just fine when testing from the browser using test-cors.org
3. The MockMvc tests that I added work just fine when I run them from within Eclipse.

Here's the problem: When I "gradle test" my MockMvc tests are skipped entirely. There is no indication in the report or in the gradle output (using --debug) as to why they are being skipped.

Any idea what might cause this? I'm going to update the pull request so you can take a look at the code.

fhanik:
When you run gradlew test it skips the individual tests, but runs them all in a suite (so that the app context is only created once for all MVC tests)

https://github.com/cloudfoundry/uaa/blob/develop/uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/DefaultConfigurationTestSuite.java#L50-L70

The individual test detects if it is running in the suite

https://github.com/cloudfoundry/uaa/blob/develop/uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/InjectedMockContextTest.java#L29-29

So if you wish to run all MVC tests fast in your IDE, then you would run the DefaultConfigurationTestSuite

fhanik:
BTW, this is where your CorsFilter is already defined, not in uiSecurity, which only covers a small portion of the requests, and not all of them (see my earlier comments in this thread)

https://github.com/cloudfoundry/uaa/blob/develop/uaa/src/main/webapp/WEB-INF/spring-servlet.xml#L58-L96

amiri-ge:
@fhanik I see your point about redefining the bean. Unfortunately, my understanding of Spring is too limited to know where I could put the filter in a way that it would cover both the login and uaa endpoints. I'm open to suggestions.

fhanik:
@amiri-ge - remove your definition of the bean, and remove the insertion into uiSecurity.

The CorsFilter is already there, it was there before you started coding. All you are doing is adding features to it. So you don't have to change Spring XML files for your PR at all. Hope that makes sense.

amiri-ge:
@fhanik - It makes perfect sense. I had failed to see the import for classpath:login-ui.xml in spring-servlet.xml. Thanks for the help.

amiri-ge:
The implementation was based on guidance from:
http://www.w3.org/TR/cors/
https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS

I included this information in the comment for the CorsFilter class as well.

PR #191, opened 2015-07-10
Modified CORS filter so that allowed headers for XHR requests is configurable.

amiri-ge:
We rolled the CORS changes out into our staging environment and our front-end development teams discovered that they we needed to relax our policy on allowed headers. Thus, I made allowed headers configurable. I also discovered some issues with the validation of configuration properties while deploying and fixed those.

These changes have been tested in our staging environments by teams building single page JS apps that makes CORS XHR requests in Firefox and Chrome.

### DIFF
--- a/common/src/main/java/org/cloudfoundry/identity/uaa/web/CorsFilter.java
+++ b/common/src/main/java/org/cloudfoundry/identity/uaa/web/CorsFilter.java
@@ -76,6 +76,9 @@ public class CorsFilter extends OncePerRequestFilter {
 
     private final List<Pattern> corsXhrAllowedOriginPatterns = new ArrayList<Pattern>();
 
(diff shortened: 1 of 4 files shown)


common/src/main/java/org/cloudfoundry/identity/uaa/web/CorsFilter.java
```diff
@@ -76,6 +76,9 @@
 
     private final List<Pattern> corsXhrAllowedOriginPatterns = new ArrayList<Pattern>();
 
+    @Value("#{'${cors.xhr.allowed.headers:Accept,Authorization}'.split(',')}")
```
fhanik: The preference is still to keep the configuration in the XML files, and not let it trickle into the Java files

PR #192, opened 2015-07-15
Fix extraction of authentication exception in sample.

thomasdarimont:
We now use the `WebAttributes.AUTHENTICATION_EXCEPTION` key as a replacement for the (deprecated) and in the meantime removed constants from `AbstractAuthenticationProcessingFilter`.


thomasdarimont:
Choose wrong target branch - closing this PR and creating a new one,...

PR #197, opened 2015-07-23
Modified data source spring declaration so that it is possible to configure database connection pool in cloud profile.

amiri-ge:
Something went wrong with the first pull request so I am resubmitting.

In essence we needed a way to configure the maxActive parameter of the connection pool when using the cloud profile. This turned out to be a difficult problem to solve. We're submitting with the hope this might help someone else out.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,6 +16,8 @@ buildscript {
     classpath group: 'postgresql', name: 'postgresql', version:'9.1-901.jdbc3'
     classpath group: 'org.flywaydb', name: 'flyway-gradle-plugin', version: '3.0'
     classpath 'org.springframework.build.gradle:propdeps-plugin:0.0.7'
+    classpath 'org.springframework.cloud:spring-cloud-spring-service-connector:1.1.1.RELEASE'
+    classpath 'org.springframework.cloud:spring-cloud-cloudfoundry-connector:1.1.1.RELEASE'
   }
 }
 
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -51,11 +51,12 @@ dependencies {
   compile group: 'org.mariadb.jdbc', name: 'mariadb-java-client', version:'1.1.8'
   compile group: 'com.googlecode.flyway', name: 'flyway-core', version:'2.3.1'
   compile group: 'org.hsqldb', name: 'hsqldb', version:'2.3.1'
+  compile group: 'org.springframework.cloud', name: 'spring-cloud-spring-service-connector', version:'1.1.1.RELEASE'
+  compile group: 'org.springframework.cloud', name: 'spring-cloud-cloudfoundry-connector', version:'1.1.1.RELEASE'
 
   compile(group: 'org.springframework.security.extensions', name: 'spring-security-saml2-core', version:parent.springSecuritySamlVersion) {
     exclude(module: 'bcprov-jdk15')
   }
-
 
   provided group: 'javax.servlet', name: 'javax.servlet-api', version:'3.0.1'
 
--- a/common/src/main/java/org/cloudfoundry/identity/uaa/UaaConfiguration.java
+++ b/common/src/main/java/org/cloudfoundry/identity/uaa/UaaConfiguration.java
@@ -124,7 +124,7 @@ public class UaaConfiguration {
         public boolean logabandoned;
         public int abandonedtimeout;
         public long evictionintervalms;
-
+        public long maxwaittime;
     }
 
     public static class Logging {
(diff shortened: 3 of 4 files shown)


fhanik:
@amiri-ge This PR pulls in the spring cloud dependency while the UAA doesn't use those projects.
I'm not sure why we need it at all.

amiri-ge:
@fhanik It's up to you guys. From a user perspective, pushing UAA to CF as an app is a useful feature, and if that's going to continue to be a legitimate way to deploy UAA then there should be a better story around configuring DB connection pooling, and leveraging the bind service capabilities in CF.

As users, we had a bad experience pushing UAA to CF with DB connection pool settings we could use in production. This is the solution we came up with and we're sharing it in the spirit of helping others that might have the same problem.

It may not be the best solution to the problem but it's the only one we could come up with. If it's not worth merging, than I would at least petition your team to come up with a better solution because the status quo does not reflect well on UAA, which is otherwise a pretty cool project.

sreetummidi:
@amiri-ge 
Lets get together to discuss this further.
I will send an invite out

-Sree

fhanik:
@amiri-ge I agree that it should be easy to deploy to cloud foundry. I don't think that requirement has to be met by pulling in a dependency on the spring cloud library. after all, all we do is provide a DataSource bean, that can be manipulated anyway. Seems like that would belong in the library that is manipulated.

However, you can probably achieve the exact same thing without importing the library
```
<beans profile="cloud">
        <cloud:data-source id="dataSource">
            <cloud:pool
                max-wait-time="${database.maxwaittime:60000}"
                pool-size="${database.maxactive:100}" />
        </cloud:data-source>
</beans>
```

Just use <bean> instead of <cloud> and you can probably achieve exactly what you want without us importing a library that we don't use. The other thing that will happen immediately is that our spring cloud version will become out of sync with CF build pack cloud version. It's a dependency we don't want to maintain.

amiri-ge:
@fhanik I've been able to reverse-engineer the spring cloud namespace handler and have implemented this feature as you have directed. However, it seems like none of the java build packs in our cloud include the spring cloud libraries. Can you please provide a reference to the build pack that you were alluding to in your previous comments? I would like to see if our CF operations team can make it available to us. Otherwise, there's no way for me to test this.

fhanik:
I am probably wrong in the Java build pack comment. But I am still referring to the encompassing library that causes the issue, wherever that library comes from. The UAA war file in no way uses the Spring Cloud libraries, so it should not be a dependency from the UAA.

amiri-ge:
@fhanik I'm following up on whether such a build pack exists. In the event that it does not, what are your feelings on a build flag that would optionally include the spring-cloud libraries as a dependency?

fhanik:
It is still a negative on including the libraries. If we don't use them, we shouldn't include them. Managing dependencies and their versions is a chore that is best if minimized. 

If you ran into this issue, then somehow you included these libraries, and the libraries tweaked the UAA configuration on its own. It would be the responsibility of this library that performs that very action to do it properly or not do it at all. So a configuration flag, should be in the tweaking library. 

In your comment ```From a user perspective, pushing UAA to CF ``` implies that these libraries are added when you push UAA to CF as an application. So these libraries must be pulled in somewhere.

amiri-ge:
@fhanik I think what you are saying is that this is essentially a spring cloud issue and it needs to be resolved in spring cloud. Which, I think is a fair assessment. So it's look like we'll have to take this up with them.

In the meanwhile this is the only workaround we have. Which is useful, and currently the only solution to the problem, but I can see why you would not want to pull it into UAA.

fhanik:
If we can pull in a custom XML config without libraries, then that would be fine as a temporary solution.
But we are following up with the Spring Cloud team to address this issue.

scottfrederick:
> In your comment "From a user perspective, pushing UAA to CF" implies that these libraries are added when you push UAA to CF as an application. So these libraries must be pulled in somewhere.

The Java buildpack includes a [Spring auto-reconfiguration](https://github.com/cloudfoundry/java-buildpack/blob/master/docs/framework-spring_auto_reconfiguration.md) framework component. This framework component, when triggered, adds a [java-buildpack-auto-reconfiguration](https://github.com/cloudfoundry/java-buildpack-auto-reconfiguration) library to the application's classpath. The java-buildpack-auto-reconfiguration library uses Spring Cloud Connectors to replace the `DataSource` bean provided by the UAA XML configuration with one that is configured to use a bound database service. So the Java buildpack indirectly adds Connectors to the application classpath. 

Auto-reconfiguration does not allow customization of the `DataSource` bean. The only option for customizing the database connection properties is to pull in Connectors explicitly (which will disable the java-buildpack-auto-reconfiguration library) and use its XML or Java configuration.

PR #200, opened 2015-08-04
Added origin in JWT token and refresh token.

amiri-ge:
A number of our users are building multi-tenant applications which receive tokens from zones with multiple identity providers. In some scenarios they need to know what identity provider authenticated the user. Also this can be useful to disambiguate users from different identity providers with the same username.


sreetummidi:
Thank you Dario. We will get this merged in !

PR #202, opened 2015-08-10
Respect nonce

ysoldak:
OpenID Connect implementation doesn't respect *nonce* sent from client, while it must do that.
Here is the implementation of the feature.
This fixes [cfid-3214](https://www.pivotaltracker.com/n/projects/997278/stories/82159480)

Formal description from http://openid.net/specs/openid-connect-core-1_0.html#IDToken
**nonce**
*String value used to associate a Client session with an ID Token, and to mitigate replay attacks. The value is passed through unmodified from the Authentication Request to the ID Token. If present in the ID Token, Clients MUST verify that the nonce Claim Value is equal to the value of the nonce parameter sent in the Authentication Request. If present in the Authentication Request, Authorization Servers MUST include a nonce Claim in the ID Token with the Claim Value being the nonce value sent in the Authentication Request. Authorization Servers SHOULD perform no other processing on nonce values used. The nonce value is a case sensitive string.*

### DIFF
--- a/common/src/main/java/org/cloudfoundry/identity/uaa/oauth/Claims.java
+++ b/common/src/main/java/org/cloudfoundry/identity/uaa/oauth/Claims.java
@@ -45,4 +45,5 @@ public class Claims {
     public static final String AUTH_TIME = "auth_time";
     public static final String ZONE_ID = "zid";
     public static final String REVOCATION_SIGNATURE = "rev_sig";
+    public static final String NONCE = "nonce";
 }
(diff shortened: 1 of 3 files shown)


fhanik:
Thank you for your contribution, kindly let us know when you have a CLA signed @ysoldak

ysoldak:
@fhanik Thanks, shall send it shortly, waiting for a response from our legal guys.

ysoldak:
Fareoffice should be an approved organization now. @fhanik 
Please, check PRs #203 and #204, they are also clear now, from CLA point of view.

PR #203, opened 2015-08-10
Fix OpenID Connect code flow by always returning id_token

ysoldak:
Token Endpoint must always return id_token; client not required to send response_types with "id_token" in Token Request, expecting this violates the specification and ruins code flow (for apache module mod_auth_openidc at least).
See http://openid.net/specs/openid-connect-core-1_0.html#TokenRequest

### DIFF
--- a/common/src/main/java/org/cloudfoundry/identity/uaa/oauth/token/UaaTokenServices.java
+++ b/common/src/main/java/org/cloudfoundry/identity/uaa/oauth/token/UaaTokenServices.java
@@ -325,8 +325,7 @@ public class UaaTokenServices implements AuthorizationServerTokenServices, Resou
     }
 
     private void populateIdToken(OpenIdToken token, Set<String> scopes, Set<String> responseTypes) {
-        if (scopes.contains("openid") &&
-            responseTypes.contains(OpenIdToken.ID_TOKEN)) {
+        if (scopes.contains("openid")) {
             token.setIdTokenValue(token.getValue());
         }
     }
--- a/uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/token/TokenMvcMockTests.java
+++ b/uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/token/TokenMvcMockTests.java
@@ -691,6 +691,8 @@ public class TokenMvcMockTests extends InjectedMockContextTest {
         UsernamePasswordAuthenticationToken auth = new UsernamePasswordAuthenticationToken(p, "", UaaAuthority.USER_AUTHORITIES);
         Assert.assertTrue(auth.isAuthenticated());
 
+        //code flow defined in - response_types=code
+        //http://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth
         SecurityContextHolder.getContext().setAuthentication(auth);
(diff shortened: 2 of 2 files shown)


fhanik:
Is Hybrid flow what you need? https://github.com/cloudfoundry/uaa/blob/develop/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/OpenIdTokenGrantsIT.java#L239-239

We do not want an open id token returned unless specifically requested. We're an OAuth server first, OpenID Connect compatibility is something we are working towards, but shouldn't override or mess up our OAuth behavior.

ysoldak:
Hmm, last time I've tried UAA + mod_auth_openidc hybrid flow was not working, shall re-check.

The current behaviour breaks RFC6749 (OAuth 2.0), as I see it. :-1: 
"response_type" parameter expected to be send to authorization endpoint and nowhere else. Token endpoint doesn't expect it, uses "code" parameter instead.

I'm not going to argue about the policies you guys've decided upon. :wink: 
I just see (correct me if wrong, please) UAA doesn't follow OpenID Connect spec in the case of Code flow (expecting something not defined by the spec), if that's by design I'm fine with that, we have that thing working for us in our fork. :stuck_out_tongue_winking_eye:

fhanik:
We're still going through the motions of this PR. and will be coming back to it shortly

common/src/main/java/org/cloudfoundry/identity/uaa/oauth/token/UaaTokenServices.java
```diff
@@ -325,8 +325,7 @@ private OAuth2AccessToken createAccessToken(String userId, String username, Stri
     }
 
     private void populateIdToken(OpenIdToken token, Set<String> scopes, Set<String> responseTypes) {
-        if (scopes.contains("openid") &&
-            responseTypes.contains(OpenIdToken.ID_TOKEN)) {
+        if (scopes.contains("openid")) {
```
fhanik: The check here must be for the 'scope' HTTP request parameter. The scope parameter contains openid we return id_token.
The scopes parameter is a list of all user scopes. We're still looking into how to consolidate this

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/token/TokenMvcMockTests.java
```diff
@@ -722,10 +724,6 @@ public void testOpenIdToken() throws Exception {
             .session(session)
             .param(OAuth2Utils.GRANT_TYPE, "authorization_code")
             .param("code", code)
-            .param(OAuth2Utils.RESPONSE_TYPE, "token id_token")
-            .param(OAuth2Utils.SCOPE, "openid")
```
fhanik: The scope parameter MUST be present, and contain openid as a scope. Otherwise the behavior is unmodified

fhanik:
Fixed via https://github.com/cloudfoundry/uaa/commit/5f8286d2e9aa383611c39370d04b33e981ec23a8

ysoldak:
Thank you, Filip!

fhanik:
@ysoldak Take note, it's according to spec. Slightly different than you had it. You have to explicitly define the scope=...openid... parameter in the /oauth/authorize request.

PR #206, opened 2015-08-12
FastRemoteTokenServices is a client side JWT token verifier that does not call UAA's /check_token endpoint all the time.

amiri-ge:
The UAA RemoteTokenService is used in the UAA samples to verify UAA issued tokens by making a call to the UAA "/check_token" endpoint. The downside of that approach is that the client incurs the penalty of a network round trip for every request that requires token verification. However, when UAA uses asymmetric cryptography to sign tokens the client should not have to necessarily incur this penalty.

This "FastRemoteTokenService" loads the UAA server public key at startup and leverages the properties of asymmetric cryptography to verify tokens without the need for a call to "/check_token". This speeds up the time needed to verify tokens issued by UAA.

### DIFF
--- a/common/src/main/java/org/cloudfoundry/identity/uaa/oauth/FastRemoteTokenServices.java
+++ b/common/src/main/java/org/cloudfoundry/identity/uaa/oauth/FastRemoteTokenServices.java
@@ -1,0 +1,352 @@
+package org.cloudfoundry.identity.uaa.oauth;
+
+import static org.cloudfoundry.identity.uaa.oauth.Claims.EXP;
+
+import java.io.IOException;
+import java.io.UnsupportedEncodingException;
+import java.util.Collection;
+import java.util.Collections;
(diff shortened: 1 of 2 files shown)


william-tran:
@amiri-ge This has already been done in spring-cloud-security: https://github.com/spring-cloud/spring-cloud-security/blob/1.0.2.RELEASE/spring-cloud-security/src/main/java/org/springframework/cloud/security/oauth2/resource/ResourceServerTokenServicesConfiguration.java#L244 

You need to set ```spring.oauth2.resource.jwt.keyUri``` to the /token_key endpoint of your UAA to enable the configuration. And I think this functionality is going to be part of Spring Boot 1.3, but until then, you need to pull in spring-cloud-security.

sreetummidi:
Closing. This will be addressed as part of https://www.pivotaltracker.com/story/show/108267354

PR #209, opened 2015-08-19
Add flyway dependency and use interface for tokenServices

mikeroda:
Pull request to add flyway dependency and use an interface for tokenServices

### DIFF
--- a/common/src/main/resources/spring/data-source.xml
+++ b/common/src/main/resources/spring/data-source.xml
@@ -81,7 +81,7 @@
         </property>
     </bean>
 
-    <bean id="jdbcTemplate" class="org.springframework.jdbc.core.JdbcTemplate">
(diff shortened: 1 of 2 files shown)


fhanik:
I'm confused about the flyway commit, what does it address?

mikeroda:
The SecurityFilterChainPostProcessor does actually depend on flyway migration so the database tables are created before this bean is initialized. Right now it is getting luck with the bean initialization order but this change will ensure the correct order of dependencies. I can go into more specifics of how I made it fail if necessary but this is more of a cleanup item.

fhanik:
the security filter chain depends on other beans that depend on other beans that depend on the jdbctemplate - wouldn't it therefor be more prudent that the jdbc template depended on flyway?

https://github.com/cloudfoundry/uaa/blob/develop/common/src/main/resources/spring/data-source.xml#L84-84

and thus, your dependency would be resolved.

mikeroda:
Good point. Let me rework the commit that way.

PR #211, opened 2015-08-20
Update docs to show aud derivation from scope when there's multiple dots

willtran-:
Eg scope zones.12345.admin results in aud zones.12345


fhanik:
Reopen for develop please. This way we don't run into conflicts

PR #215, opened 2015-08-27
Fix obvious bugs found during automatic code inspection.

thomasdarimont:
Fixed usage of approval parameters in JdbcApprovalStore.
Previously the approval parameter wasn't used at all when revokeApproval was called.
Corrected validation check in ClientDetailsModification. Previously the check returned
always true and didn't consider the "none" action.
Corrected incorrect equals check in ExtendedUaaAuthority.

I found these when I ran "analyze code" with the latest intellij idea version against
the uaa codebase and it brought up several additional warnings and potential problems.
I strongly suggest to run a code analysis on the code yourself.


fhanik:
thank you very much

Clearly the first method is missing a test case in the existing code. This means we don't have code coverage on that method, even after your commit. 

You can either add code coverage to this PR, or I will merge it in manually and add the tests later.

plfx:
This has been merged in to cloudfoundry:develop. https://github.com/cloudfoundry/uaa/commit/c52dda0ebf181d952afa67ba93b622a20d7df7e1

PR #217, opened 2015-08-29
SAML provider: possibility to specify 'addShadowUserOnLogin' in login.yml

MatthiasWinzeler:
At the moment, it is not possible to specify the newly introduced property `addShadowUserOnLogin` for saml providers in `login.yml` - in the database, it always ends up as `"addShadowUserOnLogin":false`.

This PR extends the `IdentityProviderConfigurator` to respect the this property also - so it can be specified in login.yml.

I also added the property in the example configurations in login.yml (this is only for demonstration, it still defaults to `true` to not break things).


fhanik:
Thank you very much for your contribution
Please make the pull requests for `develop` branch
And test cases would be great

PR #225, opened 2015-09-15
Do not enforce CORS filter on same origin requests

### LINKED ISSUE #224: CORS filter blocks same origin requests
#173 added XHR support to the CORS filter. This code always expects an Origin header with the "X-Requested-With: XMLHttpRequest" header. However, the browser does not add an Origin header on same origin GET requests. The result is that same origin requests get a 403 error, and there is no way to whitelist these requests since there is no Origin header passed. The only workaround is to remove the CORS filter completely.

mikeroda:
Resolves #224. Browsers do not send the Origin header on same origin (scheme/host/port match) requests, so the filter should ignore such requests.

### DIFF
--- a/common/src/main/java/org/cloudfoundry/identity/uaa/web/CorsFilter.java
+++ b/common/src/main/java/org/cloudfoundry/identity/uaa/web/CorsFilter.java
@@ -125,6 +125,9 @@ public class CorsFilter extends OncePerRequestFilter {
                 return;
             }
             String origin = request.getHeader(HttpHeaders.ORIGIN);
+            if (!isCrossOriginRequest(origin)) {
+                return;
+            }
             String requestUri = request.getRequestURI();
             if (!isCorsXhrAllowedRequestUri(requestUri) || !isCorsXhrAllowedOrigin(origin)) {
                 response.setStatus(HttpStatus.FORBIDDEN.value());
@@ -156,6 +159,15 @@ public class CorsFilter extends OncePerRequestFilter {
         return StringUtils.hasText(xRequestedWith)
                 || (StringUtils.hasText(accessControlRequestHeaders) && containsHeader(
                         accessControlRequestHeaders, "X-Requested-With"));
+    }
+
+    private boolean isCrossOriginRequest(final String origin) {
+        if (StringUtils.isEmpty(origin)) {
+            return false;
+        }
+        else {
+            return true;
+        }
     }
 
     void buildCorsXhrPreFlightResponse(final HttpServletRequest request, final HttpServletResponse response) {
@@ -224,10 +236,6 @@ public class CorsFilter extends OncePerRequestFilter {
     }
 
     private boolean isCorsXhrAllowedOrigin(final String origin) {
-        if (StringUtils.isEmpty(origin)) {
-            return false;
-        }
-
         for (Pattern pattern : this.corsXhrAllowedOriginPatterns) {
             // Making sure that the pattern matches
             if (pattern.matcher(origin).find()) {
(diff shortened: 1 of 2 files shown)


fhanik:
Thanks for the submission @mikeroda 

I'm a bit confused. When you say it should 'ignore the filter'. Doesn't the code still do the exact same thing as before? If origin is empty it will abort the filter chain. Meaning, it doesn't invoke chain.

    filterChain.doFilter(request, response);

It also does a return without setting any response status. Meaning, it will be up to the container to produce a response, if any.

mikeroda:
The difference is before it would return a 403, now it will return without setting an error code in the response. Good catch on your last point. I've added another commit so it continues the filter chain rather than just return.

amiri-ge:
If I understand mike's logic correctly: we should not assume that every XHR request is a CORS request. I always thought that Spring framework would only invoke this filter if it detected a CORS request. If it does not then it makes sense that the filter should simply bypass the filter and move on to the next one if the request is not cross origin. In retrospect, I suppose we never realized this before since all of our relying party applications are on different origins than UAA.

fhanik:
@amiri-ge @mikeroda The filter is the one that is responsible for detecting if the request is a CORS request. There is no logic involving CORS anywhere outside this filter. So the filter must perform correctly.

I am still not fully up to speed on the CORS stuff, that's why I'm soliciting your comments here.

amiri-ge:
According to http://www.w3.org/TR/cors/

"Server-side applications are enabled to discover that an HTTP request was deemed a cross-origin request by the user agent, through the Origin header."

So detecting the presence of the "origin" header is sufficient to determine whether the request is cross origin. In light of that, simply moving on to the next filter in the filter chain if there is no origin header is the right thing to do.

amiri-ge:
@mikeroda I would simply test for the presence/absence of the origin header and if it is not there simply call filterChain.doFilter(request, response)

fhanik:
yes, I believe that is what @mikeroda 's PR actually does. If there is no Origin, it just calls next filter

amiri-ge:
@fhanik I'm looking at the code and I don't see that. Instead, what I see is that it skips the XHR CORS code if there is no CORS header header and goes into the CORS code after that.

I think it would be cleaner to do:

if (!isCrossOriginRequest(request)) {
    filterChain.doFilter(request, response);
    return;
}

fhanik:
@amiri-ge line 161 only returns true if there is an Origin header

amiri-ge:
@fhanik What I mean is that, in the code I am looking at, lines 142-150 are still potentially executed.

fhanik:
are you saying that pre-flight request is not valid without Origin?

amiri-ge:
@fhanik Well yes. Once again, referencing http://www.w3.org/TR/cors/#resource-preflight-requests

"6.2 Preflight Request

In response to a preflight request the resource indicates which methods and headers (other than simple methods and simple headers) it is willing to handle and whether it supports credentials.

Resources must use the following set of steps to determine which additional headers to use in the response:

1.   If the Origin header is not present terminate this set of steps. The request is outside the scope of this specification."

I'm reading that as "if there is no origin header then it's not a CORS request so don't do anything".

fhanik:
@mikeroda Thanks, that means this PR is still not complete. I would go with @amiri-ge suggestion.

mikeroda:
Yep that makes sense. I'll submit another commit later this evening.

mikeroda:
The latest commit in this PR puts the check for the Origin header up front and just continues the filter chain and return if it's not present, thus also not adding the Access-Control-* headers.

amiri-ge:
Looks good to me, though I think the maintainers might ask you to squash your commits. Nice job finding this BTW.

fhanik:
Thanks, we will get this in shortly.

fhanik:
Merged in https://github.com/cloudfoundry/uaa/commit/6fbfb589c830a48790430cea9dac0746fb821709
Squashed the commits

PR #226, opened 2015-09-16
Added code to validate origin header that is reflected back in the CO…

amiri-ge:
…RS response.


fhanik:
@amiri-ge Can you comment on https://github.com/cloudfoundry/uaa/pull/225

PR #236, opened 2015-10-06
Set the template abandonedtimeout to an integer

extremesurfer:
abandonedtimeout should be set to an integer, not a boolean. With it set to a boolean startup fails and no migrations run.


fhanik:
@extremesurfer  Thank you for your PR. 

Merged via https://github.com/cloudfoundry/uaa/commit/8d50bb7ede5fd80bb5bc341b2983cbc1a0ed7561

Please direct future PRs to the `develop` branch. Then we can automatically merge them from here.

PR #243, opened 2015-10-09
Try to obtain the user details origin from x-forwarded-for header

mikeroda:
Use the x-forwarded-for HTTP request header, if it's available, for the UaaAuthenticationDetails origin. So if UAA is behind a reverse proxy, the origin will reflect the actual user IP address and not the proxy server's.


mbhave:
@mikeroda; we are going to revert this commit.  Upon closer inspection we realized that we rely on tomcat remoteipvalve to handle the x-forwarded-for header and set the remoteaddress.  

Are you using UAA on tomcat with remoteipvalve?  Or some other setup?

fhanik:
@mikeroda Reading x-forwarded-for header, means that a client can modify it's own IP address by simply setting the header. This header should only be read when the request comes from a trusted reverse proxy. See the tomcat.server.xml.erb in cf-release, where we configure the RemoteIPValve

mikeroda:
OK

PR #247, opened 2015-10-19
Serialize/deserialize UserAuthenticationDetails in oauth_code data

mikeroda:
The data column of the oauth_code table contains serialized bytes of
the Oauth2Authentication; however, the UserAuthenticationDetails were
omitted. This commit adds the details so they can be deserialized
when reading the oauth2 code out of the database.

### DIFF
--- a/common/src/main/java/org/cloudfoundry/identity/uaa/oauth/token/UaaTokenStore.java
+++ b/common/src/main/java/org/cloudfoundry/identity/uaa/oauth/token/UaaTokenStore.java
@@ -16,6 +16,7 @@ package org.cloudfoundry.identity.uaa.oauth.token;
 
 
 import com.fasterxml.jackson.core.type.TypeReference;
+
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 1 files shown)


mbhave:
@mikeroda thanks for the PR. This has a small backwards compatibility issue for existing codes that don't have USER_AUTHENTICATION_DETAILS stored on them. Could you add code that preserves the existing behavior for these cases? 

We're also curious about where you intend to use these details.

mikeroda:
I'll take care of the backward compatibility issue. 

I need to add the end-user's remote IP address as an additional claim in the tokens. The IP is in the origin attribute of UaaAuthenticationDetails, not to be confused with the origin attribute of UaaUser, which was also added recently as a claim in the token.

fhanik:
Don't merge this one in. This PR highlights that we are losing more data from the UaaAuthentication object. I will work on this as part of https://www.pivotaltracker.com/story/show/106030472

fhanik:
@mikeroda This has been completed as part of 
https://github.com/cloudfoundry/uaa/commit/c6e86757fde23bfae6df10e9debe2f92b1ce7156

PR #263, opened 2015-11-29
[identity-zone] enhance logging to avoid silent failure

tony-kerz:
i'm new to uaa, and still attempting to acclimated to the configuration in uaa.yml.

i initially didn't have a `zones.internal.hostnames` stanza and the log messages gave 
me no indication of what the issue was, so these log changes attempt to assist with that.

regards,
tony.

### DIFF
--- a/common/src/main/java/org/cloudfoundry/identity/uaa/zone/IdentityZoneResolvingFilter.java
+++ b/common/src/main/java/org/cloudfoundry/identity/uaa/zone/IdentityZoneResolvingFilter.java
@@ -22,7 +22,6 @@ import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
-import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Set;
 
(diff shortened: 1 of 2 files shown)


fhanik:
Three things we would like for you to review

1. Make sure you have a CLA signed. Let us know when you do.
2. Remove all tabs and replace them with 4 spaces
3. When a zone is not found, it should still be a debug message, not info. The error message to the request itself contains all details. That is why it is not considered an info level message.

tony-kerz:
thanks filip,

1. i sent in the CLA.
1. roger on spaces v tabs
1. can u do me a favor and place a github comment on the lines u want as `debug` v `info`

cheers,
tony.

common/src/main/java/org/cloudfoundry/identity/uaa/zone/IdentityZoneResolvingFilter.java
```diff
@@ -44,22 +43,28 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
...
         String hostname = request.getServerName();
         String subdomain = getSubdomain(hostname);
+        logger.debug(String.format("hostname=[%s], subdomain=[%s]", hostname, subdomain));
+        
         if (subdomain != null) {
             try {
                 identityZone = dao.retrieveBySubdomain(subdomain);
             } catch (EmptyResultDataAccessException ex) {
-                logger.debug("Cannot find identity zone for subdomain " + subdomain, ex);
+                logger.info(String.format("Cannot find identity zone for subdomain %s", subdomain));
```
fhanik: This should still be debug because the response body contains an error message. The identityZone will be null, and a message will be part of the HTTP response body (line 63)

common/src/main/java/org/cloudfoundry/identity/uaa/zone/IdentityZoneResolvingFilter.java
```diff
@@ -44,22 +43,28 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
...
             } catch (Exception ex) {
-                logger.debug("Internal server error while fetching identity zone for subdomain" + subdomain, ex);
-                response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, "Internal server error while fetching identity zone for subdomain " + subdomain);
+            	String msg = String.format("Internal server error while fetching identity zone for subdomain %s", subdomain);
+                logger.info(msg, ex);
```
fhanik: I would make this WARN, because here, we are trapping unknown messages.

common/src/main/java/org/cloudfoundry/identity/uaa/zone/IdentityZoneResolvingFilter.java
```diff
@@ -44,22 +43,28 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
...
+                response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, msg);
                 return;
             }
         }
         if (identityZone == null) {
-            response.sendError(HttpServletResponse.SC_NOT_FOUND, "Cannot find identity zone for subdomain " + subdomain);
+        	String msg = String.format("Cannot find identity zone for subdomain %s", subdomain);
+        	logger.info(msg);
```
fhanik: This is an expected use case, there is no need to do any logging here. The only way to this code path, is logged above already.

common/src/main/java/org/cloudfoundry/identity/uaa/zone/IdentityZoneResolvingFilter.java
```diff
@@ -86,11 +91,13 @@ public void setIdentityZoneProvisioning(IdentityZoneProvisioning dao) {
     public void setAdditionalInternalHostnames(Set<String> hostnames) {
         if (hostnames!=null) {
             this.defaultZoneHostnames.addAll(hostnames);
+            logger.info(String.format("additional internal hostnames: %s", hostnames));
```
fhanik: debug for sure here, this is not informative. I would also recommend doing 

    Arrays.toString(hostnames)

fhanik:
Closing via: https://github.com/cloudfoundry/uaa/commit/4121eceee867e91c4391098e42aa49d1b1740fbc

But I think the most important change here is 
https://github.com/cloudfoundry/uaa/commit/1e39fe3d3ab03c7ae6ef10e64ee1d8e45292ca24#diff-7c09d2eec726a47a79796f5e12ae97d4

That commit makes the UAA respond on any hostname if root domains for zones have not been configured (zones.internal.hostnames)

Issue #271, opened 2015-12-11
SamlConfig defaults don't match ZoneAwareMetadataGenerator defaults


amiri-ge:
SamlConfig has:
```
    private boolean requestSigned = false;
    private boolean wantAssertionSigned = false;
```
Whereas saml-providers.xml has:
```
            <property name="requestSigned" value="${login.saml.signRequest:true}" />
            <property name="wantAssertionSigned" value="${login.saml.wantAssertionSigned:false}" />
```
I would recommend setting them both to true by default so that they are secure by default.

amiri-ge:
To add to this. It might be better for SamlConfig requestSigned and wantAssertionSigned to have some undefined state, perhaps by using ```Boolean``` instead of ```boolean```. If this case if the properties are null then they should default to the global configuration.

PR #273, opened 2015-12-11
Properties requestSigned and wantAssertionSigned true by default.

amiri-ge:
To address https://github.com/cloudfoundry/uaa/issues/271


fhanik:
Closing. We have decided not to change the defaults, but to align them. We believe that changing wantAssertionSigned could break systems that currently are not configured and don't use signed assertions.

amiri-ge:
@fhanik Would it be possible to configure the defaults in the YAML configuration since there are already settings for this already exist?

Issue #270, opened 2015-12-11
question: any way to get the "typ" in the generated token header?


tony-kerz:
i'm trying to have [a third party](https://getkong.org/plugins/jwt/) validate a uaa jwt, 
and it is rejecting the uaa token because it doesn't have:

```
"typ": "JWT"
```
in the header...  

i gather `typ` is optional, but anyone have any suggestions as to how to (optionally),
get uaa to include it in the generated tokens?

regards,
tony.

fhanik:
Thanks for filing this, we have added this to our enhancement list.

tony-kerz:
thanks filip,

i'd be happy to attempt a p/r, if you happen to be able to direct me to the area of code which might be appropriate to effect that change. i tried some file searches in eclipse against the uaa source for things like `alg` or `HS256`, but didn't turn up anything obvious. wasn't sure if that kind of change may require a change to one of the members of the spring-security family. if it requires too much time to even research right now, i understand. 

have a good holiday!

cheers,
tony.

fhanik:
https://github.com/cloudfoundry/uaa/blob/develop/common/src/main/java/org/cloudfoundry/identity/uaa/oauth/token/UaaTokenServices.java

see the class JwtHelper

PR #274, opened 2015-12-13
[token] allow for configurable inclusion of typ header

tony-kerz:
proposal to address: https://github.com/cloudfoundry/uaa/issues/270

relies on: https://github.com/spring-projects/spring-security-oauth/pull/659

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -81,6 +81,7 @@ subprojects {
     archives javadocJar
   }
   repositories {
+    mavenLocal()
     maven { url "http://maven.springframework.org/release" }
     maven { url "http://maven.springframework.org/milestone" }
     maven { url "http://maven.springframework.org/snapshot" }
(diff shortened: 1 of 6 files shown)


fhanik:
We'll have to wait for Spring Security release, we don't include snapshots, since that means the build is not reproducible when the snapshot changes.

tony-kerz:
yep, just dropped it as a place-holder. in the mean time, feel free to provide any feedback on this or spring-security commits. tx!

cromark:
@fhanik , @dsyer has updated https://github.com/spring-projects/spring-security-oauth ( see  [commit](https://github.com/spring-projects/spring-security-oauth/commit/3efc6b3b4885af9ec867085bc769ce470e2a5258) ). 

All that is required is to update [shared_versions.gradle](https://github.com/cloudfoundry/uaa/blob/master/shared_versions.gradle) with `springSecurityJwtVersion = '1.0.4.RELEASE'`

staylor14:
Current UAA adds in `typ` header. We have for now included the header code directly in the UAA, because we require the `kid` header too

    {
      "alg": "RS256",
      "kid": "legacy",
      "typ": "JWT"
    }

PR #315, opened 2016-02-16
Integrate Spring REST Docs into MockMvc-based tests

wilkinsona:
@Priyata25 As discussed on Slack, this PR puts the fundamentals in place for using Spring REST Docs with your MockMvc-based tests.

PR #316, opened 2016-02-16
Integrate Spring REST Docs into MockMvc-based tests

wilkinsona:
See #315, this time against `develop` rather than `master`


sreetummidi:
Spring rest docs have been incoporated

PR #345, opened 2016-04-18
Upgrade to Flyway 4.0

klaus-sap:


klaus-sap:
Wrong branch. Follow up in CR #349

PR #351, opened 2016-04-24
Allow setting of mail.smtp.auth and mail.smtp.starttls.enable props

wwadge:
Exposes two more SMTP properties: auth and starttls


wwadge:
I have signed and submitted the CLA

staylor14:
Merged via dbe2dd2ddcfedaa9fa5fefd35694a558b8e2ee1c

PR #387, opened 2016-06-09
Correct Content-Type in check_token endpoint

rosenhouse:
With `Content-Type: application/x-www-form-encoded` and the `POST` body equal to `token=....` we get the following `400` error from UAA:
```
Required String parameter 'token' is not present
```

When we change to `Content-Type: x-www-form-urlencoded` then we get a `200 OK` response.


staylor14:
Merged via https://github.com/cloudfoundry/uaa/commit/21fd892d908e954997f40c687a7331b1b728e6a8

Issue #392, opened 2016-06-15
External Identity Providers issuer validation issue


nigelelliott75:
I have been testing out UAA and have been trying to configure it to use Google OpenId as an External Provider. I managed to get so far and get an issue when UAA is verifying the issuer is correct coming back from Google. The issue is that the UAA code is using the tokenUrl set in uaa.yml as the expected issuer and Google is sending back a different value. 

I made a change to the UAA code to add a new configuration item called "issuer" within login.oauth.providers.<providerName> which can be set to the expected value and will be used if available, otherwise fallback to tokenUrl. Once I did this change UAA was able to work sign me in using a Google account.

I currently have the change locally and still need to do unit tests for the change, however before I submit the change wanted to check if this hasn't already been corrected or if this is the correct way of solving the issue.

I have attached a snippet of the uaa.yml file showing the configuration section the new attribute "issuer" will appear.

[uaa-extIdp-yml.txt](https://github.com/cloudfoundry/uaa/files/316210/uaa-extIdp-yml.txt)

sreetummidi:
@nigelelliott75 Happy to accept a pull request for this issue

sreetummidi:
Fixed by https://www.pivotaltracker.com/story/show/121653769

PR #393, opened 2016-06-16
External Identity Providers issuer validation issue #392

nigelelliott75:
Fix issue when validating the token's issuer(iss) coming back from an external identity provider. Previously the tokenUrl was used as the value the issuer was compared against. This may not always be the case and so a new configuration item (issuer). If "issuer" is set it will be used instead of the tokenUrl.


staylor14:
Can you please submit test cases for this?
Preferably we like tests that fail before the commit is applied, and the test pass after the commit is applied.

fhanik:
It looks like we also need to a signed CLA before we can approach this.

sreetummidi:
@nigelelliott75 Can you please provide the CLA

nigelelliott75:
I now have sent the signed CLA via email. I have also created some unit tests to test the new functionality.

PR #396, opened 2016-06-20
Add interface UaaTokenEnhancer

klaus-sap:
Add possibility to enhance access tokens and refresh tokens with custom attributes. You may plug-in a UaaTokenEnhancer into the bean UaaTokenServices. We don't use the existing interface org.springframework.security.oauth2.provider.token.TokenEnhancer, because it would be applied too late (after token serialization) and it doesn't seem to support the refresh flow.


staylor14:
Work in progress, pending merge.
https://travis-ci.org/cloudfoundry/uaa/builds/139359093

staylor14:
Pulled by 5fce9a992409310cc7f88bd6ef1b461e3a902cbc

PR #399, opened 2016-06-27
Added env variables to configure proxy for unit tests.

sanjeevchopra:


fhanik:
Let's solve this without duplicating the same code twice, to avoid inconsistencies should this change in the future.

sanjeevchopra:
@fhanik  - agreed - I will reopen this, when I get some time to finish this.

PR #400, opened 2016-07-01
Feature/scim user provisioning improvement

klaus-sap:
Before creating a user, check if user exists. This makes the implementation more robust and we can distinguish between the error "user already exists" and other database errors. In addition, we prepare the UAA for running on databases which don't support case insensitive indizes.


fhanik:
@klaus-sap we can't accept this PR until there is a test commit first. That commit should fail, until the feature commit is applied, at which point the test passes.

klaus-sap:
@fhanik Added new test method testCreateUserWithBrokenConnection as requested. The test fails before the correction is applied and it passes afterwards. Certainly, we could fix the issue of unexpected exceptions in a simpler way. But still it seems more robust to me to add an explicit search if the user exists before creating it.

klaus-sap:
Failing test testTwoLdapServers does not seem to be related to this change

fhanik:
Thanks @klaus-sap - the SCIM exception on broken connection is kind of awkward. It's not a SCIM exception. And no other SCIM endpoint does it that way. So it is inconsistent. 

The change also has the following side effect. If the index (unique key) changes in the DB, we no longer catch DuplicateKey exception, and always return 500 instead of 409.

PR #420, opened 2016-07-31
Allow customizable SAML XML Dsig algorithms (SHA-256 and SHA-512)

svrc-pivotal:
SHA-1 is the current default signature algorithm for SAML XML digital signatures in Spring SAML and thus UAA but is compromised and not accepted at some customers.   

This patch enables UAA to default to SHA-256 or SHA-512 signatures while preserving SHA-1 as the current default.


plfx:
Hello Stu.

Thanks for the pull request. We've addressed this in a slightly different way that we preferred architecturally. We thank you for raising the issue. You should see it as a configurable property in the next release, taking simply the values `SHA1`, `SHA256`, or `SHA512`.

Issue #411, opened 2016-07-22
OIDC code for getting token ignores proxy settings


DarioAmiri:
During the OIDC flow there is a step where UAA will post an authorization code to an external IdP in order to obtain the user token. Unfortunately, this does not work from behind a proxy because the underlying HttpClient is constructed in such a way that it ignores any proxy settings passed to the JVM. This issue manifests itself in the following integration tests:
- OIDCLoginIT.successfulLoginWithOIDCProvider
- OIDCLoginIT successfulLoginWithOIDCProvider_withClientContext

See

https://github.com/cloudfoundry/uaa/blob/094f75188cc0606c5194cc64cf722c54ec8bfd2b/server/src/main/java/org/cloudfoundry/identity/uaa/util/UaaHttpRequestUtils.java#L31-L34

fhanik:
I agree, we should start reducing the number of classes used to make external HTTP calls and instead consolidate them. The referenced class should probably be consolidated with all other classes that do the same thing. For example, when we fetch meta data for SAML

amiri-ge:
@fhanik Cool, I'll try to submit a pull request for this early next week unless you guys want to address it.

amiri-ge:
@fhanik Unfortunately, the SAML FixedHttpMetaDataProvider and the OIDC UaaHttpRequestUtils use different HttpClient APIs. The SAML stuff uses HttpClient whereas the OIDC stuff uses CloseableHttpClient which are incompatible. Thus, reconciling these to use the same HttpClient would require far more extensive code changes. I will submit a more targeted fix for now.

PR #421, opened 2016-08-01
Modified UaaHttpRequestUtils to build an HTTP client that uses the system proxy settings.

DarioAmiri:
See [`issue 411`](https://github.com/cloudfoundry/uaa/issues/411)


amiri-ge:
@fhanik Took me a while but here it is:
- Fix for OIDC get token request behind a proxy
- Single commit
- All tests pass
- I can confirm this works behind our proxy

Few comments
- UAA uses different HttpClient APIs across the code which makes it hard to unify the code for handling proxy issues. Ideally, the code would migrate toward a single reusable solution. I did not undertake this because the code changes would have been significant.
- The presence of a skipSslValidation flag is worrisome to me. In my experience, features like this are often abused. Why not import your trusted certificates to the java cacerts store or use a custom trust store or get a free certificate from [let's encrypt](https://letsencrypt.org)? 
- The skipSslValidation mechanism that is in place replaces the rest template in XOAuthAuthenticationManager for every request made because the skipSslValidation flag comes from a parameter. This seems wasteful.

fhanik:
This story is in progress
https://github.com/cloudfoundry/uaa/commit/871e92511fe6876b928b4ed4ef74749d6f027f4c

fhanik:
Completed by https://github.com/cloudfoundry/uaa/commit/dd8372657bdce317702e6c83ec6d2da981dc5864

fhanik:
Added proxy validation tests.

Issue #424, opened 2016-08-12
Token Signing Keys are improperly encoded


whitlockjc:
Per [RFC 7517](https://tools.ietf.org/html/rfc7517), the modulus _(`n`)_, and the exponent _(`e`)_ should both be base64url encoded, not just base64 encoded.  [Appendix A.1](https://tools.ietf.org/html/rfc7517#appendix-A.1) gives an example of this.  Based on my testing, `e` isn't really a problem because its value is a constant and that constant doesn't have values that would encode differently based on base64 vs. base64url.  But for `n`, you often see keys that fail to base64url decode.  I have created a simple Go application that demonstrates this using a modulus value from your documentation: https://play.golang.org/p/rwb8uQSDPP

plfx:
@whitlockjc Thank you for reporting this. A fix is forthcoming in the next version of the UAA, version 3.7.

whitlockjc:
If you agree with the change, I'll gladly submit a PR.  The reason I filed a bug instead of submitting a PR was because I wasn't sure if you'd agree to the change until you tackled OIDC officially.

Issue #282, opened 2015-12-29
/token_key possibly returns the wrong value for `alg`


jasonkeene:
I was a bit confused as to why the JWK that is returned by `/token_key` has `alg` set to `SHA256withRSA`:

```
curl https://uaa.run.pivotal.io/token_key -s | jq .
{
  "alg": "SHA256withRSA",
  "value": "-----BEGIN PUBLIC KEY-----\nMIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEA8/aXmEImpdwWHJlYc4G8\n3BgZVmyhCdy7SCL0kM7wV5xCvRKK0k4nKjH0QW2E+0GIKzIj4JQhYU+MeZHrArfC\nrfthIXcio/Ll6NvoTPY77XA7U6vBGCiLdGYSGrN8y064cF2uM8d3AEgTT0RzLK3E\n688Ltq38GxnoXOUuLZmXS2HeHNd2bW/k6Eyd9Z3ymmdpVZXMyLwepNxU38WQS2bJ\nPYXYvRkzoZ1ru/deExwbecI18NCeO/GKp3f8bwKuC2j3LKFJIAwW3zFoDrcAxpC/\nJDG2RSTj//CRvhtd7JkeQLVKGyIHNtACaPT3tFT6scvVXHGPB5fRTLB8Lr+mK4RI\nBwIDAQAB\n-----END PUBLIC KEY-----\n",
  "kty": "RSA",
  "use": "sig",
  "n": "APP2l5hCJqXcFhyZWHOBvNwYGVZsoQncu0gi9JDO8FecQr0SitJOJyox9EFthPtBiCsyI+CUIWFPjHmR6wK3wq37YSF3IqPy5ejb6Ez2O+1wO1OrwRgoi3RmEhqzfMtOuHBdrjPHdwBIE09EcyytxOvPC7at/BsZ6FzlLi2Zl0th3hzXdm1v5OhMnfWd8ppnaVWVzMi8HqTcVN/FkEtmyT2F2L0ZM6Gda7v3XhMcG3nCNfDQnjvxiqd3/G8Crgto9yyhSSAMFt8xaA63AMaQvyQxtkUk4//wkb4bXeyZHkC1ShsiBzbQAmj097RU+rHL1VxxjweX0UywfC6/piuESAc=",
  "e": "AQAB"
}
```

[...]

PR #425, opened 2016-08-12
oidc: set JWT iss claim to UAA server location

whitlockjc:
Right now, the iss claim in a JWT points to the OAuth endpoint used to
get the token.  This makes OIDC integration impossible because the iss
claim must match the issuer property of the OIDC configuration document.
With this change, we no longer append "oauth/token" to the issuer URL.

**Note:** I realize that OIDC support isn't official or complete but after
discussing this in Slack with @sreetummidi, I figured I would whip up a
PR to get the conversation started.  I will supply more detail into why this
PR came about in the comments below.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java
@@ -1209,6 +1209,8 @@ public class UaaTokenServices implements AuthorizationServerTokenServices, Resou
         this.issuer = issuer;
     }
 
+    /* TODO: This should be renamed to getIssuerEndpoint or something similar as its current name is confusing
+             (The method is for getting the issuer endpoint, not the OAuth token URL which the name could suggest.) */
     public String getTokenEndpoint() {
         try {
             return UaaTokenUtils.constructTokenEndpointUrl(issuer);
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/util/UaaTokenUtils.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/util/UaaTokenUtils.java
@@ -185,6 +185,6 @@ public final class UaaTokenUtils {
         if (hasText(IdentityZoneHolder.get().getSubdomain())) {
             hostToUse = IdentityZoneHolder.get().getSubdomain() + "." + hostToUse;
         }
-        return UriComponentsBuilder.fromUriString(issuer).host(hostToUse).pathSegment("oauth/token").build().toUriString();
+        return UriComponentsBuilder.fromUriString(issuer).host(hostToUse).build().toUriString();
     }
 }
--- a/server/src/test/java/org/cloudfoundry/identity/uaa/oauth/CheckTokenEndpointTests.java
+++ b/server/src/test/java/org/cloudfoundry/identity/uaa/oauth/CheckTokenEndpointTests.java
@@ -617,7 +617,7 @@ public class CheckTokenEndpointTests {
         setAccessToken(tokenServices.createAccessToken(authentication));
         Claims result = endpoint.checkToken(getAccessToken(), Collections.emptyList());
         assertNotNull("iss field is not present", result.getIss());
-        assertEquals("http://some.other.issuer/oauth/token", result.getIss());
+        assertEquals("http://some.other.issuer", result.getIss());
     }
 
     @Test
@@ -630,7 +630,7 @@ public class CheckTokenEndpointTests {
             setAccessToken(tokenServices.createAccessToken(authentication));
             Claims result = endpoint.checkToken(getAccessToken(), Collections.emptyList());
             assertNotNull("iss field is not present", result.getIss());
-            assertEquals("http://subdomain.some.other.issuer/oauth/token", result.getIss());
+            assertEquals("http://subdomain.some.other.issuer", result.getIss());
         } finally {
             IdentityZoneHolder.clear();
         }
--- a/server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServicesTests.java
+++ b/server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServicesTests.java
@@ -149,7 +149,7 @@ public class UaaTokenServicesTests {
     public static final String REFRESH_TOKEN = "refresh_token";
     public static final String IMPLICIT = "implicit";
     public static final String CLIENT_AUTHORITIES = "read,update,write,openid";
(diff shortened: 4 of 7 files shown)


whitlockjc:
I am trying to use UAA as an authentication provider for Kubernetes.  Kubernetes supports OIDC as one of the mechanisms for this and since OIDC is implemented on top of OAuth 2, I figured it was worth a shot.  How I got to the point where the JWT `iss` claim became a problem is documented below.

Let's say I have UAA running at `https://uaa.notyou.com`.  To work around UAA not implementing OIDC connect, I created the following file and served it at `https://uaa.notyou.com/.well-known/openid-configuration`:

```json
{
  "issuer":"https://uaa.notyou.com",
  "authorization_endpoint":"https://uaa.notyou.com/oauth/authorize",
  "token_endpoint":"https://uaa.notyou.com/oauth/token",
  "jwks_uri":"https://uaa.notyou.com/k8token_keys",
  "response_types_supported":[
    "code"
  ],
  "grant_types_supported":[
    "authorization_code",
    "client_credentials"
  ],
  "subject_types_supported":[
    "public"
  ],
  "id_token_signing_alg_values_supported":[
    "RS256"
  ],
  "token_endpoint_auth_methods_supported":[
    "client_secret_basic"
  ],
  "claims_supported":[
    "aud",
    "email",
    "exp",
    "user_name",
    "iat",
    "sub"
  ]
}
```

**Note:** You'll notice that `jwks_uri` points to `https://uaa.notyou.com/k8token_keys` instead of `https://uaa.notyou.com/token_keys`  and that is because of #424.

With this document configuration, Kubernetes wires up to UAA via OIDC fine but when it comes to me validating my JWT to authenticate against Kubernetes, I get this error:

```
Failed to fetch provider config, trying again in 3s: "issuer" in config (https://uaa.notyou.com/oauth/token) does not match provided issuer URL (https://uaa.notyou.com)
```

The gist of this is that Kubernetes is configured with an issuer URL, which we set to `https://uaa.notyou.com`, and it verifies the `issuer` in the OIDC discovery document against that value.  If they don't match, Kubernetes isn't happy.  _(In our example case, that isn't an issue but I wanted to mention it.)_  Then what happens is when the JWT token is verified, it checks that the `iss` claim URL matches the issuer URL.  This is where we have a problem because UAA hard codes the appending of `oauth/token` to the issuer URL, even if you set the issuer URL in the UAA manifest yourself.

When I looked into UAA, I found out that not only can I turn it off but the only workaround is to disable the `iss` claim globally as this isn't an option at the OAuth client application level.  Not only that but you cannot work around this because setting the `issuer` in the UAA manifest ends up with `oauth/token` to that value as well.

whitlockjc:
Also, I realize this appears to be specific to Kubernetes but it's not.  I tried a few different OIDC client libraries and they all bailed due to the constraints mentioned above.  This is more about OIDC compliance.

whitlockjc:
Where we're at is we need to figure out what the spec says.  While it is pretty clear that the `issuer` value in the OpenID Provider Metadata document _(https://openid.net/specs/openid-connect-discovery-1_0.html#rfc.section.3)_ **must** match the JWT `iss` claim _(https://openid.net/specs/openid-connect-discovery-1_0.html#rfc.section.3)_, what isn't clear is if the `issuer` value must be the same as the OIDC discovery url without the discovery subpath _(`/.well-known/openid-configuration`)_.

whitlockjc:
Speaking with some of the `go-oidc` authors, people more well versed in OIDC than myself, found this: https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationValidation  Here is what it says:

> The issuer value returned MUST be identical to the Issuer URL that was directly used to retrieve 
> the configuration information. This MUST also be identical to the iss Claim value in ID Tokens
> issued from this Issuer.

I think it's pretty clear now that the location of the `/.well-known/openid-configuration` and the value of the `issuer` property within the `/.well-known/openid-configuration` document must be the same.  And those values must also match the `iss` claim in the JWT token.

At this point, I think that this is pretty clear but if I'm wrong, please let me know.

whitlockjc:
One thing I'm not aware of is the impact of this change on non-OIDC JWT tokens.  I think we need to figure this out.  I have some ideas on how we could support this if there is in fact an issue.

fhanik:
> The issuer value returned MUST be identical to the Issuer URL that was directly used to retrieve the configuration information.

Suggested URL
>"issuer":"https://uaa.notyou.com"

URL that " was directly used to retrieve the configuration information"

    https://uaa.notyou.com/.well-known/openid-configuration

this does present a new definition of "identical" doesn't it? :)

whitlockjc:
hahaha I'm just the messenger.

whitlockjc:
I found an even better reference in the document:

> OpenID Providers supporting Discovery MUST make a JSON document available
> at the path formed by concatenating the string `/.well-known/openid-configuration` 
> to the Issuer.

This is mentioned here: https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig

fhanik:
Meaning if we decide to support discovery we would have 

https://uaa.domain/oauth/token/.well-known/openid-configuration

is that what it says?

we're still looking for the dictation what the issuer URL should be.

thanks for bringing all this info to us

whitlockjc:
That's what I read based on the last piece of information I shared.  Here are the various constraints/rules that we have identified, and their location in the OpenID Provider Metadata document used using OpenID Connect Discovery:

* The _Issuer_ is the portion of the OpenID Connect Discover URL prior to the `/.well-known/openid-configuration` _(Example: If we serve the OpenID Provider Metadata at `https://localhost:8080/uaa/.well-known/openid-configuration`, _Issuer_ would be `https://localhost:8080/uaa`)_: <https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig>
* The `issuer` property in the OpenID Provider Metadata document _(located at the OpenID Connect Discovery URL)_ must be the same value as _Issuer_ identified in the first rule: <https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationValidation>
* The `iss` claim _(when provided)_ must be the same value as _Issuer_ identified in the first rule: <https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata>

I hope this helps.

whitlockjc:
When I get back from vacation, I'll get the CLA signed and update the PR to fix the merge conflicts.

whitlockjc:
I've submitted a CLA and I've updated the PR based on the latest `master` branch.

whitlockjc:
@fhanik Do I need to rebase this from `develop` instead of `master`?  I know when working on #282, you suggested branching from `develop`.

sreetummidi:
Closing this as this functionality has been implemented since UAA 3.9.0

PR #443, opened 2016-09-15
Add SCIM Patch Method for Users and Groups

tack-sap:
This pull request provides an implementation of the SCIM Patch Method for Users and Groups. 
Additionally it includes the corresponding Unit tests and documentation.


fhanik:
I'm working on this right now. There are a few modifications that I will be making

* reverse the `patch` method

    ScimUser.patch(ScimUser patch)

The object to be patched is invoked with the patch, not the other way around.
So in the code it will look like this

    oldUser.patch(userWithPatchData)

* The very long `if else` sequence will be changed to a `switch statement

* Tests should yield 100% code coverage

fhanik:
PR has been merged. 
https://github.com/cloudfoundry/uaa/commit/48566c972610fb82f008cc49d694123f052b383d

Thank you for the contribution

PR #457, opened 2016-10-04
[Henry, Dario, Brittany] Added client lockout functionality, fixed er…

6palace:
…rors in ExternalIdentityProviderDefinition and UaaAuthentication

Signed-off-by: Dario <amiri@ge.com>


model/src/main/java/org/cloudfoundry/identity/uaa/provider/ExternalIdentityProviderDefinition.java
```diff
@@ -41,7 +43,7 @@
     }
 
     public void setExternalGroupsWhitelist(List<String> externalGroupsWhitelist) {
-        this.externalGroupsWhitelist = new LinkedList<>(externalGroupsWhitelist!=null ? externalGroupsWhitelist : Collections.EMPTY_LIST);
+        this.externalGroupsWhitelist = new LinkedList<>(externalGroupsWhitelist!=null ? externalGroupsWhitelist : emptyList());
```
fhanik: Can we leave out refactoring like this to minimize the diff to be reviewed.

model/src/main/java/org/cloudfoundry/identity/uaa/provider/ExternalIdentityProviderDefinition.java
```diff
@@ -50,7 +52,7 @@ public void addWhiteListedGroup(String group) {
     }
 
     public void setAttributeMappings(Map<String, Object> attributeMappings) {
-        this.attributeMappings = new HashMap<>(attributeMappings!=null?attributeMappings:Collections.EMPTY_MAP);
+        this.attributeMappings = new HashMap<>(attributeMappings!=null?attributeMappings: emptyMap());
```
fhanik: Can we leave out refactoring like this to minimize the diff to be reviewed.

model/src/main/java/org/cloudfoundry/identity/uaa/zone/IdentityZoneConfiguration.java
```diff
@@ -24,6 +25,7 @@
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class IdentityZoneConfiguration {
 
+    private LockoutPolicy clientLockoutPolicy = new LockoutPolicy();
```
fhanik: Isn't null the correct backwards compatible value? That way, it can fall back on a global policy.

server/src/main/java/org/cloudfoundry/identity/uaa/audit/JdbcFailedLoginCountingAuditService.java
```diff
@@ -52,6 +52,16 @@ public void log(AuditEvent auditEvent) {
                                                 - saveDataPeriodMillis));
                 super.log(auditEvent);
                 break;
+            case ClientAuthenticationSuccess:
```
fhanik: Unit test is missing in `JdbcFailedLoginCountingAuditServiceTests`

fhanik:
Thank you. Please add unit tests for all new code. We follow TDD, that means any component that changes or is added, we do our best to always have a test that fails before the functional commit is added.

model/src/main/java/org/cloudfoundry/identity/uaa/zone/IdentityZoneConfiguration.java
6palace: Like tokenPolicy, clientLockoutPolicy defaults empty values to -1. The LockoutPolicyRetriever returns the global policy in getLockoutPolicy() when it finds an empty lockout policy in the identity zone config, so there is a fallback.

server/src/main/java/org/cloudfoundry/identity/uaa/audit/JdbcFailedLoginCountingAuditService.java
6palace: Okay, will do. Should the revisions be in a separate commit or should I tack it onto the existing commit?

fhanik: The preferred method is all tests in the first commit - tests go red. This would limit the number of changes we have to look through.

The second commit, would make all tests go green. We should then see an increase in test coverage, rather than a drop.

6palace: Okay, sounds good. Will reorder commits and repush incrementally.

6palace:
Hi @fhanik, requested changes are completed.

fhanik:
Still seeing coverage decreases, meaning we're adding in untested code. We're also confused how the identity provider has anything to do with client lockout?

I'm going to look into this right now, I may squash it into one commit (retain author)

6palace:
Hi Filip,
So the identity provider config includes a user lockout policy, and its associated classes, which is used by the UAA identity provider. We generalized the user lockout policy for client lockout, including some interface changes. 
I will look into coverage soon.

fhanik:
Closed via https://github.com/cloudfoundry/uaa/commit/6ce445630cd94fbdf4240ad8cf01dbca19eb7c17

PR #461, opened 2016-10-07
Fix MemberNotFoundException after external authorizations

mikeroda:
The delete of existing group memberships when receiving an ExternalGroupAuthorizationEvent can cause a MemberNotFoundException when multiple threads are authenticating the same user. This can happen because one thread has already done the delete and is adding the external group memberships while another thread is deleting them. The first thread does a get on each group membership it adds and that get will fail with an exception if the row doesn't exist.

This isn't too difficult to happen under load because this code path is invoked not just for web logins but also when obtaining a token using a password grant.

This change fixes the problem by only deleting and adding any changes to the group memberships. Thus after the first login, there will rarely be deletes and inserts, which is more efficient also.


plfx:
Hello, @mikeroda 

Thank you for identifying the issue. We cannot accept this pull request without a test demonstrating the failure before the change. Additionally, our existing test suite is failing. You can view these failures here. https://travis-ci.org/cloudfoundry/uaa/builds/170853971

If you would like to make a new pull request with tests added, we would happily look at it. Otherwise, we will eventually come around to this issue.

Issue #463, opened 2016-10-12
/check_token endpoint can be optimized as far as indexing and number of SQL queries


ilipkind:
We noticed that some of the SQL queries get called as part of /check_token endpoint taking ~ 200 ms. per query and probably can be indexed based on search field and hopefully also number of SQL calls per transaction (especially repetitive calls) can be reduced. Now it is about 30 SQL queries per transaction.
![screen shot 2016-10-11 at 4 44 10 pm](https://cloud.githubusercontent.com/assets/11415583/19292934/acedf960-8fd4-11e6-93d5-7bf8a405280f.png)
![screen shot 2016-10-11 at 4 44 36 pm](https://cloud.githubusercontent.com/assets/11415583/19292937/b3041104-8fd4-11e6-91b8-0533084b8097.png)
![screen shot 2016-10-11 at 4 44 58 pm](https://cloud.githubusercontent.com/assets/11415583/19292940/b7f9019c-8fd4-11e6-9e95-89e3b769c9f8.png)

mwdb:
We identified the same issue.
Since UAA version 3.X is supporting nested groups. This causes the number of SQL statements to the group_membership table to grow with the numbers of scopes assigned.

We have scenarios where 100 scopes are assigned to a user. During issuing of a token, the JdbcUaaUserDatabase.getUserById is called multiple times. For every call, it:
1) reads information from the user table
2) reads the 100 scopes with one SQL statement from the group,group_membership table
3) For each of the 100 groups, makes a new call to the DB to check if nested groups exist.

This issue is 3), which was introduced with the nested groups. 
We'll make a contribution to optimize 3 by querying multiple groups memberships in one query.

This brings down the number of DB queries for a user with 100 scopes from 102 to 3.
 JdbcUaaUserDatabase.getUserById

We're working on contributing this change.

mwdb:
See PR https://github.com/cloudfoundry/uaa/pull/471

PR #471, opened 2016-10-21
Change getAuthorities to reduce number of executed SQL-Statements

tack-sap:
Any time a user is retrieved using this class, the getAuthorities method is executed to retrieve all of the user's direct and indirect group memberships recursively. 
As mentioned in #463 by my colleague @mwdb the number of SQL queries used to do this grows with the number of scopes assigned.
As the methods are called very frequently, this highly increases the load on the database. With this PR the number of SQL queries is reduced. According to our performance tests this highly increases the performance of the database.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/user/JdbcUaaUserDatabase.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/user/JdbcUaaUserDatabase.java
@@ -13,6 +13,7 @@
 package org.cloudfoundry.identity.uaa.user;
 
 
+import org.apache.commons.lang.ArrayUtils;
 import org.cloudfoundry.identity.uaa.zone.IdentityZoneHolder;
 import org.springframework.dao.EmptyResultDataAccessException;
 import org.springframework.dao.IncorrectResultSizeDataAccessException;
@@ -27,6 +28,7 @@ import org.springframework.util.StringUtils;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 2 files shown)


fhanik:
Thank you for the PR

It appears that we have a coverage decrease here. Meaning, not all lines of code added were exercised by the tests. We prefer code changes that are test driven. https://en.wikipedia.org/wiki/Test-driven_development

It may take longer to merge a PR, when we have to write supplementary tests or try to figure out what lines are not tested in this feature.

tack-sap:
Hi Filip,

With the added test all lines of the methods we changed are covered by tests (according to our tooling).

When we tried to cover all lines, we found, that the return in the catch-block was never executed by tests. The catch-block itself is taken from the original coding, but is empty there. We added the return because our guidelines do not allow empty catch-blocks.
The catch-block is never executed, because queryForList does not throw this exception on an empty result, but returns an empty list (which is why the catch was never used in the previous coding as well).
As there is no other catch, we removed then the try-catch from the method.

tack-sap:
Is there a reason why this PR is still "unscheduled" after such a long time?
All checks have been passed, the coverage report says "coverage increased", there a no conflicts and I got no feedback left, that some further action is required (as the feedback I got was already addressed with the second commit)..
Is there anything I can do for this PR to get processed?

sreetummidi:
We will get this scheduled for UAA 4.0

fhanik:
@tack-sap Thank you for your contribution. It is running through.

I corrected your test. In a Test Driven Development process, the test should fail before a change is applied, and succeed after. Your test succeeded before and after, thus the test does not detect a change in the code base.

I have modified your test - there is a neat trick using the `Mockito.spy` proxy

https://github.com/cloudfoundry/uaa/commit/eb0a970283cb8bf0ebf0d0b97308c73ca816ca5d

fhanik:
Merged via https://github.com/cloudfoundry/uaa/commit/d4a9baae31f97238cc0e02154e55816ed3f3854c

PR #487, opened 2016-11-10
Remove null check on UserAuthenticationFailureEvent

davbo:
If this null check was ever true the `user.getUsername()` call would
raise a NullPointerException.

These events only occur when user passwords don't match, more precisely
the event is only created here:

    org/cloudfoundry/identity/uaa/authentication/manager/AuthzAuthenticationManager.java:107

In this case the user object has already been null checked.


davbo:
One of the test runs seems to have failed. We just looked into the failure and it appears to be caused by this assertion https://github.com/cloudfoundry/uaa/blob/master/server/src/test/java/org/cloudfoundry/identity/uaa/scim/ScimCoreTests.java#L54 

The `new Date()` calls on ScimMeta construction could happen milliseconds apart (the accuracy of Date) this looks likely to be the cause of this failure and makes the test unreliable.

I don't know the code very well but given that ScimCore [doesn't implement `patch()`](https://github.com/cloudfoundry/uaa/blob/master/model/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimCore.java#L81) I'm not sure what the test is testing.

sreetummidi:
PR to be merged

PR #489, opened 2016-11-23
Feature: Add saml2 bearer grant 

strehle:
Add grant type urn:ietf:params:oauth:grant-type:saml2-bearer

Changes:
* Add Saml2TokenGranter as extension for custom grant types
 * server implementation of the saml-bearer flow
 * pack the assertion from client into a SAMLResponse and use the standard SAML login
 * use the standard authentication so that UAA issues a token as with other grant types
 * add saml grant type to UaaTokenService
* Unit Tests: Saml2TokenGranterTest
* Added one positive and on negative integration tests in SamlLoginWithLocalIdpIT to ensure code coverage


fhanik:
Thank you for the PR, can you supplement it with an integration test that works successfully? I see one that results in 401, but not one that demonstrates the end to end flow.

strehle:
Yes I am working on it. The creation of the SAML assertion is something I currently analyse. I first created a long term assertion and wanted to add this as hard coded assertion. however spring security has a hard coded life-time so that this approach does not work but I have to create an assertion in integration test

strehle:
hi , I added 
org.cloudfoundry.identity.uaa.integration.feature.SamlLoginWithLocalIdpIT > testValidSaml2Bearer PASSED
org.cloudfoundry.identity.uaa.integration.feature.SamlLoginWithLocalIdpIT > testInvalidSaml2Bearer PASSED

as positive and negative tests

sreetummidi:
Hi Marcus @strehle 

Thank you for the PR. We will review and merge for the next cycle.

Thanks,
Sree

strehle:
HI @fhanik , you can change my PR of course. I know that there are code duplication which could be prevented with a factor. I mentioned this in my comments. I dont want to change your existing code but let you decide howto proceed and howto refactor this.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/token/Saml2TokenGranter.java
```diff
@@ -0,0 +1,500 @@
...
+ *     Copyright (c) [2009-2016] SAP SE. All Rights Reserved.
+ *
+ *     This product is licensed to you under the Apache License, Version 2.0 (the "License").
+ *     You may not use this product except in compliance with the License.
+ *
+ *     This product includes a number of subcomponents with
+ *     separate copyright notices and license terms. Your use of these
+ *     subcomponents is subject to the terms and conditions of the
+ *     subcomponent's license, as noted in the LICENSE file.
+ *******************************************************************************/
```
strehle: I simply took it from our templates. If not OK, you can change it. It should only be clear that it comes from SAP ;)

fhanik:
Let me go through it and see what duplication can be removed. Looking at the password granter, it doesn't do anything with scopes

fhanik:
@strehle Working on a serverless test case for better testing coverage

https://github.com/cloudfoundry/uaa/commit/cdf7379c703765e9bfcafca28ad972c1bb4387e1#diff-f7a9f1a69eec2ce4278914f342d8a160R321

Currently failing in subject validation. Something going awry. If you have time, and know what the issue is with the mock assertion, I would gladly appreciate it

fhanik:
@strehle Ignore previous message. I got a green test to work with. Some hard coded values have been moved out.

fhanik:
@strehle This is what I am reworking. The request manager expects authentication to happen before hand if it is available in the request. This happens in the token endpoint authentication filter

https://github.com/cloudfoundry/uaa/blob/37372e80d24794ef8bb1cc40b7c13dd946230064/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/BackwardsCompatibleTokenEndpointAuthenticationFilter.java#L60

Basically, this filter will do the authentication based on the assertion, very similarly we do this in the 
`org.springframework.security.saml.SAMLProcessingFilter` 

            SAMLMessageContext context = contextProvider.getLocalEntity(request, response);
            processor.retrieveMessage(context);

            // Override set values
            context.setCommunicationProfileId(getProfileName());
            context.setLocalEntityEndpoint(SAMLUtil.getEndpoint(context.getLocalEntityRoleMetadata().getEndpoints(), context.getInboundSAMLBinding(), context.getInboundMessageTransport()));

            SAMLAuthenticationToken token = new SAMLAuthenticationToken(context);
            return getAuthenticationManager().authenticate(token);

I'm not fully clued in on the https://github.com/cloudfoundry/uaa/blob/3ed21343704feeae1b7f8c66a4dbf0f67e93fa71/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/token/Saml2TokenGranter.java#L203 method, and why it does so many things. That can probably be delegated to the above filter too

fhanik:
@strehle  Still working on this, here is the refactor without test cases

https://github.com/cloudfoundry/uaa/commit/4bac78608e4973ab4a7a5aa71b471cf863525b16

It has the following benefits

1. The metadata lists the correct endpoint for the posting

        <md:AssertionConsumerService 
            Binding="urn:oasis:names:tc:SAML:2.0:bindings:URI"
            Location="http://localhost:8080/uaa/oauth/token/alias/cloudfoundry-saml-login"
            index="2"/>
2. It performs the same validation as a regular SAML message
3. We plug into the existing Spring library using the [SamlAssertionBinding](https://github.com/cloudfoundry/uaa/blob/4bac78608e4973ab4a7a5aa71b471cf863525b16/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/SamlAssertionBinding.java#L39-L39) class, and thus inherit all the XML logic
4. We are able to do server less testing using [MockMvc](https://github.com/cloudfoundry/uaa/blob/4bac78608e4973ab4a7a5aa71b471cf863525b16/uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/token/TokenMvcMockTests.java#L325-L388).

I should have everything completed in the next couple of days.

strehle:
Hi Filip, my first prototypes were much larger, so if you can reduce the code, it good. Therefore I only submitted the bearer flow for users. Later we also want push the grant type urn:ietf:params:oauth:client-assertion-type:saml2-bearer but we will refactor it based on this

fhanik:
Thank you @strehle 

I believe I have removed all duplicate code now. Which made the Saml2TokenGranter class quite small, and easier to test.

I have one last commit, for the documentation to work on Travis now. I expect to merge this to develop today. Your original commit will be preserved, and the refactoring is done through later commits.

fhanik:
Merged and completed via https://github.com/cloudfoundry/uaa/commit/81a2045b7a62b9239f20c43ef76e0d061aeaccc4

Thank you for the contribution.

PR #491, opened 2016-11-25
Feature/rememberme

scizeron:
Here is a remember me implementation, based on a hashed token.
The remember feature is enabled for the login form and can be enabled for the saml providers.
When the feature is enabled for SAML, when the SAML authentication response is succeed, the cookie will  be set automatically.


scizeron:
I leverage the remember me feature when performing OAuth 2.0 authorization requests. One of my updates is to add IS_AUTHENTICATED_REMEMBERED concerning /oauth/authorize requests : 

<intercept-url pattern="/oauth/authorize" access="IS_AUTHENTICATED_REMEMBERED,IS_AUTHENTICATED_FULLY" />

sreetummidi:
Please see my comments on the related story

PR #496, opened 2016-12-05
Add table of contents to MD files

pradtke:
A lot of the markdown files are quite lengthy. It is nice to have a ToC to get an overview of what is in it. I used `doctoc` to generate them.
Not sure if this the way the project wants to go, but it would be nice to have auto-generated ToCs.


sreetummidi:
We have plans to consolidate all docs under doc.cloudfoundry.org. This in the in the process of getting reviewed and will be published soon

PR #514, opened 2017-01-05
Extend the scope related fields as we have exceeded the existing length

martinvaller:


fhanik:
This PR can't be merged at the moment. It does not address the other database types
Correct solution may have been to use groups like we do with users

fhanik:
@martinvaller have you considered using wildcards on the clients?

martinvaller:
@fhanik The use case is that we need the scopes in the Javascript UI and therefore all the user scopes have to be present in the JWT token. Would it be ok if we increase the size of the fields mentioned in the pull request?

fhanik:
@martinvaller friendly reminder that pull requests should be against `develop` branch.

PR #518, opened 2017-01-09
ScimUserBootstrap: Selective Update on Groups

tack-sap:
This PR addresses the same issues as #461 
We frequently experienced the same issue described there, as well as tokens with incomplete scopes, due to concurrent modification of the groups.
This is how we solved this issues. We were preparing it as PR, but #461 was quicker, so we decided not to make the PR at that time. Since the other PR was not accepted, because of failing and missing tests, we decided to make a PR out of our solution.
I also included a test, which shows the MemberNotFoundException with the old implementation. As the Exception is caused by concurrent use of the method, I could not guarantee a failure in the test. Therefore I used 100 Groups and 5 Threads, each one calling onApplicationEvent 50 times, to have a higher probability of failure. With this configuration of the test it failed every time on my machine using the old implementation.


server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupMembershipManager.java
```diff
@@ -502,4 +529,24 @@ public ScimGroupMember mapRow(ResultSet rs, int rowNum) throws SQLException {
         }
 
     }
+
+    private static final class ScimGroupRowMapper implements RowMapper<ScimGroup> {
```
fhanik: this seems to be a duplicate class. Each time we change the scim group we would be bound to change it in two places.
We will review and optimize if needed.

sreetummidi:
@tack-sap Could you please resubmit on a clean branch as this one contains more than the intended fix. Please use the latest develop as the baseline.

PR #525, opened 2017-01-27
Fix/zone aware client lockout

6palace:
client lockout policy was not zone aware, so clients with the same client id across zones would share lockout counters and times. This pull request fixes it.
To replicate: checkout master, start uaa, add 2 identity zones. Attempt to get admin token 5 times in zone 1 with wrong credentials, get locked out of admin client. target zone 2, attempt to get admin token with correct credentials, admin will still be locked out.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/audit/JdbcAuditService.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/audit/JdbcAuditService.java
@@ -43,6 +43,11 @@ public class JdbcAuditService implements UaaAuditService {
                         "principal_id=? and created > ? order by created desc", new AuditEventRowMapper(), principal,
                         new Timestamp(after));
     }
+    
+
+    public List<AuditEvent> find(String principal, long after, String IdentityzoneId) {
+        return find(principal, after);
+    }
 
     @Override
     public void log(AuditEvent auditEvent) {
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/audit/JdbcAuditService.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/audit/JdbcAuditService.java
@@ -12,6 +12,7 @@
  *******************************************************************************/
 package org.cloudfoundry.identity.uaa.audit;
 
+import org.cloudfoundry.identity.uaa.zone.IdentityZoneHolder;
 import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.jdbc.core.RowMapper;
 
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/cloudfoundry/identity/uaa/audit/JdbcAuditService.java
```diff
@@ -38,10 +38,10 @@ protected JdbcTemplate getJdbcTemplate() {
     }
 
     @Override
-    public List<AuditEvent> find(String principal, long after) {
```
fhanik: I don't think we need a name and a signature change. That's what is causing this very large diff.

    public List<AuditEvent> find(String principal, long after) {
         return find(principal, after, IdentityZoneHolder.get().getId());
    }

    //new method - do your changes here. 
    public List<AuditEvent> find(String principal, long after, String zoneId) {
    }

We drive this project through TDD. For PRs we're asking a bit more discipline regards that so we can validate the PR itself. 2 commits. The first commit introduces your test (and any interface changes needed). The test should fail when only the first commit is applied. The second commit implements the change, and the test passes when applied.

server/src/test/java/org/cloudfoundry/identity/uaa/audit/JdbcAuditServiceTests.java
```diff
@@ -41,7 +41,7 @@ public void userAuthenticationFailureAuditSucceeds() throws Exception {
         auditService.log(getAuditEvent(UserAuthenticationFailure, "1", "joe"));
         Thread.sleep(100);
         auditService.log(getAuditEvent(UserAuthenticationFailure, "1", "joe"));
-        List<AuditEvent> events = auditService.find("1", 0);
+        List<AuditEvent> events = auditService.findInZone("1", IdentityZone.getUaa().getId(), 0);
```
fhanik: This change is no needed. And is useful to ensure that the system works correctly with the method override.

fhanik:
@6palace Please also ensure that you are working off the latest develop branch. For some reason Travis shows that you are still supplying old .travis.yml indicating that you're working on an out of date branch

https://travis-ci.org/cloudfoundry/uaa/builds/195966964

fhanik:
@6palace Can you please also follow these instructions

https://help.github.com/articles/publicizing-or-hiding-organization-membership/

so that we can know that you're contributing this as part of your organization (using the organization CLA on file)

6palace:
Alright, changes sound good. About the 2 commit system, I do have separate test and implementation commits, is your point that you want me to push them one by one instead of together so the tests run twice?

fhanik:
@6palace the first commit must compile and the test be runnable. So if there is an interface change, that would go in the first commit. Otherwise I can't run a test and ensure it fails.

fhanik:
@6palace here is how I would go about pulling in this PR

1. I cherry pick the first commit, run the tests and ensure the test I expect to fail actually fails
2. I cherry pick the second commit, now all tests should pass

Hope this helps

PR #529, opened 2017-01-31
Fix circular view path errors on login

mikeroda:
The login endpoint should return text/html by default if no Accept header is specified, or if the client specifies */*. However, requests to UAA with an unexpected media type in the Accept header can get redirected to the login page and will result in a 500 status code and the following error:

Circular view path [login]: would dispatch back to the current handler URL [/uaa/login] again. Check your ViewResolver setup! (Hint: This may be the result of an unspecified view, due to default view name generation.)

This is because the loginForHtml endpoint accepts */* but when spring goes to render the page it can't find an appropriate view. The fix to this is to take an additional parameter with the Accept media types
and check to see if it is compatible.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java
@@ -41,6 +41,8 @@ import org.cloudfoundry.identity.uaa.zone.IdentityZoneConfiguration;
 import org.cloudfoundry.identity.uaa.zone.IdentityZoneHolder;
 import org.springframework.core.io.support.PropertiesLoaderUtils;
 import org.springframework.dao.EmptyResultDataAccessException;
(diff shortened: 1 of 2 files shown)


fhanik:
@mikeroda Thanks for the PR - there are some failing test cases

https://api.travis-ci.org/jobs/197102923/log.txt?deansi=true

mikeroda:
Is there a way to rerun the build? Looks like an unrelated failure.

fhanik:
Restarted it for you
https://travis-ci.org/cloudfoundry/uaa/builds/197166789

fhanik:
We do get false negatives quite often....test suite has turned into somewhat of a behemoth

sreetummidi:
Additional tests need to be added to validate the response code

fhanik:
Closing via https://github.com/cloudfoundry/uaa/commit/ceb8b72299cadc4b64bd1fc872f8527d7e3da404

PR #531, opened 2017-02-03
Fix MemberNotFoundException after external authorizations

mikeroda:
The delete of existing group memberships when receiving an ExternalGroupAuthorizationEvent can cause a MemberNotFoundException when multiple threads are authenticating the same user. This can
happen because one thread has already done the delete and is adding the external group memberships while another thread is deleting them. The first thread does a get on each group membership it adds and that get will fail with an exception if the row doesn't exist.

This isn't too difficult to happen under load because this code path is invoked not just for web logins but also when obtaining a token using a password grant.

This change fixes the problem by only deleting and adding any changes to the group memberships. Thus after the first login, there will rarely be deletes and inserts, which is more efficient also.


sreetummidi:
@mikeroda Can you provide the tests for this PR and re-do the PR against the latest of develop. Please remove the use of Scim Filters as we are trying to move away from its usage in favor of explicit filter conditions.

mikeroda:
Resolved by ec8a84f.

PR #545, opened 2017-02-20
Add origin to cors allowed headers

samu:
Safari sends `origin` in the list of required headers when doing a CORS preflight request. That request fails with a 403 status code when using the default UAA configuration. The problem has been reported in more detail [here](http://stackoverflow.com/questions/37557364/cloudfoundry-uaa-preflight-origin-header-not-allowed-in-safari).

This PR adds the `origin` header to the `corsAllowedHeaders` configuration directive. I do understand one could overwrite this config, however we are currently using the default bosh release repo for deployment and configuring this ourselves is inconvenient. I think this setting will be useful for anybody needing to talk to CF through a browser.


samu:
Before i attempt to fix the failing builds: can somebody review this and give feedback whether this change would be merged? Thanks 👍

bsekar:
@samu  With reference to [this](https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Preflighted_requests) documentation, `oauth/token` AJAX call should not require CORS preflighting. I'm wondering if the 403 is because of other custom headers that you may be sending. We were not able to reproduce the issue where UAA responds with 403 when the `origin` is sent as part of request header

samu:
The problem is resolved with the latest version of Safari (`10.1.1 `).

PR #549, opened 2017-02-25
Change idp discovery flow to fallback to zone login page.

Bitty9:
- When login_hint is provided and it does not match an idp, redirect to login page of that zone, instead of email page.
- When idp discovery is enabled and the email domain does not match an idp, redirect to login page of that zone instead of email page.

Sometimes the username and email of a user do not match and the current implementation assumes that they are always the same.


sreetummidi:
We have plans to address this by prompting the user for a username if the username and email are not the same. We will do so by updating the page which collects the password only today

PR #550, opened 2017-02-27
Add index on zone_id column for the table revocable_tokens

RoopGuron:


server/src/main/resources/org/cloudfoundry/identity/uaa/db/postgresql/V3_10_4__Add_Index_On_Identity_Zone_To_Revocable_Tokens.sql
```diff
@@ -0,0 +1,2 @@
+-- create an index to match the query
+CREATE INDEX revocable_tokens_zone_id ON revocable_tokens(LOWER(identity_zone_id));
```
fhanik: Which query are we matching? The queries accessing this are not using lower(). 

https://github.com/cloudfoundry/uaa/blob/7fffffd466ebf21cb1148f1808c92a29d0597a68/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/token/JdbcRevocableTokenProvisioning.java#L41-L49

fhanik:
Thank you for the PR @RoopGuron 

Can you confirm the `lower` function on PostgreSQL. I don't believe we are using this function when accessing this table/zone_id

server/src/main/resources/org/cloudfoundry/identity/uaa/db/postgresql/V3_10_4__Add_Index_On_Identity_Zone_To_Revocable_Tokens.sql
RoopGuron: Thanks for pointing it out. 
Postgresql flyway script has been changed to remove the `lower` function

RoopGuron:
Thanks @fhanik for pointing out, the queries are actually not using `lower`.
I have made the changes and pushed those. Please review not.

RoopGuron:
Please review Now.
Thanks

fhanik:
Merged via https://github.com/cloudfoundry/uaa/commit/4c99deaab94b4731ee8401888e9ff3d020d278df

PR #556, opened 2017-03-08
Support long scopes 3.12.0

RoopGuron:
Extend Client authorities and scope columns.


fhanik:
Thank you for the contribution. Can you add a test case to the PR?

An example can be found here:
https://github.com/cloudfoundry/uaa/blob/develop/server/src/test/java/org/cloudfoundry/identity/uaa/db/ClientDetailsHasRequiredUserScopes.java#L85-L86

PR #566, opened 2017-03-16
Encode cookies to be compliant with Tomcat 8.5x

### LINKED ISSUE #563: Cookie Saved-Account-XXX is not encoded.
Hi, 
we tested UAA on tomcat 8.5x and there is an exception, because of a cookie character validation.
In detail the problem we see is in 
https://github.com/cloudfoundry/uaa/blob/master/server/src/main/java/org/cloudfoundry/identity/uaa/login/AccountSavingAuthenticationSuccessHandler.java#L63
where Cookie "Saved-Account- + uaaPrincipal.getId()" is not encoded.

In same class there is an example of correct cookie encoding.
https://github.com/cloudfoundry/uaa/blob/master/server/src/main/java/org/cloudfoundry/identity/uaa/login/AccountSavingAuthenticationSuccessHandler.java#L77

Therefore please correct it in
https://github.com/cloudfoundry/uaa/blob/master/server/src/main/java/org/cloudfoundry/identity/uaa/login/AccountSavingAuthenticationSuccessHandler.java#L63

and on
https://github.com/cloudfoundry/uaa/blob/master/server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java#L249
where cookie is read .

thanks in advance,
-markus

strehle:
* fix #563
 * Encode Cookie Saved-Account-xx in AccountSaving
 * Decode Cookie Saved-Account-xx in LoginInfoEndpoint
 * Add test for invalid cookie

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java
@@ -89,6 +89,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.Collections.emptyMap;
 import static java.util.Objects.isNull;
 import static java.util.Optional.ofNullable;
@@ -246,8 +247,18 @@ public class LoginInfoEndpoint {
         return Arrays.asList(ofNullable(cookies).orElse(new Cookie[]{}))
                 .stream()
                 .filter(c -> c.getName().startsWith("Saved-Account"))
-                .map(c -> JsonUtils.readValue(c.getValue(), clazz))
+                .map(c -> JsonUtils.readValue(decodeCookieValue(c.getValue()), clazz))
                 .collect(Collectors.toList());
+    }
+
+    private static String decodeCookieValue(String inValue) {
+        String out = null;
+        try {
+            out = URLDecoder.decode(inValue, UTF_8.name());
+        } catch (Exception e) {
+            return "";
+        }
+        return out;
     }
 
     @RequestMapping(value = {"/invalid_request"})
@@ -613,7 +624,7 @@ public class LoginInfoEndpoint {
         }
 
         String base64Credentials = auth.substring("Basic".length()).trim();
-        String credentials = new String(new Base64().decode(base64Credentials.getBytes()), Charset.forName("UTF-8"));
+        String credentials = new String(new Base64().decode(base64Credentials.getBytes()), UTF_8.name());
         // credentials = username:password
         final String[] values = credentials.split(":", 2);
         if (values == null || values.length == 0) {
@@ -797,7 +808,7 @@ public class LoginInfoEndpoint {
     protected String extractPath(HttpServletRequest request) {
         String query = request.getQueryString();
         try {
-            query = query == null ? "" : "?" + URLDecoder.decode(query, "UTF-8");
+            query = query == null ? "" : "?" + URLDecoder.decode(query, UTF_8.name());
         } catch (UnsupportedEncodingException e) {
             throw new IllegalStateException("Cannot decode query string: " + query);
         }
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java
@@ -12,6 +12,8 @@
  *******************************************************************************/
 package org.cloudfoundry.identity.uaa.login;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.cloudfoundry.identity.uaa.authentication.AuthzAuthenticationRequest;
 import org.cloudfoundry.identity.uaa.authentication.UaaAuthentication;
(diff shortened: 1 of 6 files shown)


fhanik:
Thank you @strehle 

Got a few questions

1. What characters are causing the issue today?
2. What happens to existing non encoded cookies when this patch gets pushed to an environment?
3. Is URL encoding what we need for cookies? Is that the appropriate encoding?

strehle:
Hi @fhanik ,
as mentioned in https://github.com/cloudfoundry/uaa/issues/563 ,
There is an existing encode for Current-User, see
https://github.com/cloudfoundry/uaa/blob/master/server/src/main/java/org/cloudfoundry/identity/uaa/login/AccountSavingAuthenticationSuccessHandler.java#L77

But some lines above there is no encoding.
The invalid characters are {, ", space , etc. because currently
https://github.com/cloudfoundry/uaa/blob/master/server/src/main/java/org/cloudfoundry/identity/uaa/login/AccountSavingAuthenticationSuccessHandler.java#L63
writes raw JSON into value.

Cookies should be URLEncoded, because they belong the HTTP header.

Tomcat 8.0.x does not check these value and even with Tomcat 8.5x there is an option to deactivate the check, but I think UAA should encode all cookies. The change is small, even if build fails now and I have to check why.

regards,
-markus

fhanik:
> Cookies should be URLEncoded, because they belong the HTTP header.

https://tools.ietf.org/html/rfc6265#section-4.1.1

URL encoding works, but may be overkill. We can evaluate that later. For now...

I am still asking the same question:
What happens to existing, non encoded, cookies when this feature goes into effect?
Does it blow up, fail gracefully or continues to work as normal?

I will also make a note in a story to start evaluating using the HTML5 local storage instead of cookies.

strehle:
URL decode is done with this change, but the decode does not change unencoded values, e.g.
If you have an existing cookie value with {user:"bill"} then the URLDecode of this is always: {user:"bill"}

This change now encodes: {user:"bill"} to %7Buser%3A%22bill%22%7D, but 
URLDecode of :
a)  {user:"bill"}
b) %7Buser%3A%22bill%22%7D
is both {user:"bill"}

So answer to your question: works as normal.

The URL encoding of cookies is the rule here, therefore I dont know if overkill.
My other argument here is.
Why is 
https://github.com/cloudfoundry/uaa/blob/master/server/src/main/java/org/cloudfoundry/identity/uaa/login/AccountSavingAuthenticationSuccessHandler.java#L77
encoded and
https://github.com/cloudfoundry/uaa/blob/master/server/src/main/java/org/cloudfoundry/identity/uaa/login/AccountSavingAuthenticationSuccessHandler.java#L63
not.? 

And to HTML5 . of course if you move the cookies to other storages, go ahead, but now it is a cookie, therefore this PR. As mentioned, sooner or later you also will move to Tomcat 8.5 then you will see the problem.

server/src/test/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpointTests.java
```diff
@@ -204,6 +207,68 @@ public void testSavedAccountsPopulatedOnModel() throws Exception {
     }
 
     @Test
+    public void testSavedAccountsEncodedAndUnEncoded() throws Exception {
```
strehle: Hi @fhanik here is a test which ensures that before and after work as normal

server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java
```diff
@@ -246,10 +247,20 @@ public String loginForHtml(Model model, Principal principal, HttpServletRequest
...
+                .map(c -> JsonUtils.readValue(decodeCookieValue(c.getValue()), clazz))
                 .collect(Collectors.toList());
     }
 
+    private static String decodeCookieValue(String inValue) {
+        String out = null;
+        try {
+            out = URLDecoder.decode(inValue, UTF_8.name());
+        } catch (Exception e) {
+            return "";
```
fhanik: does `""` deserialize as `clazz`, or will an error here, return `""` cause the entire list to fail in the calling method?

and if that is the case, no logging on an exception here, not even debug, would cause us to lose the root cause of the issue, and become very hard to track down.

fhanik: can you put a `logger.debug` statement in the `catch` clause, so we don't lose the root cause of the error? That way I can merge the PR without having to modify it.

server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java
```diff
@@ -246,10 +251,21 @@ public String loginForHtml(Model model, Principal principal, HttpServletRequest
...
+                .map(c -> JsonUtils.readValue(decodeCookieValue(c.getValue()), clazz))
                 .collect(Collectors.toList());
     }
 
+    private static String decodeCookieValue(String inValue) {
+        String out = null;
+        try {
+            out = URLDecoder.decode(inValue, UTF_8.name());
+        } catch (Exception e) {
+            logger.debug("URLDecoder.decode failed for " + inValue, e);
```
strehle: ok ?

fhanik: perfect

PR #568, opened 2017-03-16
Fixing Logout Whitelist with Client Behavior

6palace:


sreetummidi:
Hi,
Can you please elaborate on what was not working before with logout which has been addressed with this. Logout URLs are matched against the redirect URI's set on the client. 

Thanks,
Sree

fhanik:
Thanks for the PR. I think it is over complicating something that can be very simple and more concise. See comments

6palace:
Previously, the logout redirect did exact matches only when considering the set of redirect uris. Since wildcards aren't allowed, zone configuration would have to add a url to whitelist for every app using that zone, and with no self-service zone specification, this was unscalable. This problem can be avoided if we use client_id to augment redirect uri. However, most application clients rely on the fact that paths are not matched in the authorization flow. Logout flow does not follow this, and so customers would need to add a specific /logout redirect uri to every client, again unscalable. This change just joins the authorize flow redirect behavior with the logout behavior so no tedious reconfiguration is necessary. If there's a simpler way to do this, We'd be happy to revise.

fhanik:
@sreetummidi @6palace 

It is still unclear from this implementation, which seems to change more than it needs, is opening the door to _open redirects_, which would mean we cannot accept it..  The PR is also against the master branch, needs to be against develop.

We have identified an inconsistency between logout redirect and oauth/authorize redirect
https://github.com/cloudfoundry/uaa/issues/579

fhanik:
We've opened this

https://github.com/cloudfoundry/uaa/issues/579

Closing this PR, it will not be merged at this time.

Issue #562, opened 2017-03-14
Empty `redirect_uri` and not working `redirectUrl` parameter when using OIDC provider


paroxp:
## Issue

UAA does not take `login.oauth.providers.<alias>.redirectUrl` configuration into consideration as [it is supposed to](https://github.com/cloudfoundry/uaa/blob/master/docs/google-oidc-provider.md).

When `redirectUrl` is set to `https://login.example.com/uaa` the [`redirect_uri`](https://github.com/cloudfoundry/uaa/blob/3.9.3/model/src/main/java/org/cloudfoundry/identity/uaa/provider/AbstractXOAuthIdentityProviderDefinition.java#L164) query parameter in the URL for our external provider (Google) is `https://login/callback/google`. The domain part is blank, as if an empty string was passed to the [getCompleteAuthorizationURI method](https://github.com/cloudfoundry/uaa/blob/3.9.3/server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java#L441).

## Expected Result

Google would require you to log in and then redirect back to the CF UAA. The redirect URI passed to Google should be `https://login.example.com/login/callback/google`.

## Actual Result

![](https://i.imgur.com/T2hwpvu.png)

## Versions

UAA releases/3.9.3 `565bfebb000034c84d7853d4f9d387a8fc84c4bb`
UAA releases/3.12.0 `011e1b86340db0b9e098e7d1eeaf4b0af8797360 `

jm4rc05:
[Here](https://docs.cloudfoundry.org/api/uaa/#oauth-oidc) the API doesn't have this key! 

Possibly the `http://localhost:8080/uaa` came from [other](uaa/uaa/src/main/resources/uaa.yml) place!

henrytk:
@jm4rc05 how does an external provider, such as Google, know where to redirect the user to after authentication?

bleach:
This is a bug in [GetBaseURL](https://github.com/cloudfoundry/uaa/blob/4e652a6b5279ea99077fcfc46c5b8842267806c8/server/src/main/java/org/cloudfoundry/identity/uaa/util/UaaUrlUtils.java#L103) 

When it's called from https://login.example.com/login, `getRequestPath == "/login"` so it returns `https:/`.

bleach:
[...]

PR #575, opened 2017-03-27
Fix BaseURL generation when hostname matches URL

bleach:
As reported in cloudfoundry/uaa#562, when the login hostname begins with login,
then generation of the redirect_uri sent to OIDC providers should look like
this before urlencoding:

https://login.domain/login/callback/ALIAS

instead it looks like this:

https://login/callback/ALIAS

Which causes the Google OIDC provider to reject the request due to a non-whitelisted redirect_uri.

This fixes that bug and also adds tests.


bleach:
Rebasing against `develop`

bleach:
Thanks @fhanik !

dcarley:
@fhanik can you give an indication of when this fix will be released?

The workaround of manually editing the callback URL in the browser is incredibly painful 🐼

PR #594, opened 2017-04-21
Switch reset password flow to use username instead of email as username

Bitty9:


sreetummidi:
To be reviewed and merged

PR #598, opened 2017-04-26
Adding oidc idp non-custom attribute mapping (e.g. given name, email)

6palace:


sreetummidi:
@6palace Can you please clarify the intent of the pull request. Is it to read the email and other attributes and set them even if attribute mappings have not been explicitly provided?

6palace:
@sreetummidi So oidc IDPs currently have two types of attribute mapping: username and custom attribute. However, other SCIM fields, specifically email, name, and phone number, cannot be mapped from incoming oidc claim to UAA user field. eg if an external oidc idp had a claim with name 'email_address' there was previously no way to map that to email.

sreetummidi:
To be reviewed further.

PR #601, opened 2017-04-27
Fixed query string (first param should be appended using ? instead of &) - fix #600

### LINKED ISSUE #600: Implicit flow: incorrect redirect URL (query params)
When using the implicit flow (in the browser) by redirecting the user to: 

http://localhost:7001/uaa/oauth/authorize?response_type=token&client_id=XXX&redirect_uri=http%3A%2F%2Flocalhost%3A8100%2F%23%2Flogin

The user gets redirected back after successful login to:

http://localhost:8100/#/login&token_type=bearer&access_token=eyJhbGci...

As the query parameters are added with an & instead of ? the authenticated application cannot get the query parameters correctly. Shouldn't the first query param be appended with an ? and only the following params with &

So a correct URL would be:

http://localhost:8100/#/login?token_type=bearer&access_token=eyJhbGci...

maust:
This will fix #600 

Would be interesting to know why nobody yet had this issue, as it seems that it is in there since 2015

Furthermore I removed boolean fragment as it is always true. If this is not wanted let me know and I revert it.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaAuthorizationEndpoint.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaAuthorizationEndpoint.java
@@ -347,7 +347,7 @@ public class UaaAuthorizationEndpoint extends AbstractEndpoint {
             }
             return new ModelAndView(
                 new RedirectView(
-                    appendAccessToken(authorizationRequest, accessToken, authentication, true),
+                    appendAccessToken(authorizationRequest, accessToken, authentication),
                     false,
                     true,
                     false
@@ -391,8 +391,7 @@ public class UaaAuthorizationEndpoint extends AbstractEndpoint {
 
     private String appendAccessToken(AuthorizationRequest authorizationRequest,
                                      OAuth2AccessToken accessToken,
-                                     Authentication authUser,
-                                     boolean fragment) {
+                                     Authentication authUser) {
 
         String requestedRedirect = authorizationRequest.getRedirectUri();
         if (accessToken == null) {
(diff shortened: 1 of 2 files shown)


fhanik:
Thank you @maust for your contribution.

We use TDD, test driven development, for all our code.
What we expect is two commits.

The first commit adds a test that tests the new or changed functionality. This test should fail(red) since the functionality has not been added. Basically, the first commit shows, `this is how I want things to work` and since the test fails, it clearly doesn't work that way today.

The second commit adds/changes the functionality of the code. and the test suite is green(passes)

By quickly glancing at the contribution, without tests, it is hard to judge it to be correct or complete.

Are you able to add tests to your PR?

maust:
@fhanik Sure, I updated the existing test. If I need to do something else let me know.

Build seems to be good now, only the default test failed because the build stalled. Not sure how I can retstart it.

fhanik:
Did you mean to close this issue?

maust:
Shouldn't I? Submitting the CLA was confirmed -> "Once you've publicized your membership, one of the owners of this repository can close and reopen this pull request, and dreddbot will take another look."

maust:
Two builds failed because the connection to LDAP failed. As I understand it is not caused by my changes.

fhanik:
no worries, we're still debugging that flakiness. I restarted the two jobs

sreetummidi:
To be reviewed

fhanik:
Requesting feedback on #600 before merging. Temporary branch at https://github.com/cloudfoundry/uaa/commit/c24a51bde9287996a8ad5b1eb5050469b5899964

fhanik:
OAuth 2 Spec shows an example as 

    HTTP/1.1 302 Found
    Location: http://example.com/cb#access_token=2YotnFZFEjr1zCsicMWpAA
               &state=xyz&token_type=example&expires_in=3600

No `?` is needed in the fragment. Parameters in the fragment are delimited by `&`
We believe the UAA to be compliant with the spec.

Please reopen if you feel we have made this decision in error.

Reference
https://tools.ietf.org/html/rfc6749#section-4.2.2

PR #605, opened 2017-05-03
Invalid Saved Account Cookie (including bad json) causes uaa to throw 500

6palace:


cbguder:
@6palace how was this invalid cookie produced? did UAA generate an invalid cookie, or was it manually crafted?

fhanik:
@cbguder Older versions of UAA produced cookies in a different format, a format that was invalid. So when that story was fixed, those cookies still existed in many user agents. 

Merging this request, as it is good behavior to only read cookies of a format we can understand without causing a 500

PR #612, opened 2017-05-12
Fix/error page branding

6palace:


fhanik:
@RequestMapping({"/error", "error404"})

shouldn't `error404` be `/error404`?

PR #626, opened 2017-06-01
Fix documentation of additional authorization attributes

torsten-sap:
The JSON object should be named "az_attr" instead of "additionalAuthorizationAttributes". See the following code:

https://github.com/cloudfoundry/uaa/blob/fe946fdc871027fcea9278e97c09cbf270831cf4/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java#L840


torsten-sap:
I publicized my membership.

PR #638, opened 2017-06-15
fix groups updates to only modify changed members

Bitty9:
Previously, each group update was making a db call for each member of that group, even if nothing had changed. Since roles is the only field which can be changed, we check if it has changed prior to making the db call. This decreases the number of db calls for group update, thus improves performance.


fhanik:
hi @Bitty9 Please provide a commit with test cases.

It would be great to have a test case that demonstrates that this works as expected. Easiest is to use mocks (like a spy) to ensure that methods were or were not called as expected.

sreetummidi:
@Bitty9 Are you able to provide the tests and resubmit the request

bsekar:
@sreetummidi Updated existing tests with the recommendation

fhanik:
You can still add a unit test that fails before the commit is applied and then passes after the commit has been applied. Make the dao object a `Mockito.spy` and see how many times `updateMember` was called

sreetummidi:
To be reviewed and merged

fhanik:
Closing via https://github.com/cloudfoundry/uaa/commit/a1df68d731d3ca65388812f885fdf241afe3ff97

PR #644, opened 2017-06-29
Reload login when CSRF token expires

bsekar:
When login page is left open until the point of expiry of CSRF token and the user goes on to submit their credentials, they are shown with the CSRF error `Invalid login attempt, request does not meet our security standards, please try again`. This can be avoided by performing a client-side page reload using `meta-refresh` when the CSRF token expires.

### DIFF
--- a/server/src/main/resources/templates/web/login.html
+++ b/server/src/main/resources/templates/web/login.html
@@ -1,5 +1,8 @@
 <!DOCTYPE html>
 <html xmlns:th="http://www.thymeleaf.org" xmlns:layout="http://www.ultraq.net.nz/thymeleaf/layout" layout:decorator="layouts/main">
+<head>
+    <meta http-equiv="refresh" th:content="${@loginCookieCsrfRepository.getCookieMaxAge()}" />
+</head>
 <div class="island" layout:fragment="page-content">
     <h1 th:text="${T(org.cloudfoundry.identity.uaa.zone.IdentityZoneHolder).uaa ? 'Welcome!':'Welcome to '+zone_name+'!'}">Welcome!</h1>
     <div class="island-content">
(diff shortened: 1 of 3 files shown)


sreetummidi:
To be reviewed and merged

cbguder:
Hey @bsekar!

A couple questions about this:

1. It looks like the default cookie age is 30 days, which seems like a really long time to stay on a login page. Do you change this value for your deployments?

2. It looks like the refresh interval is set to the default cookie age, and not the actual remaining time on the cookie. Isn't it possible for the user's cookie to have an expiration date that's less than the default cookie age (e.g. if they reload the page manually after the cookie was first created)?

Thanks,

CBG & Hieu

bsekar:
Hi Can (@cbguder) Thank you for reviewing this.

1. I agree that it can be made to expire sooner. Unfortunately, this value is [hard coded](https://github.com/cloudfoundry/uaa/blob/master/server/src/main/java/org/cloudfoundry/identity/uaa/security/web/CookieBasedCsrfTokenRepository.java#L32) and not configurable. So, we had to make a change in our [fork](https://github.com/GESoftware-CF/uaa/blob/release_4.4.0/server/src/main/java/org/cloudfoundry/identity/uaa/security/web/CookieBasedCsrfTokenRepository.java#L32). We set it to expire after 24 hours

2. During every login page load (including manual reloads), UAA will send a `Set-Cookie` header 
ex: ```Set-Cookie:X-Uaa-Csrf=tIEmEME29x0tfLeBVf1M6n; Max-Age=2592000; Expires=Sun, 27-Aug-2017 05:43:39 GMT; Secure; HttpOnly```. 
So, the browser will end up overwriting the cookie value/expiry etc. Therefore the `Max-Age` of the cookie will be approximately equal to the refresh-interval

cbguder:
Thanks for the response @bsekar!

We will merge this PR and update the CSRF cookie expiration to 24 hours.

PR #655, opened 2017-07-10
BUG: unsychronized usage of static SimpleDateFormat in JsonDateDeserializer

mwdb:
When using SimpleDateFormat as a static variable, access must be externally sychronized.
See https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html
"Synchronization

Date formats are not synchronized. It is recommended to create separate format instances for each thread. If multiple threads access a format concurrently, it must be synchronized externally."

This PR fixes parsing issues when using the model subproject as a client.


fhanik:
Thanks for the PR. I converted tabs to spaces and simplified the code by removing the thread local.

PR #662, opened 2017-07-19
OIDC token endpoint needs to support client authorization in either POST body or Basic auth

bsekar:
As per [spec](https://tools.ietf.org/html/rfc6749#section-2.3.1), client credentials can be specified using either methods.


sreetummidi:
Review and merge

PR #663, opened 2017-07-21
Fix: Remove the SAML Artifact Binding Support for all saml profiles

RoopGuron:


sreetummidi:
Review and merge

server/src/main/java/org/cloudfoundry/identity/uaa/home/HomeController.java
```diff
@@ -140,6 +140,12 @@ public String error404(Model model) {
         return "error";
     }
 
+    @RequestMapping("/error400")
```
fhanik: Is this really part of the SAML PR? This looks like a global change.

server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/LoginSamlEntryPoint.java
```diff
@@ -36,6 +46,30 @@ public void setProviderDefinitionList(SamlIdentityProviderConfigurator providerD
...
+
+        if (!processFilter(fi.getRequest())) {
+            chain.doFilter(request, response);
+            return;
+        }
+
+        try {
+            commence(fi.getRequest(), fi.getResponse(), null);
+        } catch (ServletException e) {
+            if(e.getRootCause() instanceof MetadataProviderException) {
```
fhanik: Is this the only time a MetadataProviderException is thrown? when the binding is not supported? 

Is it possible that a MetadataProviderException can be thrown for other reasons, at which point maybe we should log the error, at least in debug mode?

server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/SPWebSSOProfileImpl.java
```diff
@@ -0,0 +1,30 @@
+package org.cloudfoundry.identity.uaa.provider.saml;
```
fhanik: missing copyright section above package

server/src/test/java/org/cloudfoundry/identity/uaa/provider/saml/idp/ZoneAwareIdpMetadataGeneratorTest.java
```diff
@@ -20,20 +21,22 @@
...
+import java.util.Arrays;
 import java.util.List;
 
 import static org.cloudfoundry.identity.uaa.provider.saml.ZoneAwareMetadataGeneratorTests.cert1Plain;
 import static org.cloudfoundry.identity.uaa.provider.saml.ZoneAwareMetadataGeneratorTests.cert2Plain;
 import static org.cloudfoundry.identity.uaa.provider.saml.ZoneAwareMetadataGeneratorTests.samlKey1;
 import static org.cloudfoundry.identity.uaa.provider.saml.ZoneAwareMetadataGeneratorTests.samlKey2;
 import static org.cloudfoundry.identity.uaa.provider.saml.idp.SamlTestUtils.getCertificates;
-import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.*;
```
fhanik: Set IDE to not generate `*` imports. That way we know what classes are actually used.

uaa/src/main/webapp/WEB-INF/web.xml
```diff
@@ -82,4 +82,9 @@
         <location>/error404</location>
     </error-page>
 
+    <error-page>
```
fhanik: this is a global change. does it really pertain to removal of artifact binding?

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/SamlLoginWithLocalIdpIT.java
```diff
@@ -616,41 +727,48 @@ public void testCrossZoneSamlIntegration() throws Exception {
...
+            webDriver.findElement(By.name("password")).sendKeys("secr3T");
+            webDriver.findElement(By.xpath("//input[@value='Sign in']")).click();
+            assertThat(webDriver.findElement(By.cssSelector("h1")).getText(), Matchers.containsString("Where to?"));
+            Cookie afterLogin = webDriver.manage().getCookieNamed("JSESSIONID");
+            assertNotNull(afterLogin);
+            assertNotNull(afterLogin.getValue());
+            assertNotEquals(beforeLogin.getValue(), afterLogin.getValue());
+        } catch (Exception e) {
+            Assert.
+            assertTrue("Http-Artifact binding is not supported",e instanceof NoSuchElementException);
```
fhanik: this assertion is wrong. there is no guarantee that NoSuchElementException means artifact binding not supported.

PR #668, opened 2017-07-25
Ability to bulk create Users

bsekar:
This change implements the ability to bulk create users, `POST /Users/tx`

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/resources/ResourceManager.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/resources/ResourceManager.java
@@ -22,6 +22,10 @@ public interface ResourceManager<T> {
 
     T create(T resource, String zoneId);
 
+    default T create(T resource, String zoneId, boolean isBatchCall) {
+        return null;
+    };
+
     T update(String id, T resource, String zoneId);
 
     T delete(String id, int version, String zoneId);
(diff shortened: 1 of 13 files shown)


bsekar:
This PR has caused a rather interesting [failure](https://travis-ci.org/cloudfoundry/uaa/builds/257908410) on Postgres. @Transactional with Postgres would mean that it initiates rollback as soon as a SQL failure happens. [This](https://github.com/cloudfoundry/uaa/blob/master/server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupProvisioning.java#L166) method always causes SQL Exception during user creation, though it is suppressed in try catch. Therefore Postgres @Transactional is not possible with this change in its current state. I'm still working on fixing this

bsekar:
@tnwang Can you please provide an update on this PR?

tnwang:
Our developers took a look and passed along this note:
```
Use-case valid. SCIM defines Bulk endpoint but proposed implementation is proprietary.
```

I'll ask them to provide more details.

tnwang:
I spoke with our dojo members from SAP, and they noted that the SCIM specification defines the optional endpoint `bulk`, which allows to bundle all kinds of SCIM operations in a single request. Bulk creation of users would be one use-case for this, but it also supports e.g. deletion and operations for groups. It is defined in the spec here: http://www.simplecloud.info/specs/draft-scim-api-01.html#bulk-resources

The proposal would be to adopt this change to conform to the SCIM specification. As the endpoint is optional, in our eyes it would be ok to start with only partly implementing this endpoint (i.e. implement only the create action for users), but conform to the spec.

SAP would also welcome that endpoint, especially the group operations. By starting to implement the specified endpoint, it would be easy to extend it with other operations and still be standards conformant. Otherwise we could in the worst case end up with two different proprietary endpoints or one proprietary (if the PR is accepted) and another one according to the SCIM standard (when it is implemented).

sreetummidi:
Closing as we don't plan on supporting this flow.

PR #673, opened 2017-07-28
Feature/dockerize uaa

Bitty9:
Add gradle tasks for creating a docker file, image, pushing the image to a docker registry, and starting the docker container.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -34,79 +34,8 @@ buildscript {
   }
 }
 
-apply plugin: 'com.bmuschko.docker-remote-api'
-apply plugin: 'java'
-
-import com.bmuschko.gradle.docker.tasks.image.Dockerfile
-import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage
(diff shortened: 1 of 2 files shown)


fhanik:
@Bitty9 

Two items for suggestion

1. Please fix the CLA. The gitbot will keep warning us that we can't accept your PRs. The instructions are in the gitbot comment

2. Can I suggest to add a new file called `docker.gradle` so that we don't keep bloating the existing gradle file. You can still add a simple task in build.gradle that invokes it so that one doesn't have to specify the build file on the command line

6palace:
@fhanik will do. Also, we had feedback on this feature where teams needed better configuration options, since the pushed image is only configured with the default uaa.yml. We will add better ways to insert yml, should we close the pull request then re-open when ready?

Bitty9:
@6palace, the uaa.yml is injectable via the container environment. So if people are just pulling the image from the registry they can modify the uaa.yml, profile, and port bindings for their container using docker -e when they start the container.

PR #676, opened 2017-08-07
Add form_redirect_uri to login page during reload

bsekar:
When the user session expires, this `form_redirect_uri` is used to retain the context. But, the value is lost when login page is reloaded due to any error (ie incorrect password, csrf). This fix adds the value to the model if present in the original request


fhanik:
@bsekar I'm wondering if this fix should be as far in as a controller. 

after all, we process the redirect parameter in here
https://github.com/cloudfoundry/uaa/blob/8a20d975c89139ce50f6fe5bb1757d366f845662/server/src/main/java/org/cloudfoundry/identity/uaa/web/UaaSavedRequestCache.java#L82-L85

I'm wondering if we could call `request.setAttribute` here and save the redirect. I'll review it some more

PR #687, opened 2017-08-31
route all errors not explicitly mapped to the sad cloud error page to hide tomcat server information

RoopGuron:


fhanik:
The test can be a MockMvc test. You're welcome to submit or we will add it to the PR when it is time to merge

PR #690, opened 2017-09-05
Update UAA-APIs.rst

reinrl:
The example for external group mapping list filters in the API document I was viewing (http://docs.cloudfoundry.org/api/uaa/#list-4) was a bit confusing, hopefully this proposed change makes it a bit more clear (and matches the convention used elsewhere in this document).


tnwang:
Thanks reinrl for providing this PR. We have since deprecated the filter parameter in favor of using externalGroup and origin, and also moved on to new API documents.

PR #695, opened 2017-09-12
Feature/uaa read only

6palace:


6palace:
NOTE: currently degraded mode is configured through spring active profiles. This does not allow real-time toggling of read-only. We should explore ways to allow real-time toggling before merging this to develop.

Issue #679, opened 2017-08-21
UAA client create/update should accept wildcard in redirect_uri port


abh1kg:
```
uaac --debug client add test_client -i
New client secret:  ****
Verify new client secret:  ****
scope (list):  cloud_controller.read
authorized grant types (list):  authorization_code
authorities (list):
access token validity (seconds):
refresh token validity (seconds):
redirect uri (list):  http://localhost:*/test/path
autoapprove (list):
signup redirect url (url):

error response:
{
  "error": "invalid_client",
  "error_description": "One of the redirect_uri is invalid: http://localhost:*/test/path"
}
```

Similar redirect URI patterns worked earlier but redirection with the saved clients also fail now. This hampers local development quite a lot because only a specified port can be used in the redirect URL definition

jhamon:
In a recent release of UAA we added some security restrictions around which part of the `redirect_uri` may be wildcarded. Domains, specifically, were one big area of concern because of the way subdomain routes are assigned to other CF tenants that may not be considered trusted.

We're discussing internally whether the wildcarding of ports should be allowed or not.

sreetummidi:
@fhanik @jhamon @tnwang I had a discussion with the GE team yesterday. I have asked them to make the following changes in the PR. 


1. Introduce the concept of whitelisting allowed redirect uri protocols at the zone level The default is HTTP and HTTPS
2. When create clients needing redirect uri, we will validate against the zone whitelist


Apart from this, this PR also contains allowing regex pattern for redirect port. I have asked them to keep these code changes intact within the PR

Thanks,
Sree

RoopGuron:
@sreetummidi As we at GE are implementing the protocol whitelisting, we have noticed a break in backward compatibility caused by whitelisting. Even though its impact for GE is minimal we do want to bring this to your attention.

The protocol pattern has been for quite some time(At least since Feb this year) on the lines of
`^(http(\\*|s)?://)"    //URL starts with 'http://' or 'https://' or 'http*://`
But changing this to a protocol whitelisting (for default uaa zone) would be: 
`^(http|https)://`
Thus any clients registered with redirectUris using wildcard in the protocol part would not work for  authorization grant.
There is an explicit test `org.cloudfoundry.identity.uaa.mock.token.TokenMvcMockTests#testWildcardRedirectURL`
which sets up a redirectURI: 
`String redirectUri = "http*://subdomain.domain.com/**/path2?query1=value1";` which would fail and would need to be changed or removed.

[...]

PR #699, opened 2017-09-18
Feature/support custom redirect uri schemes

RoopGuron:


jhamon:
Could you give some background on the use-case and motivation for this change?

Bitty9:
We have teams which would like to redirect to a native mobile app using custom schemes which are registered with the phone OS after authentication. With the redirect uri enforcement changes, the ability to redirect to custom schemes was lost.  See https://tools.ietf.org/html/rfc6749#section-9 for more details.

aashah:
Hm, It'd be nice to look into better ways of validating incoming URIs than expanding the list of Regex patterns. Some languages provide standardized URI regexps (downside being we'd have to upgrade the lib/language if there's an issue) that would probably have allowed your use case. 

Something to look into as well...

Bitty9:
@aashah The reason we are using a custom regex pattern is because UAA is more restrictive than traditional uris about what is allowed for a client. For example, a redirect uri can have a * in the subdomain but not in the domain. Similarly, the scheme can have http* to match both http and https. This comes into play when a client is created because a client may provide a redirect_uri which is an ant-style path pattern. Do you have a specific library in mind which would accommodate these types of restrictions without turning into writing our own custom regex?

sreetummidi:
@RoopGuron Where are the custom URLs whitelist of allowed patterns set?

RoopGuron:
@sreetummidi 
Change only modifies the pattern applied to the **_scheme_** part of the **_URI pattern_** being applied in the `UaaUrlUtils.java.allowedRedirectUriPattern`
The **_scheme_** pattern `^(([a-zA-Z][a-zA-Z0-9+-.]*)|(http(\\*|s)?)):`conforms to the RFC Uniform Resource Identifier (URI) section on scheme: https://tools.ietf.org/html/rfc3986#section-3.1

As per the examples of Custom URLs that would satisfy this would be as follows:
```
           "ftp://ftp.is.co.za/rfc/rfc1808.txt",
            "cool-app://example.com",
            "org.cloudfoundry.identity://mobile-windows-app.com/view",
            "org+cloudfoundry+identity://mobile-ios-app.com/view",
            "org-cl0udfoundry-identity://mobile-android-app.com/view"
         "http://localhost",
        "http://localhost:8080",
        "http://localhost:8080/uaa",
        "http://valid.com",
        "http://sub.valid.com",
        "http://valid.com/with/path",
        "https://subsub.sub.valid.com/**",
        "https://valid.com/path/*/path",
        "http://sub.valid.com/*/with/path**",
        "http*://sub.valid.com/*/with/path**",
        "http*://*.valid.com/*/with/path**",
        "http://*.valid.com/*/with/path**",
        "https://*.valid.com/*/with/path**",
        "https://*.*.valid.com/*/with/path**",
        "http://sub*.valid.com/*/with/path**",
        "http://*.domain.com",
        "http://username:password@some.server.com",
        "http://username:password@some.server.com/path"
```

Bitty9:
Last commit includes a wildcard in redirect uri for #679

sreetummidi:
@RoopGuron I was initially under the impression that we are whitelisting known set of mobile/iot protocols (I am not sure if any such list exists). This PR is adding support for any kind of text passing as a protocol which is fairly open. Is the protocol part the name of the mobile application?

RoopGuron:
@sreetumidi the protocol part(custom scheme) is not necessarily the name of the native/mobile app, but is associated with one particular mobile app. Please refer to comment by @bitty9 where there is a link to the rfc section addressing this.    As potentially there could be multiple mobile applications using oauth2 authorization grant flow to get refresh tokens, there could be multiple custom schemes, each scheme name has to conform to the uri scheme BMW pattern.

ilipkind:
The latest discussion on this pull request can be found here: https://github.com/cloudfoundry/uaa/issues/679

sreetummidi:
Closing in lieu of https://www.pivotaltracker.com/story/show/154735792

PR #708, opened 2017-09-29
Adding some corrections and content improvements to documentation

mysterlune:


sreetummidi:
Feature for external login server will be deprecated and removed in favor of standard integration patterns like SAML, LDAP & OpenID Connect.

PR #709, opened 2017-10-03
Allow underscores in subdomain section

sks:
A client using ` authorization_code ` and with a redirect url that has underscore in it `eg: my_fancy_app.domain.com/oauth/token ` would get an error stating `invalid_client ` .

This pull request should fix it.


sks:
Closing this and would open one against develop branch

PR #713, opened 2017-10-04
update ldap api

strehle:
our code scanner maked this library as problem, because of
	CVE-2015-3250

There is no direct usage of problematic code,
but update prevents any issues here


strehle:
hi @fhanik  thanks for the other updates in 4.7.1, but is there a reason why you missed this one ?

fhanik:
@strehle I believe @sreetummidi and @tnwang may just have missed it. 

I'll merge this for develop. 4.7.1 is not affected, we're not using the ApacheDS for very much.

PR #714, opened 2017-10-04
use isBlank form apache instead of isEmpty for hsqldb

strehle:
The use of this method from hsqldb causes problem if jar is not available
The dependent library is not needed for usage if mysql or postgesql therefore we removed it.
However with this dependency it is a hard requirement which should not be the case.

If you check the internals of isEmpty of hsqldb then it is the same if you call isBlank from apache
see also
http://technohandle.blogspot.de/2011/01/stringutilsisemtpty-vs.html


strehle:
Hi @fhanik again,  this PR is because we want remove dependency to hsql , I assume the use of this call was done by accident , 
thanks

fhanik:
actually, it should be `org.springframework.util.StringUtils.isEmpty` from the Spring Framework. 

can you modify? some the apache libraries are transitive dependencies and can change

strehle:
@fhanik done , because fine for us

strehle:
thanks

PR #727, opened 2017-10-24
Do not update identical client secret

klaus-sap:
When updating a client secret with an identical value, the client secret hash changes. As a consequence, all issued client tokens become invalid ("revocable signature mismatch"). If the client secret doesn't change, then a client token should remain valid as long as the token has not been revoked.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/client/ClientAdminEndpoints.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/client/ClientAdminEndpoints.java
@@ -425,8 +425,8 @@ public class ClientAdminEndpoints implements InitializingBean, ApplicationEventP
                 clientId = change[i].getClientId();
                 clientDetails[i] = new ClientDetailsModification(clientDetailsService.retrieve(clientId, IdentityZoneHolder.get().getId()));
                 boolean oldPasswordOk = authenticateClient(clientId, change[i].getOldSecret());
-                if (!change[i].getSecret().equals(change[i].getOldSecret())) {
-                    clientDetailsValidator.getClientSecretValidator().validate(change[i].getSecret());
+                clientDetailsValidator.getClientSecretValidator().validate(change[i].getSecret());
+                if (!authenticateClient(clientId, change[i].getSecret())) {
                     clientRegistrationService.updateClientSecret(clientId, change[i].getSecret(), IdentityZoneHolder.get().getId());
                     clientSecretChanges.incrementAndGet();
                 }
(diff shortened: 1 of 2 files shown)


klaus-sap:
Remark: in class ClientAdminBootstrap, a client secret is updated only if the secret has changed (method didPasswordChange).

tnwang:
@strehle @klaus-sap Is there a reason for which you are trying to update the secret to be an identical secret? Wouldn't it be enough to not change the secret at all?

One concern is that such a check exposes information about the existing client secret, though this would be limited to client admins.

strehle:
@tnwang a consourse based (re-)deployment causes such situations . the client update is triggered from bootstrap during start of new uaa. depending apps which requested a token during green/blue deployment end in an invalid token error.

fhanik:
@strehle - setting `override: false` for such clients would prevent that. 

I think that the operation

    set client secret

indicates a desired change, regardless of the new value.
it is our policy that when such change takes place, invalidate previous tokens

I'm not sure moving away from this policy is the answer. we should get together on slack and talk some more

klaus-sap:
@tnwang @fhanik In class ClientAdminBootstrap, a client secret is updated only if the secret has changed (method didPasswordChange). This PR is about doing the same in the rest endpoint "/oauth/clients/{client_id}/secret".

It's not about desired changes, it's about automated deployments. In my opinion, re-deployments should not change microservices as long as the configuration is unchanged.

If you wish, I will contact you on slack to talk some more.

fhanik:
@klaus-sap I see, so some software calls `/oauth/clients` to change a secret. 
I'm still inclined to think that should follow the `update` password procedure.

It would seem as the applications shouldn't be calling APIs if there was no intent on changing values, and that this PR is simply asking for the UAA to work around the applications behavior to make the API idempotent. For example, should a `password last modified` timestamp value change if the value was changed to the same plaintext value?

Let me digest this a little bit more before we make a decision to make sure that if we merged this PR everything is still to expectation.

tnwang:
Some more context on what's the use case to call the automated deployment method with the same secret would help. 

Why would the automated deployment need to call an API over a manifest bootstrap? Is this because there are non-manifest operators using the APIs or is this to avoid having to do a restart of UAA to redeploy certain clients?

Do they need to call secret change on every redeployment? Would a check for the current client secret succeeding be usable to avoid performing update the secret?

One thing we would want to avoid is disclosing any information about the existing client secret.

Know that this is a lot of questions - this would greatly help us understand what you're facing.

klaus-sap:
@tnwang Our automated script does not deploy the UAA, but another application which needs a dedicated OAuth client. As the script doesn't have access to uaa.yml, it calls the rest API to update the OAuth client according to the current configuration of the application.

In order to check the current client secret, we would either need to keep a history of the configuration or request a token.

I think that your question about information disclosure might decide about this PR. We might adapt the coding to not leak any information regardless if the secret has updated or not. On the other hand: is this really an issue for an admin who is authorized to overwrite any client secret?

tnwang:
Even if the message were to send back a successful message that the secret changed even if it was not, you could still tell via the token that the revocable sig has not changed leading to information disclosure about the secret.

PR #728, opened 2017-10-31
Docs how to restrict LDAP auth based on memberOf

keymon:
Covers the case of tweaking the search attribute to restrict the
authentication to users members of a given group in LDAP.


shamus:
Closing due to inactivity.

PR #729, opened 2017-11-02
Fix type for claim az_attr

strehle:
Documentation says that with authorities additional claims
can be added into JWT.
However with /check_token this fails because of invalid json
object if you add here a string map.
See:
https://github.com/cloudfoundry/uaa/blob/master/docs/UAA-APIs.rst

Example:
authorities={"az_attr":{"external_group":"domain\\group1","external_id":"abcd1234"}}


strehle:
hi @sreetummidi / @fhanik   , 
our application users found this issue.
az_attr must be type Map<String, String>, see
https://github.com/cloudfoundry/uaa/blob/fe946fdc871027fcea9278e97c09cbf270831cf4/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java#L840

PR #732, opened 2017-11-14
Change docker plugin version. Run docker image as non root user.

Bitty9:


tnwang:
Hi @Bitty9, can you provide a description of what this pull request does and what drove the creation of it?

sreetummidi:
@Bitty9 Can you provide the use case for building docker images via gradle. This can already be done with the available docker tools.

sreetummidi:
@DennisDenuto @jhamon Can you please provide some insight on how to accomplish this via docker.

DennisDenuto:
+1 on understanding why we need to add a non-root user `dockeruser`. Is this coming from a security concern?

@sreetummidi if the goal is to run uaa for a developer, this can be done without building a 'uaa docker image' and instead to use the official gradle docker image:
```
cd ~/workspace/uaa
docker run --rm -p 8080:8080 -v "$PWD":/home/gradle/project -w /home/gradle/project gradle gradle cargoRunLocal
```

Bitty9:
@sreetummidi, when we originally put in the docker pull request we used the gradle plugin because it is as simple as calling a gradle task to get the container running and kept track of the commands/setup we were using. You could just as easily do this with pure docker commands. 

@DennisDenuto, We changed the docker container by default run as a non root user as a security requirement from a customer and is more of a safety precaution.

jhamon:
@Bitty9 I think one of the things we're trying to understand is who is using the resulting image and for what purpose. I thought this was only for development purposes, but if customers are using it as you said I'd like to know more about that. How are they using it? For what purpose?

I'm trying to clarify the line between what is a product and what is a developer tool.

Bitty9:
@jhamon, the customer is using the docker image to deploy their own UAA (with their own configurations) in an isolated environment that is not a CF environment.  This is not used for development purposes.

jhamon:
We're going to decline the PR and probably remove this docker-related gradle configuration for now. If you would like to build a dockerized UAA, you can do that by migrating most of these instructions into a dedicated `Dockerfile` and build it using the `docker build` cli command. But as it is, we're not ready to support Docker use cases. Others are now asking us to publish #754 official builds but we currently don't have the infrastructure in place to build, test, publish, patch, and support these as new forms of our product. 

Of particular concern to me is that it looks like this image is being built with the `required_configurations.yml` file included. This file contains private keys that have been committed and published with the rest of the UAA's open source code. These are intended only for development use, but now there is a danger of running in production with these compromised keys if those values are not injected as envisioned in the original PR #673. It's not a safe way to distribute UAA, since anyone who doesn't realize they need to configure those properties will get the compromised keys.

Bitty9:
I understand that you guys aren't able to support an official docker image due to limitations on your end. However, I am not sure removing the ability for others to build their own image locally makes sense either. It seems the open source community may find it useful.

The current image runs out of the box with the dev keys, similar to running UAA locally (which could be removed --- but then we lose the ability to start a container out of the box and must provide instructions to get it running). However for prod cases you can easily override the LOGIN_CONFIG_URL environment variable on your container. Perhaps the right way to approach this is to supplement with some documentation for configuring/starting a prod docker UAA instance. (We give similar instructions to our customers).

jhamon:
Removing this configuration does not impede the ability of anyone in the community to build a Docker image of the UAA. You may still create a `Dockerfile` and build it containing the UAA and anything else you might need. Keeping this in the repo implies a level of support that we are not ready to provide.

Having an insecure behavior that "works" out-of-the-box is doing a disservice to those who may not be experts in how the UAA and larger identity domain work. The friction of having to provide configuration to get into a working state guides users toward the correct behavior, which is specifying their own private keys and certificates.

We have an epic coming up to revamp our release process automation. We may look at Docker deliverables as a follow-on if there is enough interest, but our initial focus will be on better automated testing of bosh releases since they are the building blocks that make up Cloud Foundry.

ge-guardians:
This is a bit confusing that docker support suddenly dropped by Pivotal and prior docker checkins was welcomed and advertised as part of UAA release. @jhamon @sreetummidi don't we need to have public input on this matter as Open Source project and then make a decision? There are ways around to make sure keys not copied from image to image.

ilipkind:
This is a bit confusing that docker support suddenly dropped by Pivotal and prior docker checkins was welcomed and advertised as part of UAA release. @jhamon @sreetummidi don't we need to have public input on this matter as Open Source project and then make a decision? There are ways around to make sure keys not copied from image to image.

PR #738, opened 2017-11-29
user delete in other zone does not remove members correctly

arjunnayak:
Currently deleting a user in an identity zone will not correctly delete the group membership associated with that user. This fixes that and adds a test for the use case.


arjunnayak:
I should be under the Predix organization (was previously private), so please reopen the pull request for CLA.

tack-sap:
Merged.

Thanks for the PR!

martinsubash:
Hello Team,
The similar issue still in the latest 4.27.0 released version.

My Scenario:
1) Create UserA with **origin = auth0** and assigned to AdminGroup
2) Delete the UserA, but still AdminGroup had the member UserA. membership not removed properly.

This issue occurs only for **origin=auth0** for **origin=uaa** its working fine.

Is the above fix includes this also or its a new issue?

PR #739, opened 2017-12-02
Fix/saml login with incorrect attributes

RoopGuron:
Use Case: SAML Federation with External IDP assuming Shadow users pre provisioned
Team not sure of Attributes on the External IDP and thus configures the external IDP in UAA with incorrect attributeMappings.

Current Behavior:  UAA while consuming the SAML Assertion for IDP with misconfigured attributeMappings, defaults the uaaUser with email,givenName and familyName like 
    (Either: userName@unknown.org, userName and “unknown” for givenName/familyName  OR if the userName is an email :: email: userName givenName: userNameportionbefore@ familyName:emailDomainPart)

Desired Behavior: UAA while consuming SAML Assertion for IDP with misconfigured attributeMappings, **would not overwrite the existing attributes(email,givenName and familyName) with the defaults(mentioned above in Current Behaviour) and rather set those to null for the shadow user.**


shamus:
Closing due to inactivity.

PR #747, opened 2017-12-15
Add input validation for az_attr

strehle:
In addition to pull request #729 add a type check for az_attr
and prevent write if type is not valid.
Remark: with PR #729 the validation already fails if az_attr claim
has not correct type, however now we prevent that a JWT is created with
an invalid az_attr input.

The result JWT does not contain the az_attr claim if type check fails


strehle:
Hi, @tack-sap , @torsten-sap 
in addtion to PR #729 here an input validation check for az_attr

regards,
-markus

PR #759, opened 2018-01-17
Remove token values from Exceptions

tack-sap:
Remove any token values from the Exceptions thrown in UaaTokenServices to prevent leakage of these values.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java
@@ -226,13 +226,12 @@ public class UaaTokenServices implements AuthorizationServerTokenServices, Resou
         // OAuth2RefreshToken refreshToken =
         // tokenStore.readRefreshToken(refreshTokenValue);
         // if (refreshToken == null) {
-        // throw new InvalidGrantException("Invalid refresh token: " +
-        // refreshTokenValue);
+        // throw new InvalidGrantException("Invalid refresh token");
         // }
 
         String clientId = (String) claims.get(CID);
         if (clientId == null || !clientId.equals(request.getClientId())) {
-            throw new InvalidGrantException("Wrong client for this refresh token: " + refreshTokenValue);
+            throw new InvalidGrantException("Wrong client for this refresh token: " + clientId);
         }
 
         String userid = (String) claims.get(USER_ID);
@@ -256,8 +255,7 @@ public class UaaTokenServices implements AuthorizationServerTokenServices, Resou
         long refreshTokenExpireDate = refreshTokenExpiry.longValue() * 1000l;
 
         if (new Date(refreshTokenExpireDate).before(new Date())) {
-            throw new InvalidTokenException("Invalid refresh token (expired): " + refreshTokenValue + " expired at "
-                            + new Date(refreshTokenExpireDate));
+            throw new InvalidTokenException("Invalid refresh token expired at " + new Date(refreshTokenExpireDate));
         }
 
         // default request scopes to what is in the refresh token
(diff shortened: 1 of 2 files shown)


fhanik:
If you removed the signature from the logged token value, would it still be a leak?
I can see where having the token(claims) would be a useful tool in debugging.

tack-sap:
Hi Filip,
should be okay, as long as the value in the log cannot be used for anything. However I'm not sure, whether the value is a JWT value all the time.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java
```diff
@@ -226,13 +226,12 @@ public OAuth2AccessToken refreshAccessToken(String refreshTokenValue, TokenReque
...
         // if (refreshToken == null) {
-        // throw new InvalidGrantException("Invalid refresh token: " +
-        // refreshTokenValue);
+        // throw new InvalidGrantException("Invalid refresh token");
         // }
 
         String clientId = (String) claims.get(CID);
         if (clientId == null || !clientId.equals(request.getClientId())) {
-            throw new InvalidGrantException("Wrong client for this refresh token: " + refreshTokenValue);
+            throw new InvalidGrantException("Wrong client for this refresh token: " + clientId);
```
strehle: clientId may be null, see line before

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java
```diff
@@ -301,7 +299,7 @@ public OAuth2AccessToken refreshAccessToken(String refreshTokenValue, TokenReque
             }
             String newRevocableHashSignature = UaaTokenUtils.getRevocableTokenSignature(client, clientSecretForHash, user);
             if (!revocableHashSignature.equals(newRevocableHashSignature)) {
-                throw new TokenRevokedException(refreshTokenValue);
+                throw new TokenRevokedException("Invalid refresh token: revocable signature mismatch");
```
strehle: here maybe the id could be logged to that it could be found in revocable DB tablöe

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java
```diff
@@ -1026,7 +1024,7 @@ public void setUserDatabase(UaaUserDatabase userDatabase) {
     @Override
     public OAuth2Authentication loadAuthentication(String accessToken) throws AuthenticationException {
         if (StringUtils.isEmpty(accessToken)) {
-            throw new InvalidTokenException("Invalid access token value, must be at least 30 characters:"+accessToken);
+            throw new InvalidTokenException("Invalid access token value, must be at least 30 characters");
```
strehle: here it is better to omit it, because it is empty

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java
```diff
@@ -1036,8 +1034,7 @@ public OAuth2Authentication loadAuthentication(String accessToken) throws Authen
         // Check token expiry
         Integer expiration = (Integer) claims.get(EXP);
         if (expiration != null && new Date(expiration * 1000l).before(new Date())) {
-            throw new InvalidTokenException("Invalid access token (expired): " + accessToken + " expired at "
-                            + new Date(expiration * 1000l));
+            throw new InvalidTokenException("Invalid access token: expired at " + new Date(expiration * 1000l));
```
strehle: token not needed

fhanik:
@tack-sap - that's a good point.

tnwang:
Closing as it was implemented by https://www.pivotaltracker.com/story/show/154401356

PR #763, opened 2018-01-31
Fix the problem with WAS (IBM WebSphere) 9

jbortoletto:
In IBM WebSphere 9, this getServletPath() returns a blank String, and with a blank origin, UAA is unable to successfully authenticate.
This getRequestURL() works fine for both Tomcat and WAS 9.


tack-sap:
Merged via 22c9e7b

PR #768, opened 2018-02-08
Refactor token enhancer call

strehle:
- only once
- add info only if interface used


jhamon:
Hi @strehle,

These refactorings seem reasonable, but in the course of testing out this feature we noticed a few problems. To test it out, we created our own enhancer implementation like this

## Setup

```
public class UaaTokenEnhancerExample implements UaaTokenEnhancer {

    @Override
    public Map<String, String> getExternalAttributes(OAuth2Authentication authentication) {
        Map<String, String> externalAttributes = new HashMap<>();
        externalAttributes.put("purpose", "shmurpis");
        return externalAttributes;
    }

    @Override
    public Map<String, Object> enhance(Map<String, Object> claims, OAuth2Authentication authentication) {
        return new HashMap<String, Object>() {{
            put("stuff", "see what happens");
        }};
    }
}
```

and wired it into the UAA with this xml config

```
<bean id="tokenServices" class="org.cloudfoundry.identity.uaa.oauth.UaaTokenServices">
        <property name="clientDetailsService" ref="jdbcClientDetailsService"/>
        <property name="idTokenCreator" ref="idTokenCreator"/>
        <property name="userDatabase" ref="userDatabase"/>
        <property name="issuer" value="${issuer.uri}"/>
        <property name="approvalStore" ref="approvalStore"/>
        <property name="tokenPolicy" ref="globalTokenPolicy"/>
        <property name="excludedClaims" ref="excludedClaims"/>
        <property name="tokenProvisioning" ref="revocableTokenProvisioning"/>
        <property name="restrictRefreshGrant" value="${jwt.token.refresh.restrict_grant:false}"/>
        <property name="accessTokenValidityResolver" ref="accessTokenValidityResolver" />
        <property name="refreshTokenValidityResolver" ref="refreshTokenValidityResolver" />
        <property name="uaaTokenEnhancer" ref="uaaTokenEnhancerExample"/>
    </bean>

    <bean id="uaaTokenEnhancerExample" class="org.cloudfoundry.identity.uaa.oauth.UaaTokenEnhancerExample"/>
```

## Getting a token

Next we tried to get an id token for a test user by visiting `http://localhost:8080/uaa/oauth/authorize?response_type=id_token&client_id=oauth_showcase_implicit_grant&scope=openid&redirect_uri=http://localhost:8080/uaa/`.

We see the following response:

```
HTTP/1.1 302
Access-Control-Allow-Origin: *
Strict-Transport-Security: max-age=31536000
Cache-Control: no-cache, no-store, max-age=0, must-revalidate
Pragma: no-cache
Expires: 0
X-XSS-Protection: 1; mode=block
X-Frame-Options: DENY
X-Content-Type-Options: nosniff
Location: http://localhost:8080/uaa/#token_type=bearer&id_token=eyJhbGciOiJIUzI1NiIsImtpZCI6ImxlZ2FjeS10b2tlbi1rZXkiLCJ0eXAiOiJKV1QifQ.eyJzdWIiOiI4MmNkOTFlNC0xZDBhLTQzNDctOTI4NS1iNzYyMDFiNzkyMDciLCJhdWQiOlsib2F1dGhfc2hvd2Nhc2VfaW1wbGljaXRfZ3JhbnQiXSwiaXNzIjoiaHR0cDovL2xvY2FsaG9zdDo4MDgwL3VhYS9vYXV0aC90b2tlbiIsImV4cCI6MTUyMDQyNzYzNywiaWF0IjoxNTIwMzg0NDM3LCJhbXIiOlsicHdkIl0sImF6cCI6Im9hdXRoX3Nob3djYXNlX2ltcGxpY2l0X2dyYW50Iiwic2NvcGUiOlsib3BlbmlkIl0sImVtYWlsIjoibWFyaXNzYUB0ZXN0Lm9yZyIsInppZCI6InVhYSIsIm9yaWdpbiI6InVhYSIsImp0aSI6IjRkMDU5ZWU3OTcxZTQ1OTY5NmI0MTk2MGY1OGY5YmE2IiwicHJldmlvdXNfbG9nb25fdGltZSI6MTUyMDM4MjU0OTk0MCwiZW1haWxfdmVyaWZpZWQiOnRydWUsImNsaWVudF9pZCI6Im9hdXRoX3Nob3djYXNlX2ltcGxpY2l0X2dyYW50IiwiY2lkIjoib2F1dGhfc2hvd2Nhc2VfaW1wbGljaXRfZ3JhbnQiLCJ1c2VyX25hbWUiOiJtYXJpc3NhIiwicmV2X3NpZyI6ImE0ZTc3NTc2IiwidXNlcl9pZCI6IjgyY2Q5MWU0LTFkMGEtNDM0Ny05Mjg1LWI3NjIwMWI3OTIwNyIsImF1dGhfdGltZSI6MTUyMDM4NDMxN30.l_JMFUg4aOsK44QQZbxHjhR3Z8iV63l-Qz_dID5TUbU&expires_in=43199&jti=4d059ee7971e459696b41960f58f9ba6&stuff=see%20what%20happens
Content-Language: en-US
Content-Length: 0
Date: Wed, 07 Mar 2018 01:00:37 GMT
```

## Observations

1. We see that the token enhancements have been added into the `Location` header of the 302 response. We think this is happening because additional info is included as query params [here](https://github.com/cloudfoundry/uaa/blob/45d45b554996a50d7ce6fb470d2012658e919495/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaAuthorizationEndpoint.java#L499). Was that intended? We think it probably doesn't make sense to be leaking extra information into the URL in this way.

2. We also noticed from stepping through the code that the enhancements are being added to the refresh token in addition to the access token. We think this seems like unintended behavior since the refresh token's only purpose is request a new access token from the UAA which doesn't care about these extra claims.

Any thoughts on these issues?

fhanik:
Answer to question 2.

When refreshing an access token (refresh_token grant) the new access token should NOT be run through the token enhancer. Because this can yield a different token.

Therefor, a refresh token must contain all claims that the original access token had. so that a refreshed access token can be reconstructed. 

If the token enhancer changes the refreshed access token, this contract is broken.

strehle:
Hi,
I will check, however I dont see a leak. The implicit grant from returns the full id_token in Location (and would be return access token , too). I dont see a reason, why only the value from enhance is see in location.
Filip already answered to Q2 and that was the point for this pull , enhancer call should be done only once.

jhamon:
If you look closely (at the very end) at the `Location` header in my example above you will see where the additional information has been included in the header (`&stuff=see%20what%20happens`).

@strehle Our PMs think this shouldn't go out in a release in its current form, but we don't have the resources to take this on ourselves. You need to come up with a plan for a fix soon or we will have to revert the work-in-progress that has already been merged.

strehle:
I see the information in Location at end, my point is that the token and the information from enhancer belong to the user and both are returned in Location in case of implicit grant, therefore I dont see a leak here.

However to omit this I suggest , either we 
a) enhance the token only if grant_type != null, means not implicit grant
b) I refactor the enhancement so that the additional information is putted directly to Json Response but not to the object OAuth2AccessToken , so that it is part of the token but not part of the response URI (Location header).

So a) or b) ?

tnwang:
Hi @strehle, our thoughts are that B is better since the token is already available in the token, and looking at the user attributes from the token should be the standard use case.

strehle:
hi @jhamon , @tnwang 
refactored as described in https://github.com/cloudfoundry/uaa/pull/768#issuecomment-371614744 b)
failing test is in sqlserver, dont know why, however please proceed with your tests to accept this PR,
thanks

torsten-sap:
I retriggered the sqlserver test and it went through

PR #770, opened 2018-02-08
zone level SAML signatureAlgorithm

bsekar:
Provide functionality to set SAML `signatureAlgorithm` in Zone config
1. When UAA is SP, this algorithm will be used to sign the SAML Authentication request.
1. When UAA is IdP, this algorithm will be used to sign the SAML Assertion.
1. yaml properties follow the `global` notation used by `ClientSecretPolicy`, `PasswordPolicy` etc.

Note: Even if this value is updated to a different signatureAlgorithm, it does not need a metadata update. Existing SP-IdP integrations will continue to work


sreetummidi:
@bsekar Can you please update the PR not to introduce a new policy for a global setting for the signature algorithm. We can re-use the existing one under SAML Settings as a global setting

6palace:
@sreetummidi The reason for keeping two settings is for consistency with other configurations like password policy, lockout, or token validity. Just wanted to confirm that this kind of separation is no longer needed.

6palace:
added extra commit to use zone saml signature method when signing saml metadata as well.

sreetummidi:
We don't plan on supporting this.

PR #772, opened 2018-02-09
UaaTokenStore: Don't fall over when failing to expire oauth codes

mook-as:
MySQL does row-level locking, which means when we delete expired oauth codes in parallel we can sometimes hit a harmless deadlock (where the two delete updates execute across the rows in different order simultaneously).  This is not an issue as the winning update will expire tokens as appropriate, plus the tokens are checked for expiry on use.

This was encountered when repeatedly running multiple instances of `cf ssh` in parallel on a Cloud Foundry deployment using a single non-HA UAA instance (backed by a single non-HA MySQL database).

Thanks for building UAA!


fhanik:
Closed via: https://github.com/cloudfoundry/uaa/commit/ebdecc457e64d63a26841b3478ab497c8d6d97d0

Issue #774, opened 2018-02-13
Account Cookies


strehle:
**SECURITY NOTICE: If you have found a security problem in the UAA, please do not file a public github issue. Instead, please send an email to security@cloudfoundry.org**

Thanks for taking the time to file an issue. You'll minimize back and forth and help us help you more effectively by answering all of the following questions as specifically and completely as you can.

### What version of UAA are you running?

4.10.0

### How are you deploying the UAA?

gradlew

### What did you do?

Logon and checked received cookies.

### What did you expect to see? What goal are you trying to achieve with the UAA?

Cookies with secure flags and httpOnly = true

### What did you see instead?

httpOnly = false

---------------
The cookie is sent out in

https://github.com/cloudfoundry/uaa/blob/master/server/src/main/java/org/cloudfoundry/identity/uaa/login/AccountSavingAuthenticationSuccessHandler.java#L72-L77

and used only if IdpDiscovery is active

https://github.com/cloudfoundry/uaa/blob/master/server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java#L473-L489

Therefore the cookie should not be written in case IdpDiscovery is not active.

Problem with the cookie is that there are personal information like an email and this information is stored for a year.

jhamon:
As I commented on #775, the `Current-User` cookie is needed for our implementation of OIDC session management. It is intentionally `httpOnly=false` because it must be readable by [javascripts](https://github.com/cloudfoundry/uaa/blob/a4132cbb27f5b444de2a9b85f4f157694bebd6fc/uaa/src/main/webapp/resources/javascripts/session/session.js#L53) running in the UAA's OP iframe; javascripts cannot access cookies that are httpOnly. It may be used whether or not account discovery is in use if there are relying parties using uaa-singular or other library to discover login status. 

This [blog post](https://medium.com/@piraveenaparalogarajah/openid-connect-session-management-dc6a65040cc) gives a pretty good overview of the OIDC session management flow if you're not familiar with how it works. The goal is for relying parties to be notified of a change in user login status without each one having to poll the UAA continuously for updates. The polling is replaced with message passing to a UAA iframe that reads this cookie on the UAA domain to see if a change has occurred in the user's login status (e.g. if the user has logged out in a different browser tab). 

As for the content of the cookie, it is a serialization of the user's guid. It does not contain an email address or other personally identifiable information.

tnwang:
[...]

PR #775, opened 2018-02-13
missing cookie switch

strehle:
see issue #774 

This PR is a suggestion to omit the cookie.
The existance is required in mock tests but only in mocks not integration tests


jhamon:
We actually do need the Current User cookie. It is used by [uaa-singular](https://github.com/cloudfoundry/uaa-singular) in our implementation of OIDC session management. It seems if any change is needed here, it is to add assertions about the presence of Current User cookie to relevant integration tests.

tack-sap:
Merged via 3ae3e72

Thanks for the PR!

PR #781, opened 2018-03-02
Escape the correct regex characters

mikeroda:
There is a typo in the regex pattern consisting of characters that don't need to be escaped. The pattern "A-z" actually includes all the characters between the uppercase A and lowercase Z, namely "[", "\", "]", "^", "_", and "`". Without this fix, those characters are treated as regex characters instead of string literals.

### DIFF
--- a/model/src/main/java/org/cloudfoundry/identity/uaa/util/UaaStringUtils.java
+++ b/model/src/main/java/org/cloudfoundry/identity/uaa/util/UaaStringUtils.java
@@ -143,7 +143,7 @@ public class UaaStringUtils {
      * @return a regular expression string that will only match exact literals
      */
     public static String escapeRegExCharacters(String s) {
-        return escapeRegExCharacters(s, "([^a-zA-z0-9 ])");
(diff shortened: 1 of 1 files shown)


aeijdenberg:
Nice catch. I hope the UAA folk see this and merge it quickly. Depending on how this utility class is used, there could be some potential for mischief here...

jhamon:
This looks like a very reasonable change, but it would be much better if there were also a test added or modified to  drive out this need. @mikeroda

mikeroda:
There are currently no unit tests on the UaaStringUtils class and it seems silly to add one just for this obvious typo. This bug affects group names with those unescaped characters in them, particularly ^ since it is used in regular expressions. So I guess I could go and try to find some tests that deal with groups and add a group name with a ^ in it. I just don't have time at the moment.

strehle:
@jhamon I added a test and I would like to see a soon patch because [ ] are also not encoded which is bad for json

mikeroda:
@strehle thanks, I didn't notice that there was an existing unit test module under the server subproject, weird.

PR #784, opened 2018-03-02
Test for PR 781

strehle:
@jhamon  I created extra PR to #781  but the test shows the problem
see https://www.asciitable.com/ 
The rule A-z excludes A-Z and a-z but also all chars between Z - a [\]^_`


DennisDenuto:
Looks good to us. Thanks @strehle !

PR #786, opened 2018-03-05
Add and update migrations to ensure all tables have a primary key

APShirley:
The mysql clustered database in cf-deployment has previously allowed tables without primary keys, although this is not recommended. This will be changed soon to enforce primary keys as the lack of primary keys can cause issues with the clustered database. This pull request is intended as a starting point for conversation and potentially for cross-team efforts to address the issue of tables without primary keys.

With primary key enforcement enabled, older UAA migrations that update tables without primary keys will fail. Because of this we've modified older migrations to add the missing primary keys. We also added a new migration at the end that will add the primary keys on deployments that are upgrading and have already run the older migrations.

Important points:
- These are currently only for the mysql database. Although the primary keys are not needed for the other database engines as far as we know, they should probably be added there to keep the schemas in sync.
- Migration `V2_4_1_1__Add_primary_keys.sql` was added in the past, but no work was done to update the `schema_version` table to say that it was run. We saw that adding migrations to the `schema_version` table was done in `V4_0_4__FixFailedBackPortMigrations`. However, we saw that this "older migration" was simply skipped and did not cause errors, so we saw no reason to manually add it to the `schema_version` table.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/db/mysql/V4_9_2__AddPrimaryKeysIfMissing.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/db/mysql/V4_9_2__AddPrimaryKeysIfMissing.java
@@ -1,0 +1,22 @@
+package org.cloudfoundry.identity.uaa.db.mysql;
+
+import org.flywaydb.core.api.migration.spring.SpringJdbcMigration;
+import org.springframework.jdbc.core.JdbcTemplate;
+
+
+public class V4_9_2__AddPrimaryKeysIfMissing implements SpringJdbcMigration {
+
+    private final String checkPrimaryKeyExists = "SELECT COUNT(*) FROM information_schema.KEY_COLUMN_USAGE WHERE TABLE_SCHEMA = ? AND TABLE_NAME = ? AND CONSTRAINT_NAME = 'PRIMARY'";
+
+    @Override
+    public void migrate(JdbcTemplate jdbcTemplate) throws Exception {
+        String[] tables = {"group_membership", "external_group_mapping", "oauth_code", "sec_audit"};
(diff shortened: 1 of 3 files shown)


fhanik:
I would recommend not be modifying migration scripts that have been run in the past as they do not run again. This also avoids seeing database in different states depending on what version of code was used.

The change can be accomplished using new scripts, and recommend to have migrations for all database types.

jpalermo:
@fhanik: It's always preferable to not modify old migrations. Unfortunately the migration being modified both removes the primary key from a table, then runs an update statement on that same table.

With primary key enforcement turned on, this migration can never run and will fail on a clean installation.

fhanik:
@jpalermo Thank for the description. I understand the problem now. There simply is no way around the old scripts.

I would recommend that there is a script for `V4_9_2__AddPrimaryKeysIfMissing` for all databases. 

It will be difficult for developers of the UAA maintaining different schemas based on database type. Much simpler if the schema looks the same for all database types.

jpalermo:
Yeah, we definitely want the schemas to be in sync with each other. We didn't have the infrastructure to write and test the migrations for the other database engines when doing the PR.

Unsure if the changes from `V2_4_1` and the additions in `V2_4_1_1` should also be mirrored over to the other engines at the same place, or if those should be combined with the later `V4_9_2` migration in those cases. Both options have varying degrees of grossness. The mysql galera team is available to cross-team pair though.

DennisDenuto:
Merged into [develop](https://github.com/cloudfoundry/uaa/commit/c0f8016bbfd5b0eba81f3ba1b437970f8982b0f9)  with db migration tests and changes to postgres / sqlserver migration scripts to also contain primary keys.

PR #791, opened 2018-03-22
Updates the meaning of SCIM

asantosca:


DennisDenuto:
Oops my bad.. I see you the change from system to simple

DennisDenuto:
Merged into develop.

PR #806, opened 2018-04-21
Use zonifed cors configuration if configured.

Bitty9:
Currently, the identity zone configuration offers the ability to set a cors policy per zone. However, that cors policy is never used. Instead, the global cors policy is always used. This pull request uses the cors policy in the identity zone configuration if it differs from the global one. Additionally, since zonified cors policy is based on user input, we set a timeout for a long running regex match as well as a max timeout for matching with all patterns. In order to implement zonified cors policy, it was necessary to switch the filter order slightly so that the cors filter could access the correct policy for a given zone.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/security/web/CorsFilter.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/security/web/CorsFilter.java
@@ -16,6 +16,8 @@ package org.cloudfoundry.identity.uaa.security.web;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.cloudfoundry.identity.uaa.zone.CorsConfiguration;
+import org.cloudfoundry.identity.uaa.zone.CorsPolicy;
+import org.cloudfoundry.identity.uaa.zone.IdentityZoneHolder;
 import org.springframework.util.StringUtils;
 import org.springframework.web.filter.OncePerRequestFilter;
 
@@ -100,29 +102,9 @@ public class CorsFilter extends OncePerRequestFilter {
     @PostConstruct
     public void initialize() {
         for (CorsConfiguration configuration : Arrays.asList(xhrConfiguration, defaultConfiguration)) {
-            String type = (configuration == xhrConfiguration ? "xhr" : "default");
             configuration.getAllowedUriPatterns().clear();
             configuration.getAllowedOriginPatterns().clear();
-            if (configuration.getAllowedUris() != null) {
-                for (String allowedUri : configuration.getAllowedUris()) {
-                    try {
-                        configuration.getAllowedUriPatterns().add(Pattern.compile(allowedUri));
-                        logger.debug(String.format("URI '%s' is allowed for a %s CORS requests.", allowedUri, type));
-                    } catch (PatternSyntaxException patternSyntaxException) {
-                        logger.error("Invalid regular expression pattern in cors."+type+".allowed.uris: " + allowedUri, patternSyntaxException);
-                    }
-                }
-            }
-            if (configuration.getAllowedOrigins() != null) {
-                for (String allowedOrigin : configuration.getAllowedOrigins()) {
-                    try {
-                        configuration.getAllowedOriginPatterns().add(Pattern.compile(allowedOrigin));
-                        logger.debug(String.format("Origin '%s' is allowed for a %s CORS requests.", allowedOrigin, type));
-                    } catch (PatternSyntaxException patternSyntaxException) {
-                        logger.error("Invalid regular expression pattern in cors."+type+".allowed.origins: " + allowedOrigin, patternSyntaxException);
-                    }
-                }
-            }
+            compileCorsAllowedPatterns(configuration);
         }
     }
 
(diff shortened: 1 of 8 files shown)


6palace:
Hi, travis failed one job on a stalled build, we used to be able to restart travis jobs, but it looks like that privilege was revoked? Can one person in our org get travis authorities back for situations like this?

DennisDenuto:
@Bitty9  It looks like the `unit-test` CI jobs are red for changes related to this PR. Specifically:
```
org.cloudfoundry.identity.uaa.mock.DefaultConfigurationTestSuite > org.cloudfoundry.identity.uaa.mock.limited.LimitedModeLoginMockMvcTests.testXhrCorsPolicyforNonDefaultZone FAILED
    java.lang.AssertionError: Status expected:<200> but was:<403>
org.cloudfoundry.identity.uaa.mock.DefaultConfigurationTestSuite > org.cloudfoundry.identity.uaa.login.LoginMockMvcTests.testXhrCorsPolicyforNonDefaultZone FAILED
    java.lang.AssertionError: Status expected:<200> but was:<403>
```

@6palace wrt permission to retry jobs, I believe you need write access to the repo to allow that (I cannot see an option to allow the retrigger of jobs). As a workaround you could also close the PR and open it again.

server/src/main/java/org/cloudfoundry/identity/uaa/security/web/CorsFilter.java
```diff
@@ -142,15 +125,66 @@ protected void doFilterInternal(final HttpServletRequest request, final HttpServ
             logger.debug("CORS Processing request: "+getRequestInfo(request));
         }
         if (isXhrRequest(request)) {
-            handleRequest(request, response, filterChain, getXhrConfiguration());
+            handleRequest(request, response, filterChain, resolveXhrCorsConfiguration());
```
DennisDenuto: It looks like this will be called on every request that passes this filter. `resolveXhrCorsConfiguration` doesn't appear to be thread-safe (since it [clears](https://github.com/cloudfoundry/uaa/pull/806/files#diff-5b4ad26806da832e039024c9173adbaeR142) the uri patterns and then [adds valid regex](https://github.com/cloudfoundry/uaa/pull/806/files#diff-5b4ad26806da832e039024c9173adbaeR144) patterns.

I *think* this should be calling a different method to only return the identity zone cors policy (if it has one, otherwise use the 'global' one). This should suffice since regex patterns have already been compiled during the 'initialize' phase.

I think that the `PUT` and `POST` identity zone endpoints should [validate](https://github.com/cloudfoundry/uaa/blob/c1a10a915229e02b6e4f0e75f86efbb9271ed5f7/server/src/main/java/org/cloudfoundry/identity/uaa/zone/GeneralIdentityZoneConfigurationValidator.java#L28) that the cors policy is a valid regex pattern. If we do that we can assume that the cors policy from an identity zone will be valid (even if its created/updated during runtime) removing the need to compile the uri regex patterns on every request.

WDYT?

Bitty9: @DennisDenuto, sorry for the late response. The initialize call on this bean only occurs on startup of UAA. This means that we can't compile and store patterns which are configured per zone as that occurs after startup. Unless you mean go through all existing identity zones with cors policies at startup and compile them and store them in a zonified cache? However, the Pattern.compile call is only an optimization. The matcher call will always compile before it does the matching if Pattern.compile has not been called. I like the idea of validating the regex patterns at creation time even though  we would still need to recreate the compiled pattern object each request (unless using a zonified cache). Will update the PR pending your response.

DennisDenuto: Hi @Bitty9 thanks for the reply. I suppose its my turn to apologize for the late reply (this must have gotten lost when i went on vacation). Anyhoo, my concern actually is *not* on the `initialize` method. However it is for the `doFilterInternal` method. 

Specifically, if two identical requests are made in parallel, and if both of these requests reach this filter at the same time, then it seems to me that there is a data race condition problem here. 

i.e. 

Time | Req 1 | Req 2 | notes
------------ | ------------- | ------------- | -------------
0 | reaches filter | reaches filter | 
1 | clears out the patterns zoneXhrConfiguration.getAllowedUriPatterns().clear(); | |
2 | compileCorsAllowedPatterns | |
3 | | clears out the patterns zoneXhrConfiguration.getAllowedUriPatterns().clear();  |
4 | handleRequest with cors filter (refer to notes column)  | | Problem is here because I think the cors filter is now empty because of req 2 
5 | | compileCorsAllowedPatterns |
6 | | handleRequest with cors filter |

Also a reminder that there are also 2 failing tests:
```
org.cloudfoundry.identity.uaa.mock.DefaultConfigurationTestSuite > org.cloudfoundry.identity.uaa.mock.limited.LimitedModeLoginMockMvcTests.testXhrCorsPolicyforNonDefaultZone FAILED
    java.lang.AssertionError: Status expected:<200> but was:<403>
org.cloudfoundry.identity.uaa.mock.DefaultConfigurationTestSuite > org.cloudfoundry.identity.uaa.login.LoginMockMvcTests.testXhrCorsPolicyforNonDefaultZone FAILED
    java.lang.AssertionError: Status expected:<200> but was:<403>
```

Are you able to reproduce these failing tests locally on your machine?

shamus:
Closing due to inactivity.

PR #820, opened 2018-05-11
update oauth2

strehle:
https://pivotal.io/security/cve-2018-1260


DennisDenuto:
The change looks good, however CI is red, and it *seems* related to this change.  
```
org.cloudfoundry.identity.uaa.oauth.token.Saml2TokenGranterTest > test_no_grant_type FAILED
org.cloudfoundry.identity.uaa.oauth.token.UserTokenGranterTest > test_requesting_client_id_missing FAILED
```

tnwang:
We have also investigated this CVE previously and found that we were not affected. So there's no urgency behind this bump.

strehle:
@tnwang thanks.

however for develop you should prio the update story, because dependency scanners (ours and our customers one) get red now.

PR #822, opened 2018-05-11
update guava

strehle:
https://github.com/google/guava/wiki/CVE-2018-10237

### DIFF
--- a/shared_versions.gradle
+++ b/shared_versions.gradle
@@ -12,7 +12,7 @@ ext {
   esapiVersion = '2.1.0.1'
   flywayVersion = '4.2.0'
   googleauthVersion = '1.1.2'
-  guavaVersion = '19.0'
+  guavaVersion = '24.1.1-jre'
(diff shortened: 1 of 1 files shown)


DennisDenuto:
CI looks red: `Could not find com.google.guava:guava:24.1.1.`

Looking at maven central it looks like the version is missing '-jre'. i.e. It looks like it should be [compile group: 'com.google.guava', name: 'guava', version: '24.1.1-jre](https://mvnrepository.com/artifact/com.google.guava/guava/24.1.1-jre)

shared_versions.gradle
```diff
@@ -12,7 +12,7 @@ ext {
   esapiVersion = '2.1.0.1'
   flywayVersion = '4.2.0'
   googleauthVersion = '1.1.2'
-  guavaVersion = '19.0'
+  guavaVersion = '24.1.1'
```
DennisDenuto: looking at maven central the version should be 24.1.1-jre

PR #826, opened 2018-05-17
Fix to process expired codes periodically

mikeroda:
This change fixes a bug that was causing expired codes to be processed almost every time rather than periodically based on an expiration time, as it was designed. Without doing this the code will more likely encounter the dead locks under significant load.


shamus:
Closing due to inactivity.

PR #829, opened 2018-05-21
update bc

strehle:


tnwang:
Hi @strehle - Did you have any more information you could give about this change? Thanks!

strehle:
hi @tnwang , BouncyCastle is essential for uaa. there is no sec. issue, but it should be uptodate , see notes
https://www.bouncycastle.org/releasenotes.html

regards,
-markus

Issue #666, opened 2017-07-25
Spring placeholder in YAML is not processed


kannangce:
The placeholders in config file is not parsed for all the attributes.
I am able to use placeholders for DB attributes. Ex, `username: ${DB_USERNAME:postgres}` . But I am not able to do the same for `redirect-uri` . The placeholders that I give for the attribute `redirect-uri` is not replaced with the value of the placeholders.
To explain in detail,
I the `yml` block containing a client data, I gave the redirect-uri as `${STANDARD_REDIRECT_URI:http://localhost:8081/testapp}` . I expect this to get the value from the environment variable `STANDARD_REDIRECT_URI`, if it is not available then use the URL `http://localhost:8081/testapp` . But the whole string `${STANDARD_REDIRECT_URI:http://localhost:8081/testapp}` is saved as URL in DB.

jhamon:
Could you give some more details about the exact steps you are following, the behavior you expect, and the behavior you are seeing? There's not enough information in this issue for me to reproduce and troubleshoot.  

If you're only seeing this issue with `redirect-uri`, I suspect it might be related to the specific url you are trying to set; for security reasons, there are some limitations on what URL's may be configured.

kannangce:
@jhamon  Added details to the question.

PR #849, opened 2018-06-14
For cloudfoundry/uaa#666

kannangce:
*Problem:* Currently the `environment.resolvePlaceholders` is called only for the type string. In reality the values are of type Map, List.
*Fix:* Added logic to resolve the values.
*Reference:* https://github.com/cloudfoundry/uaa/issues/666

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/impl/config/EnvironmentMapFactoryBean.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/impl/config/EnvironmentMapFactoryBean.java
@@ -82,16 +82,43 @@ public class EnvironmentMapFactoryBean implements FactoryBean<Map<String, ?>>, E
                     EnumerablePropertySource enumerable = (EnumerablePropertySource) source;
                     for (String name : enumerable.getPropertyNames()) {
                         Object value = source.getProperty(name);
-                        if (value instanceof String) {
-                            // Unresolved placeholders are legal.
-                            value = environment.resolvePlaceholders((String) value);
-                        }
-                        result.put(name, value);
+                        result.put(name, resolvePlaceHolders(value));
                     }
                 }
             }
         }
         return result;
+    }
+    
+    /**
+     * Try to resolves the place-holders in the given object. In case of any
+     * exception, returns the object as is.
+     * 
+     * @param value
+     *            Any value in which the object need to be resolved.
+     * @return The given Object instance with the values resolved.
+     */
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+    private Object resolvePlaceHolders(Object value){
+        if (value instanceof String){
+            // Unresolved placeholders are legal.
+            return environment.resolvePlaceholders((String) value);
+        }
+        try{
+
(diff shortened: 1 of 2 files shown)


DennisDenuto:
@kannangce This PR currently [does not compile](https://travis-ci.org/cloudfoundry/uaa/jobs/392091431) 
```
:cloudfoundry-identity-server:compileJava FAILED
:cloudfoundry-identity-server:compileJava took 11922ms
```

kannangce:
@DennisDenuto  Sorry for the missed `git add`. Now made sure everything got committed and pushed. It builds and runs in my local.

On the testing part, I tried parameterizing the value of `oauth.clients.*.redirect-uri`  from `uaa.yml` and passed the value in environment variable. Which got picked up and the client created with the respective values from the environment variable. Before the fix that was not working.

kannangce:
@DennisDenuto Any other comments that I need to check?

DennisDenuto:
Hi @kannangce Thanks for explaining how to use this new env config logic. 

Can you please provide some tests for the new logic in `EnvironmentMapFactoryBean`. You can add unit tests here: https://github.com/cloudfoundry/uaa/blob/31a6dfba592acad3652b24ecdbb3fb7c6c0eac78/server/src/test/java/org/cloudfoundry/identity/uaa/config/EnvironmentMapFactoryBeanTests.java#L32-L32

Some tests to add might be (these are pseudocode examples):
# "happy case tests"
1.1) a property that is a map
given: System.setProperty("some_env", "some_env_value");
input map: 
```
fake-client:
  key1: ${some_env}
```
expected output:
```
fake-client:
  key1: some_env_value
```

1.2) a property that is a map
input map: 
```
fake-client:
  key1: ${spam}
```
expected output:
```
fake-client:
  key1: ${spam}
```

2.1) a property that is a list
given: System.setProperty("some_env", "some_env_value");
input list: 
```
- fake-client: ${some_env}
```
expected output:
```
- fake-client: some_env_value
```

2.2) a property that is a list
input list: 
```
- fake-client: ${spam}
```
expected output:
```
- fake-client: ${spam}
```

# error cases:
1.1) a property that is a map with null values
input map: 
```
fake-client:
  key1: null
```
expected output:
```
fake-client:
  key1: null
```

1.2) a property that is a list with null values
input list: 
```
[ ${some_env}, null ]
```
expected output:
```
[ ${some_env}, null ]
```

kannangce:
@DennisDenuto Thanks for the suggestion!

Added and verified the tests.

The below are the details of the tests that I added,

**`testPlaceholderResolutionFor{X}Value`** - To test the resolution of the placeholder from System environment.
**`testPlaceholderResolutionForNoEnv{X}Value`** - To test the case when no environment for the resolvable placeholder present.
**`testPlaceholderResolutionFor{X}NullValue`** - To test the null values in the values.

Where **{X}** can be a simple value, List or Map.

Please verify and let me know if things are good.

DennisDenuto:
Nice. LGTM.
@jhamon WDYT? Can you think of any security concerns with regards to this change?

shamus:
Closing due to inactivity.

PR #850, opened 2018-06-15
update jackson

strehle:
findings in vula scan
CVE-2018-11307
CVE-2018-12022
CVE-2018-12023


jhamon:
@strehle I'm not finding much information about these CVEs online but it seems simple enough to patch bump. For future security related concerns, please go through security@cloudfoundry.org

strehle:
thank you, @jhamon the CVEs are listed here https://github.com/FasterXML/jackson-databind/blob/master/release-notes/VERSION-2.x and thus they are public I dont see a reason for extra mails. If we see a critical issue  within uaa then we follow this process

PR #832, opened 2018-05-23
update hibernate

strehle:
CVE-2017-7536

PR #861, opened 2018-07-10
update javax.validation 

strehle:
since version 1.0.0 is not compatible with hibernate 4.3.x

follow-up to https://github.com/cloudfoundry/uaa/pull/832


DennisDenuto:
@jhamon Might be worth looking to see if any of the new bean validation library can be leveraged for the token validator class
https://beanvalidation.org/news/2017/08/07/bean-validation-2-0-is-a-spec/

PR #869, opened 2018-07-16
Prevent unnecessary exceptions from being thrown and showing up in logs

pivotal-ben-chacko:
Previous logic would not check to see if invalid char's were present in the string until after an exception was thrown.  This led to many unnecessary and misleading exceptions showing up in the debug logs.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/resources/jdbc/SimpleSearchQueryConverter.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/resources/jdbc/SimpleSearchQueryConverter.java
@@ -173,14 +173,12 @@ public class SimpleSearchQueryConverter implements SearchQueryConverter {
     }
 
     protected SCIMFilter scimFilter(String filter) throws SCIMException {
-        SCIMFilter scimFilter;
-        try {
-            scimFilter = SCIMFilter.parse(filter);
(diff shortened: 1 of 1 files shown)


jhamon:
Hi @pivotal-ben-chacko.  I'm not sure I understand the problem you're trying to address with this change. It seems like you would see just as many log lines after this change as before except now the log message is (misleadingly) logging the modified filter query. The tests failing on travis show the misleading exception message issue:

```
org.cloudfoundry.identity.uaa.scim.jdbc.ScimSearchQueryConverterTests > canConvertValidFilters FAILED
    java.lang.IllegalArgumentException: Invalid SCIM Filter:username eq "'bar" Message:Invalid filter 'username eq ""bar"': End of input at position 18 but expected an attribute operator
```

Is there some kind of exception other than `SCIMException` that you think we should be catching here?

pivotal-ben-chacko:
The problem I was trying to address was that anytime the character " ' "
was part of the string being parsed there would be an exception thrown
because it was considered an unsupported character by the parser.  We would
then catch this exception, remove the unsupported char and re-process the
string.

The code change was to preemptively remove the unsupported character from
the string before sending it through the parser which would avoid an
exception being thrown unnecessarily.

Ben.

shamus:
Closing due to inactivity.

Issue #810, opened 2018-04-25
Token request should not include "response_type" per Oauth2 RFC


aashah:
### What version of UAA are you running?

`"version": "4.13.0-SNAPSHOT"`

### How are you deploying the UAA?

I am deploying the UAA using cf-deployment

### What did you do?

Read the docs here: http://docs.cloudfoundry.org/api/uaa/version/4.12.0/#token

### What did you expect to see? What goal are you trying to achieve with the UAA?

We expect to **not** see `response_type` as a request parameter for this endpoint.

Additional notes:

This is to be compliant with the Oauth2 OAuth 2.0 Authorization Framework. Relevant section is: [Access Token Request](https://tools.ietf.org/html/rfc6749#section-4.1.3)

- Aakash & @peterhaochen47

shamus:
Closing due to inactivity.

Issue #809, opened 2018-04-25
Authorization code token response (/oauth/token) is not compliant with OpenID connect 1.0 RFC by not including "id_token"


aashah:
### What version of UAA are you running?

`"version": "4.13.0-SNAPSHOT"`

### How are you deploying the UAA?

I am deploying the UAA using cf-deployment

### What did you do?

Given a client with "openid" scope, and "authorization_code" grant type.
We followed the authoritization_code grant flow and inspected the response from `/oauth/token`

### What did you expect to see? What goal are you trying to achieve with the UAA?

We expect to see `id_token` as part of the response, alongside `access_token` and `refresh_token`. Something like:

```
{
  "access_token": "<token>",
  "id_token": "<token>",
  "token_type": "bearer",
  "refresh_token": "<token>",
  "expires_in": 43199,
  "scope": "openid",
  "jti": "4f09a0e76ed743f9b70623e81d048fd5"
}
```

### What did you see instead?

```
{
  "access_token": "<token>",
  "token_type": "bearer",
  "refresh_token": "<token>",
  "expires_in": 43199,
  "scope": "openid",
  "jti": "4f09a0e76ed743f9b70623e81d048fd5"
}
```

Additional notes:

This is to be compliant with the OpenID Connect 1.0 RFC. Relevant section is [3.1.3.3.  Successful Token Response](http://openid.net/specs/openid-connect-core-1_0.html#TokenResponse)

- Aakash & @peterhaochen47

shamus:
Closing due to inactivity.

Issue #777, opened 2018-02-23
Support id_token in refresh_token


making:
### What version of UAA are you running?

``` json
{
  "app": {
    "version": "4.10.0"
  },
  "links": {
    "uaa": "https://uaa.<...>",
    "login": "https://login.<...>"
  },
  "zone_name": "uaa",
  "entityID": "login.<...>",
  "commit_id": "a4132cb",
  "idpDefinitions": {},
  "prompts": {
    "username": [
      "text",
      "Email"
    ],
    "password": [
      "password",
      "Password"
    ]
  },
  "timestamp": "2018-02-06T10:24:36-0800"
}
```

### How are you deploying the UAA?

using cf-deployment 1.15.0 on bosh-lite

### What did you do?

I'm trying to integrate UAA and Kubernetes with CFCR 0.14.0 using https://github.com/cloudfoundry-incubator/kubo-release/pull/101 .

It worked to access k8s with `id_token` returned by UAA.
After the token expired, `kubctl` used `refresh_token` to refresh the token but failed with the following error

```
Unable to connect to the server: token response did not contain an id_token, either the scope "openid" wasn't requested upon login, or the provider doesn't support id_tokens as part of the refresh response.
```

From [the doc](https://docs.cloudfoundry.org/api/uaa/version/4.8.0/index.html#refresh-token),
refresh token does not seem to support id token.

I've also confirmed the response of bellow didn't contain `id_token` field

```
curl -k -v https://uaa.<...>/oauth/token -d client_id=xxxxx -d client_secret=xxxxx -d grant_type=refresh_token -d refresh_token=xxxxx -d response_type=id_token
``` 

### What did you expect to see? What goal are you trying to achieve with the UAA?

I expected UAA to support id_token in refresh_token in order to refresh token with `kubectl`.

making:
@tnwang 

Can you please prioritise this? I believe this is important for k8s integration.

https://kubernetes.io/docs/admin/authentication/#option-1-oidc-authenticator

> Once your id_token expires, kubectl will attempt to refresh your id_token using your refresh_token and client_secret storing the new values for the refresh_token and id_token in your .kube/config.

tnwang:
Thanks for raising this issue. We are aware of this and will be working on it soon.

making:
is this issue fixed in 4.20.0?

jhamon:
@making No, it's not fixed in 4.20.0. I have work in progress for this that I plan to PR for the OD-PKS team.

[...]

PR #902, opened 2018-09-20
Remove non-spec response_type param from /oauth/token endpoint

jhamon:
Hey team. Work in this PR addresses the following issues:

#810 response_type param should not be required with /oauth/token
#777 id_token refresh working for k8s
#809 UAA returns id_token from /oauth/token to clients that the UAA grants openid scope without requiring response_type=id_token param

I also updated docs in c1b7d5b and removed this parameter from tests where it no longer has any effects (8585edd, 762f4bb) . `/oauth/authorize` usages of `response_type` param were not changed.

I have not done my own testing/acceptance with kubernetes yet, but all tests are passing locally and I am able to get id_tokens during refresh flow using the client we are using in PKS:

```
uaac token client get admin -s adminsecret
uaac client add pks_cluster_client --authorized_grant_types password --secret "" --scope openid,roles
uaac token owner get pks_cluster_client marissa -s "" -p koala
uaac context # see the id_token. Grab the refresh token, put it in a variable called REFRESH_TOKEN.
curl -X POST http://localhost:8080/uaa/oauth/token -d"grant_type=refresh_token&client_id=pks_cluster_client&client_secret=&refresh_token=$REFRESH_TOKEN"
```


server/src/main/java/org/cloudfoundry/identity/uaa/oauth/DisableIdTokenResponseTypeFilter.java
```diff
@@ -34,7 +34,6 @@
 
 public class DisableIdTokenResponseTypeFilter extends OncePerRequestFilter {
 
-    public static final String CONFIG = "oauth.id_token.disable";
```
jhamon, line 37: This was unused so deleted it even though it's not really part of my change.

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/token/TokenMvcMockTests.java
```diff
@@ -3071,7 +3044,7 @@ public void testLoginAuthenticationFilter() throws Exception {
                 .param("username", developer.getUserName())
                 .param("user_id", developer.getId())
                 .param(OriginKeys.ORIGIN, developer.getOrigin()))
-                .andExpect(status().isUnauthorized());
+                .andExpect(status().isForbidden());
```
DennisDenuto, line 3047: How come the status changed from a 401 -> 403?

jhamon: I think removal of the non-spec response_type param from the request matcher [here](https://github.com/cloudfoundry/uaa/blob/b227e8302bacc9af510f0a7c27cce579e33dd859/uaa/src/main/webapp/WEB-INF/spring/oauth-endpoints.xml#L166-L178) meant that these requests started getting caught in the [statelessTokenApiSecurity policy](https://github.com/cloudfoundry/uaa/blob/b227e8302bacc9af510f0a7c27cce579e33dd859/uaa/src/main/webapp/WEB-INF/spring/oauth-endpoints.xml#L184-L194) for the first time whereas before they would have matched [loginTokenRequestMatcher (which I think should be removed entirely ASAP)](https://github.com/cloudfoundry/uaa/blob/ab292e78b239e01d4c5dfcccb8c372b6bae1fc29/uaa/src/main/webapp/WEB-INF/spring/login-server-security.xml#L111-L126). Since the tokens in these tests do not have the `uaa.user` scope required by the stateless api security policy they get 403; even if they requested another unexpired token it would not allow them to successfully pass this filter.

jhamon:
Is there anything you want changed or reworked in this PR? From Tracker is appears there are concerns about `/oauth/authorize` but I believe the `/oauth/token` changes in this PR stand on their own. @DennisDenuto

DennisDenuto:
@jhamon we initially added a  list containing acceptance criteria from  issues:  [#810, #777, #809] and the tracker  https://www.pivotaltracker.com/story/show/160260269

However https://www.pivotaltracker.com/story/show/160260269 has been split out to remove any reference to `/authorize` (We have updated the  list)

After further analysis on the change made to the TokenMvcMockTests.java, we believe that some of these test are mis-configured. 
e.g. https://github.com/cloudfoundry/uaa/blob/4d9e0108e367c80238dd5cf595660cb14325d29c/uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/token/TokenMvcMockTests.java#L3279-L3287 provides correct credentials, except expects a `401`. The reason being that it is missing the `Content-Type` header. 

With [this commit](https://github.com/cloudfoundry/uaa/commit/ce652a3c482d11626a1af57a1bdd2dd6ba3d2d3e
) (that tries to fix the test config) The change in this PR *does* return a `401` is credentials are invalid - which we believe is the correct behavior. 

fyi: We are in the process of backfilling some tests before merging in this change.

jhamon:
Thanks @DennisDenuto :partyparrot:

DennisDenuto:
Thanks @jhamon && @pgoodwin for the PR!

Issue #688, opened 2017-09-01
Public clients & PKCE


sebastianGit:
Is it planned that CF UAA supports public clients (clients without any secret) using "token_endpoint_auth_method"="none" (RFC 7591) and PKCE (RFC 7636) ?

Thanks,
Sebastian

jhamon:
We have implemented some parts of RFC 7591 (specifically the `urn:ietf:params:oauth:grant-type:jwt-bearer` and `urn:ietf:params:oauth:grant-type:saml2-bearer` grant types) but we do not currently have plans to tackle the `token_endpoint_auth_method` soon. We'll take this as a feature request.

The workaround you can do today is to create a second non-public client that will only be used for calling the check_token endpoint.

Issue #455, opened 2016-10-03
Support PKCE (Proof Key for Code Exchange)


gberche-orange:
As part of using UAA for mobile SSO use-cases (see https://tools.ietf.org/html/draft-ietf-oauth-native-apps-03), the ability to [protect authorization codes](https://tools.ietf.org/html/draft-ietf-oauth-native-apps-03#section-8.2) relies on Proof Key for Code Exchange (PKCE) support specified in https://tools.ietf.org/html/rfc7636

The [UAA api](http://docs.cloudfoundry.org/api/uaa/) does not seem to yet support the related `code_challenge` endpoint parameters of PKCE as an example.

Is there anything planned to support PKCE in the future ?

sreetummidi:
No plans to implement this at this time.

markmabe:
@sreetummidi Hi is there a reason there are no plans to implement PKCE in UAA?

minyangu:
@sreetummidi 
Hi, same question, why UAA has no plan to implement the PKCE for mobile apps? 

Please also note this:
The OAuth 2.0 Security Best Current Practice document recommends **against using the Implicit flow entirely**, and OAuth 2.0 for **Browser-Based Apps** describes the technique of using the **authorization code flow with PKCE**.

Although this is still under draft status, this may soon become a confirmed standard by IETF so that browser-based apps like SPA are also required to use the Auth Code + PKCE flow mechanism.

SHRIVP:
@sreetummidi ,

Does UAA support PKCE now?

PR #939, opened 2019-01-22
PKCE support according to RFC 7636

ZMaradics:
This PR implements PKCE support according to RFC 7636 (https://tools.ietf.org/html/rfc7636) which is current best practice when using OAuth2.0 in the context of mobile apps (RFC 8252 https://tools.ietf.org/html/rfc8252). This has already been requested in https://github.com/cloudfoundry/uaa/issues/455 and https://github.com/cloudfoundry/uaa/issues/688.
The authorize endpoint accepts code_challenge and code_challenge_method parameters (supported methods: plain and S256).
If these parameters are present the token endpoint will require and check the code_verifier parameter.
As org.springframework.security.oauth2.provider.code.AuthorizationCodeTokenGranter does not support PKCE up to now we created an enhanced version: PkceEnhancedAuthorizationCodeGranter


fhanik:
Thanks for the PR could you please

1. Squash your commits so it's less reviews per commit
2. Remove all copyright modifications

ZMaradics:
Done.
Let me know is anything else required.

sebastianGit:
Do you still consider accepting this merge request? In case, we can support the "integration" / "merge" of this pull request, please let us know.

cherukumilli (approved):

cherukumilli:
@ZMaradics 
Can you please help resolve the conflicts?

ZMaradics:
Hi, Sorry for the late answer, Yes I can help.

ZMaradics:
I fixed the conflicts and did unit and integration tests with no fails.

cherukumilli:
@fhanik - Can you please review this PR and help merge it?

cherukumilli:
@ZMaradics - can you please sign the EasyCLA?

ZMaradics:
Done

cherukumilli:
@fhanik 
Can you pls review this PR and see if it can merged?

sebastianGit:
@cherukumilli, @fhanik: is there any news on this topic?

fhanik:
@sebastianGit I'm not longer on this project, and will have to defer to the current contributors directly

strehle (approved):
@dbeneke  to me this PR is reasonable and should be merged

PG2000:
@dbeneke is there any news on this topic?

sjw1066:
What is the status of this? Lack of this functionality is preventing us from using UAA on a number of projects that require PKCE.

PG2000:
@cf-identity  Is there any news on this topic? Would love to see more transparent discussion on that?

i059917:
Hello, any progress on this? We have cases need this potentially, it will be great we can know if there is a draft timeline of this can be enabled. Thanks.

bsekar:
@strehle @shamus, any updates on this? Implicit with response_type token is not recommended anymore. PKCE is the way to go for single page apps.

strehle:
Hi, if @shamus has no VETO (anymore) I will merge it . 

The current issue within UAA is, that many things should move from https://spring.io/blog/2019/11/14/spring-security-oauth-2-0-roadmap-update to new spring security 5 and then this PR could be done in a different way, however I see the feature and implementation itself worth to be merged, even if we need to refactor many things in UAA aways from spring-security-oauth2 towards new framework

shamus:
Hi y'all. I know there's a lot of interest in this feature but our focus is on keeping the UAA stable for cf-for-k8s. We do not want to merge this at this time. 

I'm inclined to close this issue until we're actually ready to make this feature set a priority.

sebastianGit:
Looking at the evolution of OAuth, the need to use PKCE is continuously increasing:

- OAuth 2.0 for Native Apps (RFC 8252): PKCE for mobile and native Apps with public clients

- OAuth 2.0 for Browser-Based Apps (draft-ietf-oauth-browser-based-apps-06): PKCE for SPAs without backend making use of public clients

-  The OAuth 2.1 Authorization Framework (draft-ietf-oauth-v2-1-00): PKCE for all kind of OAuth clients (incl. confidential clients)

All three documents are asking for the OAuth authorization server to support PKCE.

Could you share any planned timeline when CF UAA is to provide PKCE suppport (independent from whether this pull request gets merged or not)

bsekar:
> Looking at the evolution of OAuth, the need to use PKCE is continuously increasing:
> 
> * OAuth 2.0 for Native Apps (RFC 8252): PKCE for mobile and native Apps with public clients
> * OAuth 2.0 for Browser-Based Apps (draft-ietf-oauth-browser-based-apps-06): PKCE for SPAs without backend making use of public clients
> * The OAuth 2.1 Authorization Framework (draft-ietf-oauth-v2-1-00): PKCE for all kind of OAuth clients (incl. confidential clients)
> 
> All three documents are asking for the OAuth authorization server to support PKCE.
> 
> Could you share any planned timeline when CF UAA is to provide PKCE suppport (independent from whether this pull request gets merged or not)

In fact, OAuth2.1 is going to outright omit Implicit (access_token) https://tools.ietf.org/html/draft-ietf-oauth-v2-1-00

Richard-Boyewa:
@shamus @strehle  what is the resolution on this pull request. We are currently planning to implement PKCE in our authentication layer, and since we already use UAA, I think it is pointless to implement another layer on UAA just to be PKCE compliant.

Do we have a possible green light (even if this would be refactored later) or it is going to be closed?

MrWhiteABEX:
@strehle As a SAP customer I can only say please merge as soon as possible and bring it to SAP Cloud Platform. Not supporting PKCE is a no go. As SAP pushes native apps (SAP MDK, etc.) this is a must have.

strehle:
@MrWhiteABEX  
PKCE is already supported in SAP Cloud Platform, see https://accounts.sap.com/.well-known/openid-configuration , this is no uaa but SAP own OIDC server. Then there are other OIDC servers, like UAA.
This is the community UAA, SAP has a fork where we will support it if it does not reach community UAA.

My Hope is, that after the CF summit there is time for further discussions, because I know that other vendors also simply fork UAA and add features.

MrWhiteABEX:
@strehle Glad to hear that at least SAP ID Service does. But frankly I am not building apps for managing my sub accounts and spaces ;). That is my biggest problem right know. I can host API´s in SCP Cloud Foundry using Java, NodeJS etc. and access on-Premise systems securely using Cloud Connector but I cannot securely consume them securely. The (XS)UAA does not support code challenges and I have to provide the client-secret all the time and store it in my apps.
Side note: Common authentication libraries like AppAuth can obtain authorization codes but fail to obtain tokens. That part I have to code myself to make it work. Don´t know whose fault that is.
There is also an artificial length limit in the (XS-)UAA redirect filter (You can use arbitrary lengths in the actual code grant) which breaks Universal Windows apps because their platform redirect links are too long. The workaround is to use only a wildcard e.g. "ms-app://**". That is something I already reported to SAP support.
Current state of UAA seems to be a complete mess when you want to do something different than hosting web apps. 
I really think about migrating everything to Azure AD and Azure App Service because of this.

Sorry for the rant and for hijacking the pull request :(

sebastianGit:
Can you please provide any hint, when/whether CF UAA will support PKCE? This pull request has been around for two years now.

rwwilden:
At a VMware customer right now where we are blocked by missing PKCE support in UAA. Implicit flow isn't an options because they are on pretty modern browsers that limit third-party cookie support. As asked before, can anyone provide a hint/timeline for when this is available?

strehle:
> As asked before, can anyone provide a hint/timeline for when this is available?

The PR itself is mergeable, so no technical issues, but to me a task about who takes the responsibility to merge it.
if @aramprice  @peterhaochen47  @staylor14  does not rise any veto, I can merge

peterhaochen47:
@strehle I will bring this with my team early next week.

strehle:
@ZMaradics after all this time we want merge your PR.
I only want ask you if you agree, that we use Default S256 in case code method is null (not set in reuqest) ?

ZMaradics:
@strehle: As RFC7636 ch. 4.3 (https://datatracker.ietf.org/doc/html/rfc7636#section-4.3) and also the upcoming draft for OAuth2.1 ch 4.1.1.1 (https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-02#section-4.1.1.1) specify that “plain” is the default if the code_challenge_method parameter is missing, I think we should stick to the specifications as long as there is no really important reason to deviate.
If the reason is to avoid the “plain” method it would be better to not support it at all (the consequence would be, that code_challenge_method must be present if a code_challenge is present or else an error must be returned)

strehle:
> If the reason is to avoid the “plain” method it would be better to not support it at all (the consequence would be, that code_challenge_method must be present if a code_challenge is present or else an error must be returned)

**If you agree to omit plain complety, then that would be ok for me and the following discussion could be avoided, but if you want stay with both, plain and S256, then I vote for simply switching the default.**

The reasons for me are
1. only with code_challange_method = S256 you protected in token request because only the original requestor can created the HMAC. With plain it sounds like security but it isn´t security, more security, then nonce have. That sounds to me similar to the issue in JWT validation (at the beginning) with type "alg" of JWT header; standard allows none but this caused problems, e.g. https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/  
2. I vote for secure by default and as elaborated in 1. only S256 is secure
3. In next step I am sure that we want use PKCE in combination with public clients (which we currently dont have in UAA) in order to omit client secrets but here S256 should be the default (to me) and therefore I vote to have this default now with introduction of PKCE 

I dont want reject plain but only switch the default to S256.
Finally I dont want reject your PR because of my opionion so if most voters also want stay then we can do it, only added some document issues I see

uaa/src/test/java/org/cloudfoundry/identity/uaa/login/TokenEndpointDocs.java
```diff
@@ -101,6 +104,7 @@
...
     private final ParameterDescriptor loginHintParameter = parameterWithName("login_hint").optional(null).type(STRING).description("<small><mark>UAA 4.19.0</mark></small> Indicates the identity provider to be used. The passed string has to be a URL-Encoded JSON Object, containing the field `origin` with value as `origin_key` of an identity provider. Note that this identity provider must support the grant type `password`.");
+    private final ParameterDescriptor codeVerifier = parameterWithName(PkceValidationService.CODE_VERIFIER).description("<small><mark>UAA 4.26.0</mark></small> [PKCE](https://tools.ietf.org/html/rfc7636) Code Verifier. A `code_verifier` parameter must be provided if a `code_challenge` parameter was present in the previous call to `/oauth/authorize`. The `code_verifier` must match the used `code_challenge` (according to the selected `code_challenge_method`)").attributes(key("constraints").value("Optional"), key("type").value(STRING));
```
strehle: next uaa will be 75.5.0, can you please update this , would cleanup the PR, thanks

uaa/src/test/java/org/cloudfoundry/identity/uaa/scim/endpoints/OpenIdConnectEndpointDocs.java
```diff
@@ -33,6 +33,7 @@ void getWellKnownOpenidConf() throws Exception {
                 fieldWithPath("claims_supported").description("JSON array containing a list of the Claim Names of the Claims that the OpenID Provider MAY be able to supply values for."),
                 fieldWithPath("claims_parameter_supported").description("Boolean value specifying whether the OP supports use of the claims parameter."),
                 fieldWithPath("service_documentation").description("URL of a page containing human-readable information that developers might want or need to know when using the OpenID Provider."),
+                fieldWithPath("code_challenge_methods_supported").description("<small><mark>UAA 4.31.0</mark></small>JSON array containing a list of [PKCE](https://tools.ietf.org/html/rfc7636) code challenge methods supported by this authorization endpoint."),
```
strehle: next uaa will be 75.5.0, can you please update this , would cleanup the PR, thanks

strehle (approved):

sebastianGit:
@strehle : Here my thoughts on the discussion which default to use, if the code_challenge_method is not specified in the authorization request:

As the behavior of the default is explicitly defined in the standard (and in the current version of the OAuth 2.1 draft), I believe it is best, if CF UAA implements PKCE as described in the standard (RFC 7636) and not just "close to the standard" in order to ensure compatibility with 3rd Party libs.

If there is a need for having "PKCE-like behavior" using sha256 as default (if the code_challenge_method is missing in the authorization request), a way out could be to make the default configurable as part of uaa.yml.

strehle:
@ZMaradics  you need to accept CLA, https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/1797134/3382195/939/#/?version=2

ZMaradics:
Done

PR #940, opened 2019-01-22
Fixes issue where migrations fail using MySQL8

aryanet:
When using use-external-db.yml in cf-deployment, and with MySQL8 engine, this query fails during running initial setup complaining about syntax error. Escaping the table name seems to fix it.


lasdolphin:
Hello everyone, may I kindly ask if it's gonna be merged or not?
Latest UAA docker build is still not able to run on MySql 8.

strehle:
Hi, I would merge it , but I cannot judge if it break other mysql users , we use use postgresql 
@bruce-ricard  / @peterhaochen47  in vmware you use mysql ? so you should decide here

strehle:
based on this PR / knowledge create a PR with all related fixes., e.g. flyway upgrade is also needed
https://github.com/cloudfoundry/uaa/pull/1873

PR #949, opened 2019-02-19
 Bootstrap configuration of MFA identity providers for default UAA zone

dre2901:
Support configuring identityProviders available for MFA for the default uaa identity zone. Otherwise it is not possible to configure the list of identity providers for MFA neither through uaa.yml nor via using REST API (update uaa identity zone is not allowed via REST).
And in many cases default uaa zone is easier to setup and use rather than creating brand new custom zones.


shamus:
Closing due to inactivity.

PR #962, opened 2019-03-23
ClientAdminBootstrap should not use IdentityZoneHolder

joshuatcasey:
See https://www.pivotaltracker.com/story/show/164842699


joshuatcasey:
See also https://www.pivotaltracker.com/story/show/164852169

Issue #967, opened 2019-04-24
SAML logout issue if IDP initiated


strehle:
### What version of UAA are you running?
4.24, but guess does not matter if newer

### How are you deploying the UAA?

bosh

### What did you do?

Use 2 UAA instance together with one IDP. If I start SLO with /logout.do the first UAA instance is 
performing logout but the 2nd one we see 

<saml2p:LogoutResponse ... >
    ...
    <saml2p:Status>
        <saml2p:StatusCode 
            Value="urn:oasis:names:tc:SAML:2.0:status:UnknownPrincipal"/>
        <saml2p:StatusMessage>No user is logged in</saml2p:StatusMessage>
    </saml2p:Status>
</saml2p:LogoutResponse>
similar to  https://github.com/spring-projects/spring-security-saml/issues/166

### What did you expect to see? What goal are you trying to achieve with the UAA?

Logout on all UAA instances

### What did you see instead?

Problem is that authentication is null, see
https://github.com/spring-projects/spring-security-saml/blob/1.0.9.RELEASE/core/src/main/java/org/springframework/security/saml/SAMLLogoutProcessingFilter.java#L169-L173

Security context is not set thereofore always null
https://github.com/spring-projects/spring-security-saml/blob/1.0.9.RELEASE/core/src/main/java/org/springframework/security/saml/SAMLLogoutProcessingFilter.java#L169

tack-sap:
Merged the PR. Thus closing the issue.

PR #968, opened 2019-04-24
Ensure SecurityContext is set before entering in LogoutProcessingFilter

strehle:
details in https://github.com/cloudfoundry/uaa/issues/967
#967

### DIFF
--- a/uaa/src/main/webapp/WEB-INF/spring/saml-providers.xml
+++ b/uaa/src/main/webapp/WEB-INF/spring/saml-providers.xml
@@ -35,7 +35,7 @@
                 <security:filter-chain pattern="/saml/metadata/**" filters="metadataDisplayFilter" />
                 <security:filter-chain pattern="/saml/idp/metadata/**" filters="idpMetadataDisplayFilter" />
                 <security:filter-chain pattern="/saml/SSO/**" filters="samlSecurityContextPersistenceFilter,samlWebSSOProcessingFilter" />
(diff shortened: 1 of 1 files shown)


strehle:
@torsten-sap  / @tack-sap  prefixed this for our staging landscapes and could verfy there that SAML SLO did not work in the past ( all the time ;) ) but now if more systems are enable to SAML SSO all are logout out with SAML SLO flow

uaa/src/main/webapp/WEB-INF/spring/saml-providers.xml
```diff
@@ -35,7 +35,7 @@
                 <security:filter-chain pattern="/saml/metadata/**" filters="metadataDisplayFilter" />
                 <security:filter-chain pattern="/saml/idp/metadata/**" filters="idpMetadataDisplayFilter" />
                 <security:filter-chain pattern="/saml/SSO/**" filters="samlSecurityContextPersistenceFilter,samlWebSSOProcessingFilter" />
-                <security:filter-chain pattern="/saml/SingleLogout/**" filters="samlLogoutProcessingFilter" />
+                <security:filter-chain pattern="/saml/SingleLogout/**" filters="samlSecurityContextPersistenceFilter,samlLogoutProcessingFilter" />
```
strehle, line 38: @fhanik  can you please merge this. Issue occurs in scenarios where more than 1 SAML party is involved and if SLO is triggered at this endpoint the IDP initiated the logout.
If this bean is not set here the session is not visible to the logout procedure and therefore issue #967

PR #978, opened 2019-05-21
jackson update 2.9.9

strehle:
why, see https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.9.9


andrewedstrom:
Thanks for the PR and for putting this on our radar! We changed the way we store shared versions (they're now in `dependencies.gradle`), but I've gone ahead and updated the jackson version there and pushed that change.

PR #979, opened 2019-05-21
Fix missing static resources for invalid subdomains

firnkes:
Do not check for invalid subdomains in case static content
is requested.

Enables serving of css and images on error page for invalid subdomain.


strehle:
@wc22222  @torsten-sap  see issue if you open https://test.login.run.pivotal.io

PR #1032, opened 2019-06-16
Feature/bypass 20second integration test startup time

taitzariel:
By moving comments out of uaa.yml, the 20 second integration test startup time is overcome.

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -54,7 +54,7 @@ dependencies {
     compile group: 'org.mariadb.jdbc', name: 'mariadb-java-client', version: '2.2.0'
     compile group: 'org.hsqldb', name: 'hsqldb', version: "2.5.0"
 
-    compile group: 'org.yaml', name: 'snakeyaml', version: '1.18'
+    compile group: 'org.yaml', name: 'snakeyaml', version: '1.24'
 
     compile libraries.springSecurityLdap
     compile libraries.springLdapCore
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/impl/config/AliasSupportingTypeDescription.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/impl/config/AliasSupportingTypeDescription.java
@@ -1,0 +1,29 @@
+package org.cloudfoundry.identity.uaa.impl.config;
+
+import org.yaml.snakeyaml.TypeDescription;
+import org.yaml.snakeyaml.introspector.Property;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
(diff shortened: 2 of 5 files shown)


taitzariel:
Hi,
GE has already signed the corporate CLA so please can you add me to the GE CLA?
Ariel

andrewedstrom:
@taitzariel Are there any reproduction steps we can use to verify the speedup?

taitzariel:
Hi Andrew. Indeed. Run the UAA locally. Then run a spring integration test, like org.cloudfoundry.identity.uaa.integration.feature.ImplicitGrantIT (from IDE).
Before the change, need to wait 20 seconds before the test begins.
With the change, test starts immediately.
I debugged it and found that the cause was the yaml interpreter spending all that time on the long uaa.yml

taitzariel:
@andrewedstrom Hi Andrew. Have you reviewed this pull request?

taitzariel:
closing to reopen after joining Predix org

andrewedstrom:
Hey @taitzariel! This definitely speeds things up, great find! Even so, I've talked it over with multiple members of the team and we're not ready to merge yet. 

The main concern is that uaa.yml file you've removed comments from serves as important documentation of what properties we accept. So we don't want to remove comments from it just for the sake of our tests.

Duplicating it to remove comments as you've done here is one option, but that creates the risk of the two drifting apart. This one file is already quite long, so maintaining two of them seems like it could come back to bite us very quick. In a project that already is huge and hard to maintain, we're especially hesitant to take on any changes that will make maintenance harder.

I wonder if there's a solution where we could efficiently generate a version of the file without comments when the integration tests start, so we get the speedup without duplication. That would be fine and we would probably merge it as long as the new code was easy enough to understand, but it would be _even better_ if we could figure out _why_ comments add so much time to the test startup time.

taitzariel:
@andrewedstrom Thanks for the review. I'll think about what can be done..

taitzariel:
@andrewedstrom I found that updating the snakeyaml is all that is needed to eliminate the 20 second startup time. After updating, I need to fix a few tests..

taitzariel:
@andrewedstrom Hi Andrew, I've upgraded the snakeyaml and now all tests pass. Please can you review this again?

andrewedstrom:
@taitzariel Awesome! I'll have the team review it. Made a story for you to track the prioritization: https://www.pivotaltracker.com/story/show/167157015

taitzariel:
@andrewedstrom Great. Thanks.

joshuatcasey:
@taitzariel I'm reviewing this now!

Issue #1047, opened 2019-07-09
Supporting Microsoft OIDC


AP-Hunt:
# The issue
### What version of UAA are you running?
4.30.0
### How are you deploying the UAA?

I am deploying the UAA
- using cf-deployment

### What did you do?
We configured an OIDC identity provider for Microsoft using [this](https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration) discovery document, in order to allow logins from any Microsoft product (i.e. not just a specific Azure AD) 

### What did you expect to see? What goal are you trying to achieve with the UAA?
We're in the process of enabling Microsoft single sign-on for tenants, whose accounts may be in any directory. We expected that, given the global discovery document, the login flow would just work.

### What did you see instead?
We found that Microsoft issues tokens whose `iss` claim is different from the issuer given in the discovery document. This caused UAA to believe the token was invalid.

After some investigation, we found that the `iss` claim includes the GUID of the organisation to which the user belongs in Microsoft land (as is evident from the discovery doc: `https://login.microsoftonline.com/{tenantid}/v2.0`). For some users (like Outlook or Hotmail users), the organisation GUID is well known. However, for corporate users, it will be their Azure/Office 365 tenant id.

For us, it is impossible to set an issuer URL that UAA would accept for all users.

# The suggestion 

@46bit and I have put together a small change in our [`paas-uaa` fork](https://github.com/alphagov/paas-uaa/commit/f8466b65008c8b551a2335ba8be6a6b8498f12f3) which introduces a new configuration flag for OIDC providers: `issuerValidationMode`. The flag has three possible values: `strict` (the default), `domain_only`, and `none`. 

The aim our of change is to provide a less strict form of issuer claim validation, in order to support scenarios like that posed by Microsoft SSO. In `domain_only` mode, the only the host portion of the token’s `iss` claim must match that of the configured URL. This is done as a string comparison, so subdomains are not considered to match a parent domain. E.g. `sub.domain.org` is not a match for `domain.org`, or vice versa
 
We’ve deployed this change in to our CloudFoundry deployment, and verified that we can do a successful Microsoft SSO login.

# Why an issue and not a PR 
We’re raising this as an issue and not a PR because we’re not sure if this necessarily the right approach to take, and we’d like some feedback on the problem and our proposed solution.

aeijdenberg:
[...]

PR #1048, opened 2019-07-11
[#166028126] Implement multiple modes of issuer claim validation

AP-Hunt:
What
---
Some IDPs (e.g. Microsoft) create tokens whose `iss` claim can vary from user
to user. Under the current version, UAA was unable to integrate with these
providers because it requires a single, specific issuer value to be present.

To enable UAA to integrate with providers who do this, we implement different
modes for validating the `iss` claim, under the `issuerValidationMode`
configuration property for OIDC providers

The modes are

STRICT
  The default behaviour. The string in the `iss` claim and the configured
  issuer URL must match exactly.

DOMAIN_ONLY
  The value of the `iss` claim and the configured issuer URL must be URLs. They
  are considered to match if their domains match. Subdomains are not considered
  to match a parent domain.

NONE
  The `iss` claims is always considered to be valid.

How to review
---
1. Code review
2. Run the tests via `./gradlew test`

We found that the shell scripts `./run-{unit,integration}-tests.sh` weren't reliable, even in a freshly checked out copy of `cloudfoundry/uaa`. We assume this is because we're missing some environment config. Running the tests via gradle was more reliable.

Who can review
---
Someone with some Java, and ideally UAA, knowledge

### DIFF
--- a/model/src/main/java/org/cloudfoundry/identity/uaa/provider/AbstractXOAuthIdentityProviderDefinition.java
+++ b/model/src/main/java/org/cloudfoundry/identity/uaa/provider/AbstractXOAuthIdentityProviderDefinition.java
@@ -35,6 +35,8 @@ public abstract class AbstractXOAuthIdentityProviderDefinition<T extends Abstrac
     private String relyingPartySecret;
     private List<String> scopes;
     private String issuer;
+
+    private XOAuthIssuerValidationMode issuerValidationMode;
     private String responseType = "code";
 
     public URL getAuthUrl() {
@@ -146,6 +148,13 @@ public abstract class AbstractXOAuthIdentityProviderDefinition<T extends Abstrac
         return (T) this;
     }
 
+    public XOAuthIssuerValidationMode getIssuerValidationMode() { return issuerValidationMode; }
+
+    public T setIssuerValidationMode(XOAuthIssuerValidationMode issuerValidationMode) {
+        this.issuerValidationMode = issuerValidationMode;
+        return (T) this;
+    }
+
     public String getResponseType() {
         return responseType;
     }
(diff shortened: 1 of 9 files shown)


AP-Hunt:
Ack! Targeted the wrong repository with this PR. Sorry @cf-identity !

aeijdenberg:
@AP-Hunt - I realise this has been closed, but I'm catching up on UAA Slack, and I recently spent a lot of time working through issues integrating UAA with O365, so this caught my attention, as the last comment above hints this might be re-opened elsewhere.

How sure are you that this PR is necessary?  What we found is that depending on which OpenID Connect set of configuration we used, we were in fact able to get the issuer string etc matching correctly.

The exact OpenID well-known configuration file that we needed to use was the one documented here: https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-protocols-oidc#fetch-the-openid-connect-metadata-document

IIRC there were at least 3 other URLs that would give a very similar configuration, however in those cases we had issues with the issuer not matching the confg in the doc, and eventually found the correct form (by RTFMing, instead of copy-pasting from out-of-date StackOverflow posts). :)

If it's not actually necessary to add additional config, then it would best not to. UAA seems to already contain so many variants of configuration, and it's not always clear that the security implications of the cross-product of options has been fully evaluated.

AP-Hunt:
Hi @aeijdenberg 

We've opened https://github.com/cloudfoundry/uaa/issues/1047 to discuss this further. The PR here was accidental, and was meant to be targeted at our fork of UAA.

So I understand a bit more about your scenario, can you tell me if your O365 integration is for a single tenant, or multiple?

aeijdenberg:
@AP-Hunt - we were authenticating against a single tenant. I'll add a comment in #1047.

PR #1072, opened 2019-08-07
Make UaaLoginHint Serializable

mikeroda:
UaaLoginHint is an attribute of UaaAuthenticationDetails, which itself is Serializable and part of an UaaAuthentication. This is stored in the SecurityContext and ultimately the HTTP session. These should all be serializable to avoid issues with session repositories that require serialization (e.g. for replication).

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/UaaLoginHint.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/UaaLoginHint.java
@@ -23,7 +23,7 @@ public class UaaLoginHint implements Serializable {
         }
     }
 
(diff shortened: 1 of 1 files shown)


strehle:
done with https://github.com/cloudfoundry/uaa/commit/984b19edd40539078d9748f492503d823842e7d6

mikeroda:
Glad to see this get in but I'm wondering why my original pull request wasn't merged from a year ago. I've been holding back on making any other pull requests because nothing seems to get in.

strehle:
sorry for that - cannot answer this. Hope this issue here is solved for you now if not, please rebase a new PR.

PR #1074, opened 2019-08-08
Package UAA WAR using the Spring Boot plugin

joshuatcasey:
Signed-off-by: Andrew Wittrock <awittrock@pivotal.io>


uaa/src/main/java/org/cloudfoundry/identity/uaa/UaaBootApplication.java
```diff
@@ -0,0 +1,15 @@
...
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
+
+@SpringBootApplication
+@EnableWebSecurity
+public class UaaBootApplication {
+    public static void main(String... args) {
+        SpringApplication.run(UaaBootApplication.class, args);
+    }
+
```
andrewedstrom, line 13: Rejecting on account of evil whitespace

andrewedstrom (changes requested):
It looks like this is failing to compile the tests. We should fix that before merging. Otherwise looks fine to me https://wings.pivotal.io/builds/8635595

andrewedstrom (approved):

PR #1107, opened 2019-09-27
Create a consolidated docker image for CI

aramprice:
- add `scripts/docker_run`
  - this passes `${DB}` for switching db variant, default is `hsqldb`
- add `scripts/gradle` to unify test dependency startup
  - all test scripts call into this script
- deletions
  - `scripts/ldap/install-ldap.sh`
  - => replaced by ldap setup in Dockerfile
  - `run-integration-tests.sh`
  - => use `scripts/docker_run scripts/integration-tests.sh`
  - `run-unit-tests.sh`
  - => use `scripts/docker_run scripts/unit-tests.sh`

[#168655249]

### DIFF
--- a/ci/image/Dockerfile
+++ b/ci/image/Dockerfile
@@ -1,0 +1,173 @@
+FROM ubuntu:18.04
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    curl jq git lsb-release unzip vim sudo \
+    apt-transport-https apt-utils ca-certificates gnupg \
+    && apt-get autoremove -yqq && apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+    # unzip used once
+
+ENV DEBIAN_FRONTEND=noninteractive
+ENV APT_KEY_DONT_WARN_ON_DANGEROUS_USAGE=true
+
+# JAVA
+ENV JAVA_HOME /usr/lib/jvm/java-bellsoft-amd64
+ENV PATH "${JAVA_HOME}/bin:${PATH}"
+RUN JAVA_MAJOR_VERSION="11" \
+    && JDK_METADATA="jdk-metadata.json" \
+    && curl -sLo "${JDK_METADATA}" "https://api.bell-sw.com/v1/liberica/releases?version-modifier=latest&os=linux&release-type=lts&bitness=64&package-type=tar.gz&bundle-type=jdk&arch=x86" \
+    && JDK_VERSION="$(jq -r ".[] | select(.featureVersion | contains("${JAVA_MAJOR_VERSION}")).version" "${JDK_METADATA}")" \
+    && JDK_ARCHIVE_URL="$(jq -r ".[] | select(.featureVersion | contains("${JAVA_MAJOR_VERSION}")).downloadUrl" "${JDK_METADATA}")" \
+    && JDK_ARCHIVE="$(basename "${JDK_ARCHIVE_URL}")" \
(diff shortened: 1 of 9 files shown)


aramprice:
This PR doesn't address any possible changes/improvements/consolidations that could be made in CI tasks/configuration.

ci/image/Dockerfile
```diff
@@ -0,0 +1,170 @@
...
+    && ldapadd -Y EXTERNAL -H ldapi:/// -f /ldap/ldap_db_init.ldif \
+    && ldapadd -x -D 'cn=admin,dc=test,dc=com' -w password -f /ldap/copy_of-uaa_src_main_resources_ldap_init.ldif \
+    && rm -rf /ldap/
+#/PROJECT_SETUP
+
+# CLEANUP
+RUN apt-get autoremove -yqq \
+   && apt-get clean \
+   && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+#/CLEANUP
```
sjolicoeur: Not sure why clean up would be necessary this late in the DockerFile, the data being removed will have been cached in earlier layers...

sjolicoeur:
I wonder if the following files should just be copied in instead of echoed in:
- `/bin/start-mysql`
- `/bin/start-postgresql`
- `/etc/ssl/ca.info`
- `/etc/ssl/ldap01.info`
- `/etc/ssl/certinfo.ldif`
- `/bin/start-slapd`

ci/image/Dockerfile
aramprice: upon reflection this makes sense. I think I lifted this pattern from elsewhere. given the per-layer caching the image size reduction benefit is only helpful if this cleanup happens in every layer which issues an `apt-get` command.
:-/

sjolicoeur: yeah I know :/

Also never tried [this but apparently bazel can build images without all the layers](https://github.com/bazelbuild/rules_docker) but not going there for this.

sjolicoeur:
Also I tried to build the image but I seem to be missing two files:
- `ldap_db_init.ldif` 
- `copy_of-uaa_src_main_resources_ldap_init.ldif`

should we move those files into the `uaa/ci/image` folder? 

PS: I did find `ldap_db_init.ldif`  in 4 different places
- `uaa/out/production/resources/ldap_db_init.ldif`
- `uaa/build/resources/main/ldap_db_init.ldif`
- `uaa/build/libs/WEB-INF/classes/ldap_db_init.ldif`
- `uaa/src/main/resources/ldap_db_init.ldif`

aramprice:
@sjolicoeur `uaa/ci/image/ldap_db_init.ldif` was removed in https://github.com/cloudfoundry/uaa/pull/1107/commits/a0b2ee71f75b568ceadeb06391c97503df9465df (but the reference is still in the `Dockerfile`)

sjolicoeur:
> @sjolicoeur `uaa/ci/image/ldap_db_init.ldif` was removed in [a0b2ee7](https://github.com/cloudfoundry/uaa/commit/a0b2ee71f75b568ceadeb06391c97503df9465df) (but the reference is still in the `Dockerfile`)

[yeah, is it](https://github.com/cloudfoundry/uaa/pull/1107/files#diff-ecea286bd876ec207731385e7882f5e7R159) still needed though?

ci/image/Dockerfile
aramprice: Adding `&& apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*` to every layer that uses `apt` shaved off about 200MB from the final image size.

PR #1129, opened 2019-10-24
Add the ability to build container images with Jib

mattysweeps:
https://github.com/GoogleContainerTools/jib is a gradle/maven plugin for building OCI images. This PR makes it easy to deploy UAA as a container.

There are a few caveats I ran in to:
1. I had to re-add the Gradle plugins maven repository, as Jib does not exist in Spring's plugin-release repository. Let me know if this is a blocker and I'll open up a ticket with Jib to get Jib in Spring's repository.

2. There is a bug in Jib's logic for task dependencies. https://github.com/GoogleContainerTools/jib/issues/2096 Jib does not check if the `bootWar` or `war` tasks are enabled/disabled. Since currently UAA has `bootWar` disabled, Gradle does not invoke it in time. For now, I had to manually run `./gradlew war` and then `./gradlew jib`. **UPDATE** The bug has been fixed. https://github.com/GoogleContainerTools/jib/pull/2107 I will update this PR once a new version of Jib is released

3. I had to add the following spring configuration at `./uss/src/main/webapp/WEB-INF/tomcat-web.xml`:
```
<?xml version="1.0" encoding="ISO-8859-1"?>
<web-app xmlns="http://java.sun.com/xml/ns/javaee" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/web-app_3_0.xsd" version="3.0"
  metadata-complete="true">
  <servlet>
        <servlet-name>default</servlet-name>
        <servlet-class>
          org.apache.catalina.servlets.DefaultServlet
        </servlet-class>
        <init-param>
            <param-name>debug</param-name>
            <param-value>0</param-value>
        </init-param>
        <init-param>
            <param-name>listings</param-name>
            <param-value>false</param-value>
        </init-param>
        <load-on-startup>1</load-on-startup>
    </servlet>
</web-app>
```

I'm not an expert at Spring XML configuration, but AFAIK this is required because UAA doesn't have it's own servlet configuration by default. I didn't include this in the PR since this was a requirement for UAA, not necessarily Jib itself.

Also, Tomcat was chosen for the base image for the container because of Tomcat's use with various plugins and frameworks. (When I used the default Jetty image, I ran into runtime bean initialization errors)


andrewedstrom:
@mattysweeps Can you tell me more about your use case? Why are you hoping to build a UAA image?

mattysweeps:
@andrewedstrom My use case is to deploy UAA to Kubernetes.

joshuatcasey:
Hi @mattysweeps thanks for the PR. We're hopefully pretty close to having a Spring Boot UAA - would that change anything for this PR?

Also, I'd love to hear more information about how you're using the UAA, specifically related to containers produced by Jib.

PR #1134, opened 2019-11-05
Use offline spring-security.xsd

joshuatcasey:
- Need to use https protocol
- https://www.springframework.org/schema/security/spring-security.xsd

[#169574211]

Also see https://github.com/spring-projects/spring-security/issues/7560


andrewedstrom:
This fixes https://github.com/cloudfoundry/uaa-release/issues/140

andrewedstrom:
Closed because this was addressed by #1135

PR #1140, opened 2019-11-13
allow wildcard in port

strehle:
- during URI normalize the port wildcard info is lost and then laster in match it fails
- check therefore only for clientRedirect if port is wildcard and store the info
- in match adjust URis with port wildcard

We have a scenario, where client has defined redirect_uri

https://example.com:*/**

and Request is for 

https://example.com:65000/path/subpath

This was working in uaa 4.30.0 but not now


strehle:
@torsten-sap  FYI

shamus:
Finally looking at this! From what I see, this PR does two things:

1. Restores the ability to wildcard the port via the ant path matcher syntax
1. Falls back to the default port 80 if no port is specified.

That second item, defaulting the port if not specified, is new functionality. I do not believe we want this feature.

If you would like to verify this, check out `3a488590e1b469a5def9c0d4680c832e5c63562a` (or any sha before `7ffadabf7a9084bd613b59f048b355c27723037a`) and add the following test to `AntPathRedirectResolverTests`

```
    @Nested
    @DisplayName("matching http://example.com")
    class WhenCurious {
        private final String clientRedirectUri = "http://example.com:*/whatever";
        private final String anotherClientRedirectUri = "http://example.com:*";

        @Test
        void whatHappens() {
            assertTrue(resolver.redirectMatches("http://example.com:8080/whatever", clientRedirectUri));
            assertFalse(resolver.redirectMatches("http://example.com/whatever", clientRedirectUri));
            assertTrue(resolver.redirectMatches("http://example.com:8080", anotherClientRedirectUri));
            assertFalse(resolver.redirectMatches("http://example.com", anotherClientRedirectUri));
        }
    }
```

strehle:
Hi, added your test and adapted it.
because the default port behaviour is something, why dont you want it ? - dont know if I understood your concern correctly....
but therefore added tests to the behaviour I see it as correct, e.g.
https://github.com/cloudfoundry/uaa/pull/1140/files#diff-c4272c474416658aecf1fb161807e061

default port is defined (and allowed in https://tools.ietf.org/html/rfc3986) see 3.2.3.  Port, Page 22
So to me the default port for http(s) is something I would like to have.

shamus:
Is all of the complication around detecting if the wildcard contains a port wildcard necessary? I believe that's there to handle this case:

```
assertTrue(resolver.redirectMatches("http://example.com:80", "http://example.com"));
```

I could be mistaken, but isn't that case handled when this code falls back to spring?

strehle:
> Is all of the complication around detecting if the wildcard contains a port wildcard necessary? I believe that's there to handle this case:
> 
> ```
> assertTrue(resolver.redirectMatches("http://example.com:80", "http://example.com"));
> ```
> 
> I could be mistaken, but isn't that case handled when this code falls back to spring?

yes this case . is done by spring security DefaultRedirectResolver...

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/beans/LegacyRedirectResolverTest.java
```diff
@@ -602,6 +602,32 @@ void subdomainMatchingRejectsDomainRedirectOnAntPathVariableSubdomain() {
...
+            final String clientRedirectUri = "https://example.com:*/";
+            assertTrue(resolver.redirectMatches("https://example.com:80/", clientRedirectUri));
+            assertFalse(resolver.redirectMatches("https://example.com/", clientRedirectUri));
+        }
+
+        @Test
+        void matchesEmptyPortWithWildcardPortAndPath() {
+            final String clientRedirectUri = "https://example.com:*/**";
+            assertTrue(resolver.redirectMatches("https://example.com:80/path1/path2/path3", clientRedirectUri));
+            assertFalse(resolver.redirectMatches("https://example.com/path1/path2/path3", clientRedirectUri));
```
strehle, line 629: here you see, removed the match if default port.

strehle (commented):
ok now ?

shamus:
:+1:

shamus (approved):

PR #1146, opened 2019-11-21
Add DisableSpringUaaSpringSecurityFilterRegistrationBean

rwinch:
Spring Security filters are created as beans in the XML files and then
aggregated into FilterChainProxy which is registered with Spring Boot.
However, Spring Boot registers all Beans that implement Filter with the
servlet container. This means that all the Security Fiters are executed
twice. Additionally, the Filter is going to be executed for every URL
since Boot's default is to register for every URL.

This is a post processor that marks all UAA Filter's as disabled (i.e. do
not register it with the servlet container). This way it is only executed
within FilterChainProxy


joshuatcasey:
@rwinch I'm going to close this PR. We're using the code diffs, but our `wip/bootrun` branch has been rebased off of other branches and it's in a completely different state of git history so we won't merge it.

joshuatcasey:
But thanks so much for the suggestion - it's unblocked us for now on getting to Boot!

PR #1149, opened 2019-11-27
[Don't Merge Yet] Chores/cleanup yaml servlet profile initializer

joshuatcasey:
Simplify `YamlServletProfileInitializer` by hardcoding servlet initialization parameters.

This does bring in behavioral changes specific to direct OSS UAA users (not BOSH release users [see 1]).

`YamlServletProfileInitializer` has several levels of indirection to support init params in `web.xml`. However, these params are hardcoded and not configurable, meaning that the indirection can be simplified by inlining. 

## Former behavior:
1) Load from `uaa.yml,login.yml`
1) Then try to add from 
```
  - ${LOGIN_CONFIG_URL}
  - file:${LOGIN_CONFIG_PATH}/login.yml
  - file:${CLOUDFOUNDRY_CONFIG_PATH}/login.yml
  - ${UAA_CONFIG_URL}
  - file:${UAA_CONFIG_FILE}
  - file:${UAA_CONFIG_PATH}/uaa.yml
  - file:${CLOUDFOUNDRY_CONFIG_PATH}/uaa.yml
```
1) If (1) & (2) didn't yield any results, load from 
```
  - ${APPLICATION_CONFIG_URL}
  - file:${APPLICATION_CONFIG_FILE}
```
1) If still no results, try (2) and (3) again.
1) Note that there would always be results from (1) because `uaa.yml` is packaged with the WAR file.

## New behavior:
1) Load from `uaa.yml,login.yml`
2) Load from all of:
```
  - ${APPLICATION_CONFIG_URL}
  - file:${APPLICATION_CONFIG_FILE}
  - ${LOGIN_CONFIG_URL}
  - file:${LOGIN_CONFIG_PATH}/login.yml
  - file:${CLOUDFOUNDRY_CONFIG_PATH}/login.yml
  - ${UAA_CONFIG_URL}
  - file:${UAA_CONFIG_FILE}
  - file:${UAA_CONFIG_PATH}/uaa.yml
  - file:${CLOUDFOUNDRY_CONFIG_PATH}/uaa.yml
```
3) Note that properties from later files loaded take precedence over properties from earlier files.

### Notes
[1] The UAA BOSH release only seems to care about `file:${CLOUDFOUNDRY_CONFIG_PATH}/uaa.yml`, as configured at https://github.com/cloudfoundry/uaa-release/blob/ed3111886628e41caf0acc3b388a8f297a2a4aa7/jobs/uaa/templates/config/bpm.yml.erb#L7

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/impl/config/YamlServletProfileInitializer.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/impl/config/YamlServletProfileInitializer.java
@@ -59,7 +59,7 @@ public class YamlServletProfileInitializer implements ApplicationContextInitiali
 
     private static final String DEFAULT_YAML_KEY = "environmentYamlKey";
 
-    private String yamlEnvironmentVariableName = "UAA_CONFIG_YAML";
+    private static final String YAML_ENVIRONMENT_VARIABLE_NAME = "UAA_CONFIG_YAML";
 
     private SystemEnvironmentAccessor environmentAccessor = new SystemEnvironmentAccessor() {
     };
@@ -125,7 +125,7 @@ public class YamlServletProfileInitializer implements ApplicationContextInitiali
 
     private Resource getYamlFromEnvironmentVariable() {
         if (environmentAccessor != null) {
-            String data = environmentAccessor.getEnvironmentVariable(getYamlEnvironmentVariableName());
+            String data = environmentAccessor.getEnvironmentVariable(YAML_ENVIRONMENT_VARIABLE_NAME);
             if (hasText(data)) {
                 //validate the Yaml? We don't do that for the others
                 return new InMemoryResource(data);
@@ -201,14 +201,6 @@ public class YamlServletProfileInitializer implements ApplicationContextInitiali
                 environment.setActiveProfiles(commaDelimitedListToStringArray(systemProfiles));
             }
         }
-    }
-
-    String getYamlEnvironmentVariableName() {
-        return yamlEnvironmentVariableName;
-    }
-
-    void setYamlEnvironmentVariableName(String yamlEnvironmentVariableName) {
-        this.yamlEnvironmentVariableName = yamlEnvironmentVariableName;
     }
 
     void setEnvironmentAccessor(SystemEnvironmentAccessor environmentAccessor) {
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/impl/config/YamlServletProfileInitializer.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/impl/config/YamlServletProfileInitializer.java
@@ -23,13 +23,9 @@ import javax.servlet.ServletContext;
 import java.io.FileNotFoundException;
 import java.net.URISyntaxException;
 import java.net.URL;
-import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


shamus:
Hi @jcasey

If I understand correctly, this PR does two things:
1. moves where the location of the UAA config files is specified. Before, these were specified in the `web.xml`, now they're essentially constants in `YamlServletProfileInitializer`

1. alters the load order of the UAA configuration files, moving the `APPLICATION_CONFIG_URL` & `APPLICATION_CONFIG_FILE` options earlier in the process.

I understand that changing how the location of the UAA config files are specified is motivated by a potential migration to Spring Boot. This makes sense. Spring Boot does not make use of `web.xml` and the way Spring Boot assembles the `ApplicationContext` is pretty different from a Spring MVC app. 

Additionally, moving the configuration from `web.xml` into `YamlServletProfileInitializer` is basically a noop because `web.xml` couldn't be influenced by an end user.

I don't understand why this PR is changing the load order of the UAA config files. While I agree that the load order is complicated and could be simplified, I don't believe changing this behavior is necessary to achieve any Spring Boot outcome. 

Could you help me better understand this change?

joshuatcasey:
@shamus the original logic for reading in the different file locations had a lot of checks that would never be true. `locations == null` or `resources.isEmpty()` might cause the UAA to load from one of the following. 
```
  - ${APPLICATION_CONFIG_URL}
  - file:${APPLICATION_CONFIG_FILE}
```

It would likely be more accurate to indicate that the UAA would never read from those files and drop support for them, but that was a more drastic change. So instead they are loaded at the lowest priority.

joshuatcasey:
After searching online for documentation references:
- uaa.yml
  - [see 1]
- login.yml -> found none
- LOGIN_CONFIG_URL -> found none
- ${APPLICATION_CONFIG_URL} -> found none
- file:${APPLICATION_CONFIG_FILE} -> found none
- ${LOGIN_CONFIG_URL} -> found none
- file:${LOGIN_CONFIG_PATH}/login.yml -> found none
- file:${CLOUDFOUNDRY_CONFIG_PATH}/login.yml -> found none
- ${UAA_CONFIG_URL}
  - [see 1]
- file:${UAA_CONFIG_FILE}
  - [see 1]
- file:${UAA_CONFIG_PATH}/uaa.yml -> found none
- file:${CLOUDFOUNDRY_CONFIG_PATH}/uaa.yml
  - [see 1]

### Notes
[1] https://docs.cloudfoundry.org/concepts/architecture/uaa.html
```
  classpath:uaa.yml
  file:${CLOUD_FOUNDRY_CONFIG_PATH}/uaa.yml
  file:${UAA_CONFIG_FILE}
  ${UAA_CONFIG_URL}
```

shamus:
So, it would be fair to say there's no technical reason driving a change to how the UAA config files are loaded?

joshuatcasey:
@shamus I would say the opposite - there's only a technical reason. With the property file locations inlined, this collapsed a lot of the logic in YamlServletProfileInitializer. Instead of having dead code that loaded these two locations (which would "feel like" deprecating the files) I made the choice to bring the files in at the lowest priority. Any chance we could review this code change live with @andrewedstrom ?

```
- ${APPLICATION_CONFIG_URL}
- file:${APPLICATION_CONFIG_FILE}
```

Birdrock:
Closing this in favor of a more surgical change to YamlServletProfileInitializer in https://github.com/cloudfoundry/uaa/pull/1155

PR #1151, opened 2019-11-29
Purge expired memory sessions

joshuatcasey:
[#170010973]

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/web/beans/UaaMemorySessionConfig.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/web/beans/UaaMemorySessionConfig.java
@@ -4,17 +4,26 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
-import org.springframework.context.annotation.*;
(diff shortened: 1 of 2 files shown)


mwdb:
I was looking at #1152 and #1151 

What would be important to have regular cleanup of sessions. Is the default for #1152 configured for once per hour? This would be to long for my understanding.

joshuatcasey:
@mwdb Both #1151 and #1152 allow for the configuration to drive the purge timeline. We'll take a look during Pacific working hours to figure out the best parts of #1151 and #1152 and make sure that the BOSH release allows setting any configuration options. CC: @shamus @dbeneke

shamus:
@mwdb while it is configurable, why do you feel the default of an hour would be too long? Is the session store consuming most of the heap within an hour?

It's mostly a matter of how aggressively to purge the store.

mwdb:
We have seen examples where we had 80.000 sessions taking 500 MB. Freeing this only once an hour would mean releasing a bunch of memory at once and causing a lot of work for the GC. It is better to more frequently release memory to avoid any peaks.

shamus:
I went ahead and bumped to default to 15 minutes in the other PR.  I understand that you have an active installation and will need the ability to tune how frequently sessions are purged which we can address via a property in the BOSH release.

shamus:
Closing as #1152 has been merged.

PR #1155, opened 2019-12-04
Prepare YamlServletProfileInitializer for Spring Boot

joshuatcasey:
YamlServletProfileInitializer relies on hard-coded configuration in web.xml. This migrates that configuration into YamlServletProfileInitializer directly since it's not configurable.

Also, YamlServletProfileInitializer relies on the servlet config. This will not be available in the Spring Boot lifecycle, so provide workarounds when possible.

- Hardcode init params
- Use STDOUT/STDERR instead of `servletContext.log()`
- Allow context path to default to "/" (this may be refactored for Spring Boot)

[#169991138]

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/impl/config/YamlServletProfileInitializer.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/impl/config/YamlServletProfileInitializer.java
@@ -19,7 +19,6 @@ import org.springframework.web.context.ConfigurableWebApplicationContext;
 import org.springframework.web.context.support.WebApplicationContextUtils;
 import org.yaml.snakeyaml.Yaml;
 
-import javax.servlet.ServletConfig;
 import javax.servlet.ServletContext;
 import java.io.FileNotFoundException;
 import java.net.URISyntaxException;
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/cloudfoundry/identity/uaa/impl/config/YamlServletProfileInitializer.java
```diff
@@ -68,19 +63,21 @@
     public void initialize(ConfigurableWebApplicationContext applicationContext) {
 
         ServletContext servletContext = applicationContext.getServletContext();
+        final String contextPath = servletContext != null ? servletContext.getContextPath() : "/";
 
-        HttpSessionEventPublisher publisher = new HttpSessionEventPublisher();
-        servletContext.addListener(publisher);
+        if (servletContext != null) {
+            HttpSessionEventPublisher publisher = new HttpSessionEventPublisher();
+            servletContext.addListener(publisher);
```
andrewedstrom, line 70: @joshuatcasey Am I correct in assuming that _pre-boot_ this is never null, and _post-boot_ it is always null? If so, I have a couple of questions/thoughts.

1) Once we switch to boot, how are we going to deal with the listener not getting added? It probably needs to be added elsewhere, right?
2) Since this condition is always satisfied pre-boot and never satisfied after boot, maybe we shouldn't put this in an `if` block.

andrewedstrom: LGTM otherwise

joshuatcasey: I believe that statement is correct with regards to when `servletContext` is/not `null`, so in Spring Boot that `if` statement will be dead code. 

In Spring Boot we can add a servlet listener like so:

https://www.concretepage.com/spring-boot/spring-boot-listener#ServletListenerRegistrationBean

andrewedstrom: Also this might be relevant https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/web/servlet/ServletContextInitializer.html

PR #1157, opened 2019-12-05
correction saml2 bearer documentation

strehle:
SAML2 bearer flow is often used from our side and therefore many people check the documentation on CF side , so that should be consistent with behaviour.
First version used always token endpoint but later it was refactored with behaviour of saml validation, therefore different endpoint


dbeneke:
@strehle Thank for bringing this to our attention.  Added this to the team's backlog for review and merge.

dbeneke:
The documentation change went out as a part of 74.16.0.  Thank you!

PR #1159, opened 2019-12-09
create index in transaction , which repairs test

strehle:


server/src/main/resources/org/cloudfoundry/identity/uaa/db/postgresql/V4_99_1575367461__revocable_token_index.sql
```diff
@@ -1 +1 @@
-CREATE INDEX CONCURRENTLY IF NOT EXISTS revocable_tokens_user_id_client_id_response_type_identity__idx on revocable_tokens(user_id, client_id, response_type, identity_zone_id);
+CREATE INDEX IF NOT EXISTS revocable_tokens_user_id_client_id_response_type_identity_idx ON revocable_tokens(user_id, client_id, response_type, identity_zone_id);
```
shamus: This isn't safe from a "no downtime" perspective since the table will be locked. I believe this is why the migration was written this way to begin with, correct?

Is there no way to address the test without changing the implementation?

strehle: then we would have to change/adapt the failing test in Mfa context.-

strehle:
@phschon  please check if you can repair the mfa test or if concurrent feature is not a must

@shamus  Such an index on big tables is always critical during deployment because of increasing CPU load. So we created it already on larger landscapes . For us the "IF exits" is therefore important .
New landscapes should get the index of course but then the lock is no problem because the DB there is almost empty

shamus:
Hi y'all, any updates here?

strehle:
to me the concurrent is not needed, therefore I would merge this one

andrewedstrom:
I'm pretty sure `concurrently` is needed if we want to minimize downtime. Why is the test failing? It seems like something must be wrong with our tests if adding `concurrently` to an index creation is making them fail

andrewedstrom:
@strehle We found (and fixed) a bug in our test runner that was causing the failure. https://github.com/cloudfoundry/uaa/commit/480760e909aab6de77e4bb41da5401aa34e87565

As such, can we close out this PR?

strehle:
@joshuatcasey  thanks

PR #1162, opened 2019-12-10
Minimal server.xml for UAA image

shamus:
This server.xml should represent the minimal configuration necessary for the UAA to successfully run. Tomcat behavior cannot be configured yet and logging will go to the console for now.

Configuration of the UAA, however, is possible by providing the location of the UAA config file via one of the supported environment variables.

### DIFF
--- a/k8s/image/tomcat/conf/server.xml
+++ b/k8s/image/tomcat/conf/server.xml
@@ -1,0 +1,17 @@
+<?xml version='1.0' encoding='utf-8'?>
+<Server port='-1'>
+  <Service name='Catalina'>
+    <Connector port='8080' bindOnInit='true' connectionTimeout='20000' maxHttpHeaderSize='14336'/>
+
+    <Engine defaultHost='localhost' name='Catalina'>
+      <Valve className='org.apache.catalina.valves.RemoteIpValve' protocolHeader='x-forwarded-proto'/>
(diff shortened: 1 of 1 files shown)


andrewedstrom (approved):
LGTM. By design it's very similar to https://github.com/cloudfoundry/tomcat-cnb/blob/master/server.xml, and the ways it diverges seem reasonable

Follow these steps to actually use it in the pack CLI: https://www.pivotaltracker.com/story/show/169713142/comments/209546135

strehle:
@shamus  can you please check PR 
https://github.com/cloudfoundry/uaa-release/pull/108
and add also

        <Valve className="org.apache.catalina.valves.ErrorReportValve"
               showReport="false"
               showServerInfo="false"/>

so that failures do not show the stacktrace

andrewedstrom:
@strehle we have a separate story for that valve. Linking so you can follow along: https://www.pivotaltracker.com/story/show/170193262

shamus:
@strehle Something that isn't immediately clear from @andrewedstrom's message is we're taking this opportunity to do a bit of housecleaning while doing this k8s work.  So we're looking to start with minimal & build up as we introduce features into the k8s release.

We appreciate the feedback and please keep it coming! We don't want to accidentally cut anything important.

PR #1165, opened 2019-12-12
Hardcode reference to env var UAA_CONFIG_YAML

joshuatcasey:
- There's no way to change or set it

[#170136573]


andrewedstrom (approved):
Looks fine, reasonable refactor

PR #1166, opened 2019-12-13
Remove Dead Code from YamlServletProfileInitializer

joshuatcasey:
- Remove Dead Code
- Not a behavioral change!
- ${APPLICATION_CONFIG_URL} and file:${APPLICATION_CONFIG_FILE} would never be checked because the ternary would shortcircuit to true
- resources.isEmpty() would always be false because classpath:uaa.yml always exists

[#170136573]


Birdrock:
Looks good once it passes checks. 👍

andrewedstrom (approved):

PR #1170, opened 2019-12-20
[DO NOT MERGE] Use and test ytt templates for k8s config

andrewedstrom:


joshuatcasey:
I'm curious if other tech stacks were considered. It sounds like what we want to test is that we have K8s YML files that we pump through `ytt` and come out looking a certain way. It kind of feels like we could have a semantic YML diff in bash that would just work (https://github.com/sahilm/yamldiff is an example).

shamus:
@joshuatcasey  What benefits would diffing have over parsing and making targeted expectations?

In my experience, one challenge with static fixtures is that  as use cases grow there's either a large number of fixtures or a single fixture that starts taking on many responsibilities. When there's a large number of fixtures things can get messy (hard to find the right fixture, duplication, etc). When a single fixture serves many tests, tests can become brittle.

shamus:
Some high level thoughts:

1. I don't like the go & the java co-mingling in the `src` directory, I can't quite say why. I think it's because when looking at `src` it's not immediately clear what role that code plays. I also don't have an alternate suggestion.

2. There's still a lot of boilerplate in the tests. I'd like to see some helpers for rendering the templates and some custom matchers to simplify interacting with the parsed YAML. I can help brainstorm some approaches post holiday.

andrewedstrom:
Closing in favor of https://github.com/cloudfoundry/uaa/pull/1177

PR #1177, opened 2019-12-30
[DO NOT MERGE] Explore template generation testing methods

shamus:
Two methods for testing yml generated by ytt:
- parse the generated yml into structs and make fine grained expectations
- use the new assert functionality in ytt

Posting this for discussion / following on from #1170


andrewedstrom:
@Birdrock and I took a look at this and here are our impressions:
1. We like having the tests in the `/k8s` directory and using gomodules a lot more than what we did in our PR https://github.com/cloudfoundry/uaa/pull/1170
2. We're generally opposed to the overlay testing approach. It adds additional mental overhead, it makes it harder to find where the actual assertions are, it means you have to write your test code in yaml comments (which can be fat-fingered), and it means you need a file for every test.
3. People unfamiliar with writing matchers may be intimidated when they first want to add to the custom matcher code. I think it's probably worth the tradeoff though, they're not actually that hard to understand once you read through them.
4. Let's do the matchers eyyy

k8s/test/deployment_test.go
```diff
@@ -0,0 +1,49 @@
...
+		deploymentPath = pathToTemplate("deployment.yml")
+		valuesPath = pathToTemplate(filepath.Join("values", "values.yml"))
+	})
+
+	When("Parsing YAML", func() {
+		It("Constructs the YAML from a set of files", func() {
+			ctx := NewRenderingContext(deploymentPath, valuesPath)
+
+			Expect(ctx).To(ProduceYAML(
+				RepresentingContainer("uaa").RunningImage("cfidentity/uaa@sha256:93b70b26fbb3de88d93728b0daf1ea7b001fde89a24e283c3db36bf4c6af087c"),
```
joshuatcasey, line 22: I'll admit I have trouble reading this off. It doesn't read as a "template test" to me. Also, I'm not sure if many other tests would make use of `RunningImage`.

shamus: You're right, `RunningImage` would be rarely used.  

This approach is trying to consolidate the logic for traversing the JSON documents within these matcher classes (as opposed to spreading it across the test suite). Is there another approach you would find more approachable?

PR #1181, opened 2020-01-13
Chores/matcher cleanup

shamus:
Prior to this PR, the matchers used a simplified interface to make assertions, but this proved difficult to extend.

This PR brings the k8s matchers closer to the underlying structure of the data resulting from parsing the resulting YAML. This should ease both the extension of the matchers and their expressiveness.

### DIFF
--- a/k8s/test/container_matcher_test.go
+++ b/k8s/test/container_matcher_test.go
@@ -1,0 +1,52 @@
+package k8s_test
+
+import (
+	"fmt"
+	"github.com/onsi/gomega"
+	"github.com/onsi/gomega/format"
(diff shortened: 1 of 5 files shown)


andrewedstrom:
I like this. I like that the structure of the matcher calls mimics the structure of the generated yaml. 

I think this is quite readable:
```
		Expect(ctx).To(
			ProduceYAML(
				RepresentingDeployment().WithPodMatching(func(pod *PodMatcher) {
					pod.WithContainerMatching(func(container *ContainerMatcher) {
						container.WithName("uaa")
					})
				}),
			),
		)
```

PR #1182, opened 2020-01-13
Features/deployment recommended labels 170668852

shamus:
This PR adds the recommended to the UAA deployment.

https://www.pivotaltracker.com/story/show/170668852

### DIFF
--- a/k8s/test/object_meta_matcher_test.go
+++ b/k8s/test/object_meta_matcher_test.go
@@ -1,0 +1,62 @@
+package k8s_test
+
+import (
+	"fmt"
+	"github.com/onsi/gomega"
+	"github.com/onsi/gomega/format"
+	"github.com/onsi/gomega/gstruct"
+	"github.com/onsi/gomega/types"
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type ObjectMetaMatcher struct {
+	fields map[string]types.GomegaMatcher
+
+	meta     *metaV1.ObjectMeta
+	executed types.GomegaMatcher
+}
+
+func NewObjectMetaMatcher() *ObjectMetaMatcher {
+	return &ObjectMetaMatcher{map[string]types.GomegaMatcher{}, nil, nil}
+}
+
+func (matcher *ObjectMetaMatcher) WithName(name string) *ObjectMetaMatcher {
+	matcher.fields["Name"] = gomega.Equal(name)
+
+	return matcher
+}
+
+func (matcher *ObjectMetaMatcher) WithLabels(labels map[string]string) *ObjectMetaMatcher {
+	var matchers []types.GomegaMatcher
+	for label, value := range labels {
+		matchers = append(matchers, gomega.HaveKeyWithValue(label, value))
+	}
+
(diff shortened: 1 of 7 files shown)


k8s/test/object_meta_matcher_test.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"fmt"
+	"github.com/onsi/gomega"
+	"github.com/onsi/gomega/format"
+	"github.com/onsi/gomega/gstruct"
+	"github.com/onsi/gomega/types"
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type ObjectMetaMatcher struct {
```
andrewedstrom, line 12: Can we say `ObjectMetaDataMatcher` instead of `ObjectMetaMatcher` just to be a little more clear about what part of the generated YAML this is testing?

andrewedstrom:
The tests look clean. I'll note that it takes a lot of lines of changes to add the tests for these labels getting set to the right values, but the resulting test code looks quite clean.

k8s/test/object_meta_matcher_test.go
andrewedstrom: Turns out K8s calls it ObjectMeta so nevermind: https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta

k8s/templates/values/values.yml
```diff
@@ -0,0 +1,7 @@
+#@data/values
+---
+version: "74.12.0"
```
andrewedstrom: It's worth noting we will need to manage making sure this version label is kept updated in CI

k8s/templates/values/values.yml
```diff
@@ -0,0 +1,7 @@
...
+labels:
+  instance: standalone
+  partOf: uaa
```
andrewedstrom: I'm assuming that it was an intentional choice not to have this be `Cloud Foundry`, even though that's what the [Idiomatic Components Doc](https://docs.google.com/document/d/1aVQkmCGmHMuCJS9h1UfO-THcP7ENdqIZe5lYtOzcPPA/edit?ts=5dc1b4ad#) suggests, because UAA is used by more products than just Cloud Foundry.

Birdrock:
Looks good. There's definitely some cost upfront to create the matchers, but I think it will be worth it.

k8s/templates/values/values.yml
shamus: Yes, we certainly will.

k8s/templates/values/values.yml
shamus: It was.  Since the UAA is deployed to multiple environments I chose to represent the standalone UAA in the base config.  I assumed CF, TKG, et al would provide value overrides for what's important to them.

shamus:
> The tests look clean. I'll note that it takes a lot of lines of changes to add the tests for these labels getting set to the right values, but the resulting test code looks quite clean.

For this first set of tests, it is a lot of lines.  I think that this code in particular will be reused as the pod & service & config map all should have these labels.

Furthermore this could be tightened up vi ytt functions or a more specific matcher depending on how these labels are applied!

PR #1192, opened 2020-01-22
Redirect prefixes of passcode

46bit:


andrewedstrom:
@46bit Thanks for the PR! Can you tell us more about why you think this feature should be added to UAA?

46bit:
@andrewedstrom For now this is still a draft–I'll absolutely expand things once it's ready to be considered :)

PR #1194, opened 2020-01-23
Validate state param when talking to External IDPs

andrewedstrom:
As per the spec's requirements for OAuth clients https://tools.ietf.org/html/draft-ietf-oauth-v2-30#section-10.12

Includes a lot of refactors to make the the XOAuthProviderConfigurator
more readable and a complete rewrite of the XOAuthAuthenticationFilter's
tests, as they were growing hard to extend.

[#170636851]

Co-authored-by: Joshua Casey <jcasey@pivotal.io>
Co-authored-by: Andrew Wittrock <awittrock@pivotal.io>
Co-authored-by: Andrew Edstrom <aedstrom@pivotal.io>
Signed-off-by: Andrew Edstrom <aedstrom@pivotal.io>


server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/XOAuthAuthenticationFilter.java
```diff
@@ -89,6 +109,9 @@ private boolean authenticationWasSuccessful(
             ofNullable(successHandler).ifPresent(handler ->
                     handler.setSavedAccountOptionCookie(request, response, authentication)
             );
+        // TODO: :eyes_narrowed:
```
andrewedstrom, line 112: We definitely need to remove this TODO before merging

shamus: But why not fix it? Catching all exceptions scary.

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/XOAuthProviderConfigurator.java
```diff
@@ -48,31 +52,56 @@ protected OIDCIdentityProviderDefinition overlay(OIDCIdentityProviderDefinition
...
+            uriBuilder.queryParam("nonce", nonceGenerator.generate());
+        }
+
+        return uriBuilder.build().toUriString();
+    }
+
+    private void saveStateParamForIdpToRequestSession(String state, String idpOriginKey, HttpServletRequest request) {
+        if (request != null) {
+            request.getSession().setAttribute("xoauth-state-" + idpOriginKey, state);
+        }
```
shamus: Is there any value to this null check? Even if it is null, this operation doesn't make any sense if we can't access the session.

andrewedstrom: Good point

PR #1196, opened 2020-01-28
Add liveness, readiness, and startup probes to UAA pod

andrewedstrom:
This allows the UAA pod to correctly report whether it's running, based on the return value of the healthz endpoint. If it fails, k8s will restart it.

We decided to add a StartupProbe because the UAA can take minutes to start up due to database migrations, but once the start up is complete we would like for k8s to be able to more quickly detect pod failures. Right now we give the pod 5 minutes to start up. There's probably a conversation for us to have about how we do migrations generally, but until then we are a good candidate for a StartupProbe.

We didn't add tests since these values are not currently configurable.

Relevant k8s documentation on probes: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

[#170836774]

### DIFF
--- a/k8s/templates/deployment.yml
+++ b/k8s/templates/deployment.yml
@@ -30,7 +30,8 @@ spec:
         - name: uaa
           image: "cfidentity/uaa@sha256:93b70b26fbb3de88d93728b0daf1ea7b001fde89a24e283c3db36bf4c6af087c"
           ports:
-            - containerPort: 8080
+            - name: http-uaa
+              containerPort: 8080
               protocol: TCP
           env:
             - name: spring_profiles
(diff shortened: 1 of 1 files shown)


shamus:
LGTM!  Curious if the idiomatic component guidelines have anything to say about either the naming of ports or how liveness/readiness probes should be handled.

andrewedstrom:
@shamus good find! The [idiomatic components guidelines](https://docs.google.com/document/d/1aVQkmCGmHMuCJS9h1UfO-THcP7ENdqIZe5lYtOzcPPA/edit?ts=5dc1b4ad#heading=h.gloxaecyp9pu) says this:

> Ports that are exposed by pod must have a name which should be the same as in corresponding service. 

We updated accordingly

PR #1201, opened 2020-02-04
Allow configuration of UAA port from commandline

shamus:
Prior to this commit, `build.gradle` ran the UAA on port 8080 no ifs ands or buts.  Now, that can be overridden by specifying `-Pport=<port>`.


Birdrock:
👍

PR #1203, opened 2020-02-07
Cleanup xml beans part1

joshuatcasey:


andrewedstrom:
Nothing here seems controversial to me. It's basically reformatting, moving XML config to java config, and standard refactors like moving towards constructor injection.

I'd probably avoid using the `static` init block in `ExpiringUrlCacheTests` and replace it with a `@BeforeAll` or a oneliner where we declare the variable, but that's more of a matter of taste than anything. Shouldn't affect readability really.

PR #1204, opened 2020-02-10
[Don't Merge] Chores/aop annotations vs xml

joshuatcasey:
I wanted to see what it was like to express AOP concerns as annotations instead of XML. 

I don't like what I see.

Perhaps this could inform how we handle control flow in the UAA.


joshuatcasey:
Closing since it's been open too long

PR #1212, opened 2020-02-19
avoid using text templating where unnecessary within k8s templates

cppforlife:


Birdrock:
Can you share the pro/cons of using one templating mechanism or another?

Issue #1229, opened 2020-03-06
introspect endpoint don't follow rfc7662


IIDarkKnightII:
**SECURITY NOTICE: If you have found a security problem in the UAA, please do not file a public github issue. Instead, please send an email to security@cloudfoundry.org**

Thanks for taking the time to file an issue. You'll minimize back and forth and help us help you more effectively by answering all of the following questions as specifically and completely as you can.

### What version of UAA are you running?

http://docs.cloudfoundry.org/api/uaa/version/74.15.0

### What did you do?

Hit `/introspect` endpoint to fetch the scope.

### What did you expect to see? What goal are you trying to achieve with the UAA?

According to RFC, the SCOPE should be a JSON String containing a space-separated list of scopes associated with the token. And this is how all the frameworks like Spring Security etc. and most of the major programming languages has built by following RFC.

[RFC for Reference](https://tools.ietf.org/html/rfc7662#section-3.3)

`"scope" : "scim.userids openid cloud_controller.read password.write cloud_controller.write`

### What did you see instead?
I get the scope as an array like below: 
`"scope" : [ "scim.userids", "openid", "cloud_controller.read", "password.write", "cloud_controller.write" ]`

joshuatcasey:
Thanks for bringing this to our attention @IIDarkKnightII - our Product Managers are looking into this and we'll update with our next steps, hopefully very soon!

DanSalt:
Is there any update on this? We're currently blocked on this, as we're trying to integrate UAA with Kafka (which uses KeyCloak - which expects the scope to be a single string). I appreciate there's a backwards-compatibility issue here, but could it be added with a flag to switch between the two formats? cc @joshuatcasey

shamus:
Hi @DanSalt backwards compatibility is the kicker. This issue isn't currently a priority for us, but we'd be open to a PR. If you wanted to put together a proposal describing how the flag would work and how you'd implement this feature we'd be happy to discuss it.

PR #1230, opened 2020-03-06
Complying with RFC https://tools.ietf.org/html/rfc7662 for introspect…

IIDarkKnightII:
### Fix for Issue - Introspect endpoint don't follow rfc7662
 https://github.com/cloudfoundry/uaa/issues/1229

### Description

According to RFC, the SCOPE of `introspect token` should be a JSON String containing a space-
separated list of scopes associated with the token. And this is how all the frameworks like Spring Security etc. and most of the major programming languages has built by following RFC.

[RFC for Reference](https://tools.ietf.org/html/rfc7662#section-3.3)

`"scope" : "scim.userids openid cloud_controller.read password.write cloud_controller.write`

### What did you see instead?

I get the scope as an array like below: 
`"scope" : [ "scim.userids", "openid", "cloud_controller.read", "password.write", "cloud_controller.write" ]`

## How did I fixed this?

I created a property in subclass and serializing the SCOPE property in claims as a JSON String

### DIFF
--- a/model/src/main/java/org/cloudfoundry/identity/uaa/oauth/token/IntrospectionClaims.java
+++ b/model/src/main/java/org/cloudfoundry/identity/uaa/oauth/token/IntrospectionClaims.java
@@ -16,16 +16,31 @@ package org.cloudfoundry.identity.uaa.oauth.token;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 
+import java.util.List;
+
 public class IntrospectionClaims extends Claims {
 
-    @JsonProperty("active")
-    private boolean active;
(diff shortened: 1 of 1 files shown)


joshuatcasey:
Hi thanks @IIDarkKnightII for submitting this PR. We're not in a position to accept this PR right now for a few reasons:

- We need to look at how our downstream consumers are impacted by this change
- It's good that the tests passed, but that's likely an indicator that we need an automated test to verify spec compliance. 
- We'll have to discuss this as a team, and I'd like to make sure all that conversation happens over on #1229 

We'll use this PR for reference if/when we get to fixing this issue.

PR #1250, opened 2020-03-31
UAA's admin client secret is now a K8s secret object

joshuatcasey:
Implements the story (https://www.pivotaltracker.com/story/show/171398937).

### Admin client secret

The UAA admin client secret can now be provided to `ytt` render commands via `-v admin.client_secret=<some value>`. This will create an admin client with authority of at least `uaa.admin` and grant type `client_credentials`. This admin client will be the only client configured in the UAA. 

### Other clients and the default profile

No other clients or users will be configured in the UAA (the `default` Spring profile is not applied).

### Local developer experience

Adds `./k8s/run-minikube.sh` as the default developer experience for deploying the UAA to minikube locally for development purposes.

### DIFF
--- a/k8s/Makefile
+++ b/k8s/Makefile
@@ -7,8 +7,8 @@ GOFILES=`find . -type f -name '*.go'`
 .PHONY: test
 .PHONY: render
 .PHONY: apply
-.PHONY: minikube
 .PHONY: kubectl-clean
+.PHONY: kubectl-clean-all
 .PHONY: docker-run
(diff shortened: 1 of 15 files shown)


joshuatcasey:
I deleted story https://www.pivotaltracker.com/story/show/172078743 because we already have story https://www.pivotaltracker.com/story/show/171398937

joshuatcasey:
deleting PR in favor of #1252

PR #1266, opened 2020-04-11
Remove default profile

joshuatcasey:
Use config files to express needs for tests or running locally


joshuatcasey:
Closed in favor of #1267

PR #1267, opened 2020-04-13
[#171940790] Remove default profile

joshuatcasey:
- Use config files to express needs for tests

[#171940790]


joshuatcasey:
Closing since it's been open too long

PR #1271, opened 2020-04-16
[#172158373] Make Account Chooser page aware of the origin

tack-sap:
Display the origin for each account on the account chooser page
Pass the origin of the user as login_hint when choosing an account
Adopt tests accordingly

### DIFF
--- a/server/src/main/resources/templates/web/idp_discovery/account_chooser.html
+++ b/server/src/main/resources/templates/web/idp_discovery/account_chooser.html
@@ -20,13 +20,14 @@
       <li class="list-group-item" th:each="account, iter : ${savedAccounts}" th:inline="text">
         <form th:id="${account.userId}" action="/login/idp_discovery" th:action="@{/login/idp_discovery}" method="post" role="form">
           <input type="hidden" class="form-control" name="email" th:value="${account.origin == 'uaa' || account.origin == 'ldap' ? account.username : account.email}" />
+          <input type="hidden" class="form-control" name="login_hint" th:value="${'%7B%22origin%22:%22'+account.origin+'%22%7D'}" />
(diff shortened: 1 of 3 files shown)


server/src/main/resources/templates/web/idp_discovery/account_chooser.html
```diff
@@ -20,13 +20,14 @@ <h4 class="txt-c pbxxl ptxl" th:text="${T(org.springframework.util.StringUtils).
       <li class="list-group-item" th:each="account, iter : ${savedAccounts}" th:inline="text">
         <form th:id="${account.userId}" action="/login/idp_discovery" th:action="@{/login/idp_discovery}" method="post" role="form">
           <input type="hidden" class="form-control" name="email" th:value="${account.origin == 'uaa' || account.origin == 'ldap' ? account.username : account.email}" />
+          <input type="hidden" class="form-control" name="login_hint" th:value="${'%7B%22origin%22:%22'+account.origin+'%22%7D'}" />
```
shamus, line 23: I'm surprised this json fragment needed to be urlencoded by hand.  Won't thymeleaf handle the encoding?

tack-sap: It probably would do that, but putting the double quotes there without enconding or escaping would break the statement. 
So I chose to url encode it manually, as both thymeleaf  can work with that and the UAA knows how to handle the url encoded value.

joshuatcasey:
Testing some markdown

Addresses Account chooser ignores origin and instead uses default [[#172158373](https://www.pivotaltracker.com/story/show/172158373), https://github.com/cloudfoundry/uaa/pull/1271]

PR #1283, opened 2020-05-06
Fix user details update during password grant flow

phschon:
The password grant flow does not update user information if they have changed on IDP side.
This PR fixes this issue by removing a not needed check from `ExternalOAuthAuthenticationManager`: `request.getPrincipal()` is checked for `null` but is not used afterwards in the method. During a password grant flow, `request.getPrincipal()` is `null` and thus user information are not updated if they have changed.

Scenario to reproduce:
- Register an IDP in the UAA
- Do a password grant flow with a user from this IDP
- Check the user information stored in the UAA database
- Change user information in the IDP (e.g. family name or given name)
- Do a password grant flow again
- The new information are not updated in the UAA database


shamus:
LGTM!

Issue #659, opened 2017-07-18
Excessive logging with userids_enabled: false


suhlig:
We are running CF with UAA's `userids_enabled: false`:

```yaml
properties:
  uaa:
    scim:
      userids_enabled: false
```

With a log level of WARN, we find that the uaa.log is getting filled pretty fast with the following messages:

```
[2017-07-18 03:15:03.103] uaa - 22465 [http-nio-8443-exec-62] ....  WARN --- UserIdConversionEndpoints: Request from user cloud_controller_username_lookup received at disabled Id translation endpoint with filter:id …
[2017-07-18 03:15:03.104] uaa - 22465 [http-nio-8443-exec-62] .... ERROR --- ScimUserEndpoints: Unhandled exception in SCIM user endpoints.
org.cloudfoundry.identity.uaa.scim.exception.ScimException: Illegal operation.
        at org.cloudfoundry.identity.uaa.scim.endpoints.UserIdConversionEndpoints.findUsers(UserIdConversionEndpoints.java:101)
...
```

It appears that this comes from [UserIdConversionEndpoints.java](https://github.com/cloudfoundry/uaa/blob/master/server/src/main/java/org/cloudfoundry/identity/uaa/scim/endpoints/UserIdConversionEndpoints.java#L98-L101).

As `userids_enabled` seems to be a valid flag, we wonder if it is possible to cut down the noise a bit; especially the exception stack.

Thanks!

Steffen and Bernhard

sreetummidi:
@suhlig Having this flag enabled is requirement for UAA which is used by CC. What is the reason for disabling the flag?

suhlig:
@sreetummidi The flag is fine, either way. It's just that when we set it to false, the log file is filled with tons of exceptions including their stack trace, which seems surprising with a log level of `WARN`.

We were hoping that the logging can be reduced to a one-line warning only.

fhanik:
I agree that this is an issue. Expected errors should not be logged under ERROR or WARN

phschon:
Hi,

I addresses this issue in the following PR: https://github.com/cloudfoundry/uaa/pull/1292
We also ran into the same problem and would like to solve this issue.

Best regards

shamus:
Thanks for the PR. Merged!

PR #1292, opened 2020-05-19
Reduce noise in logs of excpected error messages

phschon:
This PR addresses this issue: https://github.com/cloudfoundry/uaa/issues/659

By changing the return type of `UserIdConversionEndpoints.findUsers()` to `ResponseEntity` it is possible to return an expected response code of `400` as well as a message without throwing an exception. This also removes the exception from the logs.
Additionally, the log level should be `info`, since this is an expected behaviour.


shamus:
LGTM!

PR #1305, opened 2020-05-25
Refactor createCompositeToken 

### LINKED ISSUE #936: Performance of /oauth/token endpoint
Hi UAA team :) 

version: 4.20.0

locally only using gradlew  (osx, 2,6 GHz Intel Core i7)

###  Preface
Because I wanted to check what is UAA's limit of token generation, I ran small performance tests, simple wrk call on /oauth/token - nothing fancy. Result was not splendid so I checked for what UAA spend its CPU time, I attached async-profiler and this is the output 

screen shot 2019-01-02 at 20 35 24

UAA, is spending more than 50% of time in **org.cloudfoundry.identity.uaa.zone.MultitenantJdbcClientDetailsService.loadClientByClientId**

I made small experiment and I added simple "cache" to that class (just ConcurrentHashMap) that holds result of 
```java
ClientDetails details;
try {
    details = jdbcTemplate.queryForObject(selectClientDetailsSql, new ClientDetailsRowMapper(), clientId, zoneId);
} catch (EmptyResultDataAccessException e) {
    throw new NoSuchClientException("No client with requested id: " + clientId);
}
```
and with that piece of code included, UAA behaves a bit better on that path
screen shot 2019-01-02 at 20 42 30
only key signing (here 2048bits) and some fasterxml stuff.

here, I uploaded original SVG files (github don't allow for uploading svg files in issue)
before: https://github.com/KowalczykBartek/tmp-svg/blob/master/slow.svg
after: https://github.com/KowalczykBartek/tmp-svg/blob/master/slow-but-better.svg

###  Question
Is it possible to add some caching on MultitenantJdbcClientDetailsService layer ? Does it makes sense from security point of view (or just any point of view :D ) to cache ClientDetails in memory (it can be cached on each instance without any sharing, like Redis or Memcached) ?

thanks !

strehle:
combind usage of User and ClientDetails for the id and access token creation

Improvement for issue https://github.com/cloudfoundry/uaa/issues/936


strehle:
@shamus can you check if ok from your side. with this refactoring there are 2 selects less because of the re-use of user and client details

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java
```diff
@@ -444,10 +441,10 @@ private CompositeToken createCompositeToken(String tokenId,
         compositeToken.setValue(token);
         BaseClientDetails clientDetails = (BaseClientDetails) clientDetailsService.loadClientByClientId(clientId);
 
-        if (idTokenGranter.shouldSendIdToken(userId, clientDetails, requestedScopes, grantType)) {
+        if (idTokenGranter.shouldSendIdToken(user == null ? null : user.getId(), clientDetails, requestedScopes, grantType)) {
```
shamus: I don't think this null check is needed.

`createCompositeToken` has two call sites `UaaTokenServices#refreshAccessToken` and `UaaTokenServices#createAccessToken` both of which perform a lookup by id for the user and pass the result to this method.

That means that the user object passed into this method cannot be null, correct?

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java
```diff
@@ -503,22 +499,22 @@ private KeyInfo getActiveKeyInfo() {
         if (null != grantType) {
             claims.put(GRANT_TYPE, grantType);
         }
-        if (user!=null && userId!=null) {
-            claims.put(USER_ID, userId);
+        if (user!=null) {
```
shamus, line 502: I don't think this user reference can be null by the same logic above (I realize these null checks were already here).

Am I missing something?

strehle: User can be null if client credential

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java
strehle: refactored code

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java
shamus: I see, I missed the condition in `createAccessToken` checking for the client credentials grant... makes sense.

But I definitely smell a refactor opportunity here. So much of this code isn't applicable to the client credentials grant (find a user, generate an id token, generate a refresh token), it's easy to make mistakes (or perform extra database lookups).

strehle: to me this coding is fine now because the client credential flow is currently the only without a user in it, but there are other flows , e.g. saml-bearer for clients or the x509 (which I want PR) is also without a user but should grant an access token, so this check is clear to me.

The extra lookup in DB is something which we see a problem

PR #1307, opened 2020-05-25
Documentation improvement: revocable token

### LINKED ISSUE #464: Clarify default API revocable/opaque behavior when optional token_format parameter is omited
currently, the api at http://docs.cloudfoundry.org/api/uaa/#password-grant states

> token_format  String  Optional    UAA 3.3.0 Can be set to 'opaque’ to retrieve an opaque and revocable token.

 It does not describe the expected behavior of the token endpoints (for the various grants) when the `token_format` field is omitted, or values that can be set to request non-opaque format (`JWT` ? ).

It would be useful to document whether the token is then systematically non revocable, or whether this depends on UAA default configuration such as [uaa.jwt.revocable](http://bosh.io/jobs/uaa?source=github.com/cloudfoundry/uaa-release&version=19#p=uaa.jwt.revocable)

strehle:
issue https://github.com/cloudfoundry/uaa/issues/464


shamus:
Thanks!

PR #1309, opened 2020-05-26
Feature request: no authorities claim

strehle:
omit the authorities  claim in client credential if empty parameter is requests this


shamus:
I think this change makes sense. My biggest concern is that this changes the contents of the JWT... making this a breaking change if some app were relying on the token in it's original state.

I'm generally hesitant to introduce a breaking change. Could this behavior be made opt in? 

@tylerschultz I wonder if you have any insights into whether developers rely on the authorities claim.

strehle:
the authorities is something which is not in standard, so I would like to omit it at all, but this change is not breaking any defaults . I also had the fear to break any existing scenarios .

Therefore if a client wants create a JWT with a limited amount of scopes then this can be done with "sope=<claim>" parameter and now with an empty "authorities" parameter, see
https://github.com/cloudfoundry/uaa/pull/1309/files#diff-dd300954bc1371593f35fdf828def3c3R239

The filtering of authorities is a feature I dont see as wanted, but simply omit authorities because a oauth client rely on scope claim

tylerschultz:
I don't have a good sense for whether the authorities claim is commonly used. Seems vestigial, and seemingly any spec compliant library would not action upon the value.

I curious to know what the motivation is for removing the claim? Tidiness? Is its presence causing problems?

strehle:
The PR is because we have customers with concerns that the private data in client_credential token is stored even if a token was limited with a scope parameter. 

So the scenario is a client credential flow with extra parameter scope e.g. scim.read. As result you receive a JWT only with scope scim.read, but with all authorities. 

We see 2 concerns. 
a) size. if a client has a hugh number of authorities a JWT will always be hugh even with the limit. 
b) if a requestor wants to hide all potential authorities of a JWT that cannot be done with only scope parameter, therefore this PR

shamus:
@strehle It might have been clear what I meant by "breaking change".  I meant that because the claim has been present in the JWT, it has been available for apps to consume. Taking it away could break any application that was processing that claim.

Generally, the motivations for the change make sense. What I want to know is if the change can be introduced in a way that won't break applications which might have been relying on the claim being present.

strehle:
@shamus  with that approach the client can decide to omit the authorities, similar with the scope parameter to limit the scopes in the parameter.
Another solution could be to add it in oauth_client_details as attribute, but then it is not that flexible.
Default of any change should be same behaviour as now, but my change also does not change default behaviour.

If you have another way to add this feature, I will be happy too. The result should be that there is a possiblility to omit the authorities claim because the security argument is somehow valid. All possible authorities are exposed even if an owner of a client does not want that.

PR #1325, opened 2020-06-16
Automount service account tokens

XanderStrike:
In an environment which requires first party jwt tokens, the istio sidecar requires the service account token be mounted to authenticate with the apiserver on behalf of the application.

We are required to use first party tokens because Kind does not support third party tokens, and Kind is a target deployment for cf-for-k8s (and is used in many team's pipelines).

[Related issue](https://github.com/istio/istio/issues/22193)

[Related story](https://www.pivotaltracker.com/story/show/173346213)

### DIFF
--- a/k8s/templates/service_account.yml
+++ b/k8s/templates/service_account.yml
@@ -4,4 +4,3 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: uaa
-automountServiceAccountToken: false


shamus:
Hi @XanderStrike 

Thanks for the PR!  

The use case you mention makes sense, however we'd like to see this setting be configurable in our K8s release. I imagine a top level boolean property named `mountServiceAccountToken" would do the trick.

Is that something you're open to doing?

kauana:
Sure, we can definitely do it :balloon: 

We just submitted a [new PR](https://github.com/cloudfoundry/uaa/pull/1326).

cc my pair: @ndhanushkodi

PR #1326, opened 2020-06-17
Make automountServiceAccountToken configurable

kauana:
Follow up PR to this [older PR](https://github.com/cloudfoundry/uaa/pull/1325).

Now `automountServiceAccountToken` is configurable and defaults to false. We are working on the cf-for-k8s changes + ytt overlay that will set this field to true when using a Kind cluster.

Let us know if you have any questions.

cc my pair: @ndhanushkodi

### DIFF
--- a/k8s/templates/service_account.yml
+++ b/k8s/templates/service_account.yml
@@ -4,4 +4,4 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: uaa
-automountServiceAccountToken: false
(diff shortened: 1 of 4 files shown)


shamus:
This looks good!  I noticed that the test suite wasn't extended to include this feature, could you add some tests please?

ndhanushkodi:
@shamus we've added the test, thanks for pointing that out!

PR #1330, opened 2020-06-22
Update externalId on login

tack-sap:
With this PR the field externalId of the user is added to the attributes that are updated on user login.

Background:
When creating a shadow user manually the externalId field is typically not set. When a such a user now does a login, the field will not be updated and the field will stay empty. However if the shadow user is created during the first login of a user, the field will always be set. So the two types of users look different, which should not be the case.
With the proposed change the field will be updated when a user does a login and afterwards the field will always be set - regardless of how the shadow user was initially created.

This also updates the externalId if the value changes, e.g. if the IdP is changing the format (say from using an id to using the email)


shamus:
Hi @tack-sap, thanks for the PR!

Could you help me understand the use case driving this change? 
- What is motivating the manual creation of shadow users?
- What is the impact of the external id not being updated?

tack-sap:
> * What is motivating the manual creation of shadow users?
The motivation to create them before their first login is, that some authorizations shall be assigned to the user already before he logs on to the UAA for the first time. So if I want to onboard a new user I can do it all upfront instead of waiting for him to login to the UAA before I can onboard him. (this is not necessarily a manual step, but the shadow users could also be created automatically by an external instance)

> * What is the impact of the external id not being updated?
As mentioned above: If this value is never updated, it will be empty for some users and filled for other users - so you cannot rely on it. So any usage of this attribute will be affected, in our special case we have customers that want to use this value to map the user of the IdP accross different environments.

torsten-sap:
@shamus Any update on this PR? It looks quite straightforward to me. Any objections?

PR #1359, opened 2020-07-17
Remove `writeNewPom` task and POM files

aramprice:
Dependabot supports gradle directly, no need for pom.xml


aramprice:
Merge is blocked until https://github.com/cloudfoundry/identity-ci/pull/8 is accepted

PR #1388, opened 2020-08-27
Fix check for secure flag in cookie

strehle:
found by accident


shamus:
Good catch! It seems like some of the unit tests are failing.  Could you address those?

strehle:
started again in #1398

PR #1404, opened 2020-09-12
Fix gradle deprecation warnings, hard-mode

aramprice:
This PR will replace the `compile` and related configurations in gradle build scripts.
* [Dependencies should no longer be declared using the compile and runtime configurations](https://docs.gradle.org/6.3/userguide/upgrading_version_5.html#dependencies_should_no_longer_be_declared_using_the_compile_and_runtime_configurations)

This is a more complex since the shift from `compile` => `implementation` hides module dependencies from consuming modules. This means that consuming code which previously had transitive access to the dependencies of modules it consumes no longer has that access. The consuming modules will now need to explicitly include dependencies which had previously been transitively available.

* [The Java Library plugin configurations](https://docs.gradle.org/6.3/userguide/java_library_plugin.html#sec:java_library_configurations_graph)

### DIFF
--- a/metrics-data/build.gradle
+++ b/metrics-data/build.gradle
@@ -1,8 +1,8 @@
 description = "CloudFoundry Identity Metrics Data Jar"
 
 dependencies {
-  compile(libraries.jacksonDatabind)
-  compile(libraries.jacksonAnnotations)
+  implementation(libraries.jacksonDatabind)
+  implementation(libraries.jacksonAnnotations)
 }
 
 processResources {
--- a/model/build.gradle
+++ b/model/build.gradle
@@ -1,25 +1,25 @@
 description = "CloudFoundry Identity Model JAR"
 
 dependencies {
-    compile(project(":cloudfoundry-identity-metrics-data"))
+    implementation(project(":cloudfoundry-identity-metrics-data"))
 
-    compile(libraries.jacksonDatabind)
-    compile(libraries.jacksonAnnotations)
+    implementation(libraries.jacksonDatabind)
+    implementation(libraries.jacksonAnnotations)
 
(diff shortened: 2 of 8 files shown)


aramprice:
One deprecation warning remains which is [Legacy publication system is deprecated and replaced with the *-publish plugins](https://docs.gradle.org/6.3/userguide/upgrading_version_5.html#legacy_publication_system_is_deprecated_and_replaced_with_the_publish_plugins). This is related to the use of the `maven` plugin which appears to be pulled in as part of `apply(plugin: "propdeps-maven")`.

I believe this is used to generate the `pom.xml` files via the `writeNewPom` task. Alternatives to this plugin will need to be used before upgrading to Gradle 7.0

The task [generatePomFileFor{PubName}Publication](https://docs.gradle.org/6.3/userguide/publishing_maven.html#publishing_maven:tasks) which uses [GenerateMavenPom](https://docs.gradle.org/6.3/dsl/org.gradle.api.publish.maven.tasks.GenerateMavenPom.html) looks promising.

```
./gradlew help --warning-mode all

> Configure project :
The maven plugin has been deprecated. This is scheduled to be removed in Gradle 7.0. \
Please use the maven-publish plugin instead. Consult the upgrading guide for further information: \
https://docs.gradle.org/6.3/userguide/upgrading_version_5.html#legacy_publication_system_is_deprecated_and_replaced_with_the_publish_plugins
        at build_1sav6t7w25mbw9xe231myvzao$_run_closure1.doCall(~/workspace/uaa/build.gradle:34)
        (Run with --stacktrace to get the full stack trace of this deprecation warning.)
```

shamus:
LGTM.  Thanks for helping with this!

PR #1397, opened 2020-09-08
Enabled database session store

Birdrock:
- Session store in database allows horizontal scaling.

Signed-off-by: Eric Promislow <epromislow@suse.com>

PR #1405, opened 2020-09-14
Join Java Opts with strings library

Birdrock:
- Deployment test used string concatenation.
- Caused it to be brittle.

### DIFF
--- a/k8s/test/deployment_test.go
+++ b/k8s/test/deployment_test.go
@@ -1,13 +1,15 @@
 package k8s_test
 
 import (
+	"path/filepath"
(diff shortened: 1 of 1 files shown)


shamus:
Thanks for the test clean up :)

But this PR is now in conflict with `develop` since #1397 was merged. If you can get that sorted out, we'll merge this

Birdrock:
@shamus I absolutely knew it would, but felt it wouldn't be proper to roll the two together. I'll fix it up and push an update.

Birdrock:
Pushed an update.

PR #1411, opened 2020-09-21
exclude prometheus endpoint from Istio proxy

JamesClonk:
The prometheus /metrics endpoints needs to be excluded from being handled by the Istio proxy in cf-for-k8s, otherwise the port will not be accessible from a Prometheus deployment that resides outside the Istio service mesh.

Signed-off-by: Fabio Berchtold <fabio.berchtold@swisscom.com>

### DIFF
--- a/k8s/templates/deployment.yml
+++ b/k8s/templates/deployment.yml
@@ -24,6 +24,7 @@ spec:
         prometheus.io/scrape: 'true'
         prometheus.io/port: '9102'
         prometheus.io/path: '/metrics'
+        traffic.sidecar.istio.io/excludeInboundPorts: '9102'
       labels:
         app: uaa-deployment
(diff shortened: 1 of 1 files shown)


JamesClonk:
Never mind, I just saw the following comments, which seem to indicate this is wanted behaviour:
- https://github.com/cloudfoundry/capi-k8s-release/commit/697d1a6921480c066a29473253de673fc394ad19
- https://github.com/cloudfoundry/cf-k8s-prometheus/commit/bf9d5dfa4a526c3b7b7fa098795f91df30c96f75

I'll open a PR towards cf-for-k8s instead with an optional configurable flag for an overlay that adds `traffic.sidecar.istio.io/excludeInboundPorts` to all necessary deployments in case of users having their own custom Prometheus installation.

PR #1413, opened 2020-10-09
minify application.css

ianchov:
application.css is 244717 bytes when minified it is 130186 bytes or a reduction 87.97% This file is loaded every time a user tries to authenticate on a landscape, however usually it is not used as after load the user is rediected to SAML/OIDC authentication. This minification will save 88% of the traffic, metered in hundred of GB per month for most productive landscapes


ianchov:
Closing this one as i opened a new PR againts the develop branch https://github.com/cloudfoundry/uaa/pull/1414

PR #1422, opened 2020-10-19
revert pr #1418

strehle:
https://github.com/cloudfoundry/uaa/pull/1418


strehle:
see /postgresql-42.2.16.jar

PR #1427, opened 2020-10-21
Switch statsd_exporter image to use the one from cloudfoundry dockerhub

heycait:
[#174735977]

We've moved statsd_exporter image to live under the cloudfoundry dockerhub instead.

slack: #logging-and-metrics


thausler786 (approved):
lgtm

Issue #1430, opened 2020-10-27
UAA using insufficient default configurations for RestTemplates


tack-sap:
### What version of UAA are you running?

Latest CF release, but issue is still present in the develop branch

### How are you deploying the UAA?

We saw the issue both as part of a CF release and using the UAA locally as standalone.

### What did you do?

When using the RestTemplates configured in [RestTemplateConfig](https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/impl/config/RestTemplateConfig.java) (that are used e.g. by the password grant against an OIDC provider) we experienced connection problems for this connection. We found, that this was due to the limited size of the connection pool that is used by the underlying HttpClient. Unfortunately this property is not configurable, but even the documentation of the HttpClient states, that the default values may be to restrictive ([see here, Section 2.3.3](https://hc.apache.org/httpcomponents-client-ga/tutorial/html/connmgmt.html))

### What did you expect to see? What goal are you trying to achieve with the UAA?

Similar to the timeout, that is already configurable using the uaa.yml, we would like to introduce new configuration options for the HttpClient. To solve our problem we would need to be able to configure the poolsize of the connection pool as well as the maximum connections per route. 

We propose to contribute a PR that makes those two values configurable, similar to the timeout value that is already configurable.

shamus:
This makes sense to me.

A PR for this behavior sounds good, thank you for offering to put that together.  I assume the new config will be expressed inside of [RestTemplateConfg](https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/impl/config/RestTemplateConfig.java) like:

```
    @Value("${rest.template.timeout:10000}")
    public int timeout;

    @Value("${rest.template.maxTotal:20}")
    public int maxTotal;

    @Value("${rest.template.maxPerRoute:2}")
    public int maxPerRoute;
```

Or did you have something else in mind?

tack-sap:
Yes, that's what we had in mind. So all the parameters for the rest templates are configured in the same way and read in the same class.

tack-sap:
I opened a PR with our proposal of how to implement it. We are open for feedback on that proposal.

strehle:
@shamus because of pipeline issues, I debugged this code and even if 
[Section 2.3.3](https://hc.apache.org/httpcomponents-client-ga/tutorial/html/connmgmt.html) explain 2 / 20 as default I see 
[...]

PR #1434, opened 2020-10-30
Add configuration options for HttpClient

tack-sap:
PR for #1430 

Added the two configuration parameters to the RestTemplateConfig, as discussed there.

In addition I also added a configuration to enable connection reuse (currently also not possible to configure). As alternative a ConnectionKeepAliveStrategy is added (based on our own experience with connection reuse). This can be configured independently from the other configurations.

I also added an example configuration (commented out) to the uaa.yml so it is clear how these values should be configured there. 

Note that I implemented all new properties in a way that not configuring these values (or setting the values to 0) will fall back to the default configuration of the HttpClient - so existing configurations will not be affected by this PR (neither positive nor negative).

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/util/UaaHttpRequestUtils.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/util/UaaHttpRequestUtils.java
@@ -51,7 +51,7 @@ public abstract class UaaHttpRequestUtils {
     private static Logger logger = LoggerFactory.getLogger(UaaHttpRequestUtils.class);
 
     public static ClientHttpRequestFactory createRequestFactory(boolean skipSslValidation, int timeout) {
-        return createRequestFactory(getClientBuilder(skipSslValidation), timeout);
+        return createRequestFactory(getClientBuilder(skipSslValidation, 20, 2, 0), timeout);
     }
 
     public static ClientHttpRequestFactory createRequestFactory(boolean skipSslValidation, int timeout, int poolSize, int defaultMaxPerRoute, int maxKeepAlive) {
@@ -67,10 +67,6 @@ public abstract class UaaHttpRequestUtils {
         return httpComponentsClientHttpRequestFactory;
     }
 
-    protected static HttpClientBuilder getClientBuilder(boolean skipSslValidation) {
-        return getClientBuilder(skipSslValidation, 0,0, 0);
-    }
-
     protected static HttpClientBuilder getClientBuilder(boolean skipSslValidation, int poolSize, int defaultMaxPerRoute, int maxKeepAlive) {
         HttpClientBuilder builder = HttpClients.custom()
             .useSystemProperties()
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/cloudfoundry/identity/uaa/util/UaaHttpRequestUtils.java
```diff
@@ -55,14 +68,33 @@ protected static ClientHttpRequestFactory createRequestFactory(HttpClientBuilder
...
+    protected static HttpClientBuilder getClientBuilder(boolean skipSslValidation, int poolSize, int defaultMaxPerRoute, int maxKeepAlive) {
         HttpClientBuilder builder = HttpClients.custom()
             .useSystemProperties()
             .setRedirectStrategy(new DefaultRedirectStrategy());
         if (skipSslValidation) {
             builder.setSslcontext(getNonValidatingSslContext());
             builder.setSSLHostnameVerifier(new NoopHostnameVerifier());
         }
-        builder.setConnectionReuseStrategy(NoConnectionReuseStrategy.INSTANCE);
+        if (poolSize > 0) {
```
shamus: I think it would be simpler to provide reasonable defaults and then always build and configure the `PoolingHttpClientConnectionManager`, reducing all of these branches.

shamus (changes requested):
Apart from the comments above:
1. This is a bit more complicated than what was discussed in #1430, could you tell me about the need for a custom keep alive strategy?
2. Tests please.

server/src/main/java/org/cloudfoundry/identity/uaa/util/UaaHttpRequestUtils.java
tack-sap: I implemented it this way to guarantee, that not configuring the values would not change anything. But if you prefer it with such a simpler structure, I will update the code like this.

tack-sap:
@shamus regarding your questions:

1. When reusing the connection the HttpClient assumes that the connection can be kept open indefinitely (if nothing else is provided by the response via Header). But it turns out, depending on your infrastructure, the connection may also be terminated by intermediate components so you end up trying to reuse a broken connection and get an error. That's why for our internal developments we already use such a KeepAliveStrategy when reusing connections of RestTemplates. [Section 2.6](https://hc.apache.org/httpcomponents-client-ga/tutorial/html/connmgmt.html) of the HttpClient documentation also give some information about when to use it. 

Adding it here was just a proposal, but if you are unhappy with it we can also remove that and stay with the NoConnectionReuseStrategy only.

2. I will add some tests to UaaHttpRequestUtilsTest that the resulting RestTemplates are still working properly, as the changes are mostly configuration and there is no real functional difference with this change. Is that what you had in mind?

tack-sap:
Simplified the structure of the code (using the original default values), and adopted the test to use an HttpClient with changed configuration.

shamus (approved):

Issue #1439, opened 2020-11-04
Performance issue when using several IdPs via the account chooser


torsten-sap:
### What version of UAA are you running?

74.26.0

### How are you deploying the UAA?

I am deploying the UAA using cf-deployment

### What did you do?

* Configuration of several active OIDC IdPs (let us say 10 as an example)
* Account chooser is enabled
* IdP discovery is enabled

### What did you expect to see? What goal are you trying to achieve with the UAA?

**Issue 1**
If the account chooser page is accessed, I expect that no IdPs need to be queried from the database as this information is not required at all.

**Issue 2**
If I select a user account for a specific IdP in the account chooser, I expect that the UAA redirects me to this IdP by querying only the required information for that particular IdP from the DB. Note that this is a special case where the login_hint is passed. 

### What did you see instead?

**Issue 1**
If the account chooser page is accessed, the method login in class LoginInfoEndpoint queries all active IdPs from the database.

**Issue 2**
If I select a user account for a specific IdP in the account chooser, the UAA redirects me to this IdP as expected. However, when it executes the `login` method of class `LoginInfoEndpoint` all 10 active IdPs are queried from the DB ([coding link](https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java#L276-L279)). But as the method `login` is anyway called with the `login_hint` for selected IdP in the account chooser, this should be avoided for performance reasons.

_Further Note_
If the IdPs are configured via `discoveryUrl` the UAA even does 10 outbound calls to retrieve the `.well-known/openid-configuration` endpoints. But this can at least be avoided by a different configuration.

torsten-sap:
PR was merged

PR #1441, opened 2020-11-12
Optimize IdP requests to database on login

tack-sap:
This PR addresses Issue #1439

It includes two optimizations w.r.t. retrieving the identity providers from the database:

1) Whenever a login_hint is provided, we do not need to retrieve a full list of all identity providers, but it is sufficient to try to retrieve the identity provider with the given origin. If that does not work an error is returned anyways. So we can optimize the login method in a way, that we only retrieve a single idp, when we have a login_hint. Note that this includes the case, where we click on an account in the account_chooser page.

2) On the pages for the account_chooser and the page for entering the email for idp discovery, the list of all identity providers is retrieved but completely unused. By first determining the parameters and configuration valid for the current request, we can decide if we need to retrieve the identity providers or not. 

As the login method is called several times for each login flow, you will save the unnecessary requests also possibly multiple times per login.  

The optimized scenarios already were covered by Unit tests. The mocks for those were changed, to represent the optimized database requests. But the asserts stay the same, as the visible behavior does not change. 

The visible behavior does not change with this PR - that is also why the Integration Tests do not change. We only optimize the requests to the database (=performance improvement).

If you want to, we can also have a session where we discuss these changes.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java
@@ -274,14 +274,25 @@ public class LoginInfoEndpoint {
             clientName = (String) clientInfo.get(ClientConstants.CLIENT_NAME);
         }
 
+        //Read all configuration and parameters at the beginning to allow earlier decisions
+        boolean discoveryEnabled = IdentityZoneHolder.get().getConfig().isIdpDiscoveryEnabled();
+        boolean discoveryPerformed = Boolean.parseBoolean(request.getParameter("discoveryPerformed"));
+        String defaultIdentityProviderName = IdentityZoneHolder.get().getConfig().getDefaultIdentityProvider();
+        boolean accountChooserEnabled = IdentityZoneHolder.get().getConfig().isAccountChooserEnabled();
+        boolean otherAccountSignIn = Boolean.parseBoolean(request.getParameter("otherAccountSignIn"));
+        boolean savedAccountsEmpty = getSavedAccounts(request.getCookies(), SavedAccountOption.class).isEmpty();
+        boolean accountChooserNeeded = accountChooserEnabled && !(otherAccountSignIn || savedAccountsEmpty) && discoveryEnabled &&!discoveryPerformed;
+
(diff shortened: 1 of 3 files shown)


strehle:
@aramprice  any concerns?

aramprice:
@strehle I'll defer to the UAA team on this. I've been out of the code for too long to offer useful opinions at the moment 😞

bruce-ricard:
Hi @tack-sap 

I undid your merge commit and replaced it with a rebase:
```
git reset --hard HEAD~
git rebase develop
git push --force-with-lease
```

I wasn't able to find your commits on the command line because of the merge commit. Could you please try to do rebases in the future? AFAIU merges should only be used when merging a feature/PR branch into the main branch (and even then I personally often like to rebase), and not the other way around.

>The visible behavior does not change with this PR - that is also why the Integration Tests do not change. We only optimize the requests to the database (=performance improvement).

How would you feel about some sort of performance test? We could add a performance test section/directory in our integration tests if it doesn't exist yet. I would like to see a test that shows that indeed in practice this change is useful.

tack-sap:
@bruce-ricard Ok, I will use rebase in future - I was just not aware of this rule.

Do you have any preferences on the format of such a performance test? Should I add it here to the PR or are you just interested in having such a test to check manually?

I created a rather simple MockMvcTest today (using StopWatch and looping 1000 times over retrieving the login page with 10 Idps) - with that I see that the new implementation is constantly a few seconds faster compared to the same test executed against the develop branch (on my system: around 8s vs. constantly over 11,5s) - as it saves 10 unnecessary database reads per request. (with a growing number of idps the new version stays at the same time, while the time of the current coding keeps growing with that number).
That's only a few ms per request, but as every browser login does that, is adds up to a big amount (especially retrieving all identity providers over and over again add up to such an amount of DB requests, that the statements showed up in our monitoring as one of the most resource-consuming statements, although they are usually answered very quickly).
Should I add such a test to this PR or do you have something else in mind?

tack-sap:
Fixed a bug, that we found in our internal dev environment 
Added the MockMvc test I mentioned in my last comment that shows the performance improvements

peterhaochen47:
LGTM

Issue #1372, opened 2020-08-06
Allow an user to request his own tokens without a special scope like "tokens.list"


torsten-sap:
### What version of UAA are you running?

74.23.0

### How are you deploying the UAA?

I am deploying the UAA
- using cf-deployment

### What did you do?

As an end user, I want to retrieve the list of tokens for my own user. Thus, I am calling the existing endpoint `/oauth/token/list/user/{userId}` ([API documentation](https://docs.cloudfoundry.org/api/uaa/version/74.23.0/index.html#list-all-tokens-for-a-user)).

### What did you expect to see? What goal are you trying to achieve with the UAA?

I expect that I can call the endpoint without any special scope like "tokens.list" for my **own** user.

### What did you see instead?

The API requires scope "tokens.list". But this scope has too much authorizations in order to assign it to a common user.

shamus:
This seems like a reasonable request. Also, [revoke tokens](https://docs.cloudfoundry.org/api/uaa/version/74.23.0/index.html#revoke-tokens) has a similar access policy.

Is this something you'd like to provide a PR for?

strehle:
https://github.com/cloudfoundry/uaa/pull/1446#issuecomment-744879508

PR #1446, opened 2020-11-19
Allow users to list their own tokens

tack-sap:
Adresses #1372 

The logic to check for the same user was already implemented in the class IsSelfCheck, but was unused. So I just added the check and adopted the test accordingly.


bruce-ricard:
Hi @tack-sap . I have 3 questions and a comment:

* Are you aware that because of this [conditional](https://github.com/cloudfoundry/uaa/blob/e88983e3ac6e462b4d568be6f8360ca1ff5fa4d4/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/TokenRevocationEndpoint.java#L133-L140) , the endpoint wouldn't be listing all tokens but only the ones associated with the [client from the token](https://github.com/cloudfoundry/uaa/blob/e88983e3ac6e462b4d568be6f8360ca1ff5fa4d4/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/TokenRevocationEndpoint.java#L121-L122)? The endpoint would only list ALL tokens if you are authenticated with the scopes `uaa.admin` or `tokens.list` [code here](https://github.com/cloudfoundry/uaa/blob/e88983e3ac6e462b4d568be6f8360ca1ff5fa4d4/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/TokenRevocationEndpoint.java#L133-L137).
I am making sure because the title of your PR seems to mean that you want to retrieve all tokens.

* Why is this important to you? I am guessing you'd like to have users be able to look at all the tokens that have been issued in their name in order to control them and maybe revoke some of them?

* This feature used to exist in UAA until 2016, and was [removed](https://github.com/cloudfoundry/uaa/commit/0bf4223889dcd028a734adf3eb69c5e89293c53d). Look in particular at the `oauth-endpoints.xml` file which removes the `@self.isTokenListForAuthenticatedClient(request)` from the `access=` field, and the `listUserTokenAsSelf()` test which explicitly expects that this behavior is no longer supported.  The documentation either in the git commit or in the linked tracker story are very minimal, so I can't tell at this point why this decision was made, so I am a bit nervous about re-implementing it without investigating more and understanding why that feature was removed.

* The title of the PR throws me off a bit. If a user wants to connect to the UAA, they are not going to use a token. Tokens are used by clients. Since you care about the scope in the token, it seems to me like what you want if for clients owning user tokens to be able to list all user tokens. Is that correct?

torsten-sap:
@bruce-ricard Thanks for the review and comments.

The request actually came from a customer. I initially thought that the customer is interested in all tokens of the particular user. After your comment I double-checked the conversation with the customer. From this I got the impression that the tokens of the user with the corresponding client ID would be sufficient. As I am not 100% sure I contacted the customer again and asked him to comment/confirm here if possible.

Regarding the title, we usually refer to technical tokens (grant type client_credentials) and user tokens (all other grant types for which a user us associated with it). Your assumption is correct, sorry for the confusion.

Concluding, let us wait for the customer response before continuing with this PR.

bruce-ricard:
From our discussion on [Slack](https://cloudfoundry.slack.com/archives/C03FXANBV/p1607698116151300?thread_ts=1606844750.123000&cid=C03FXANBV), the feature is no longer needed.

Feel free to reopen the PR if things change in the future.

PR #1470, opened 2020-12-09
cleanup tomcat dependency

strehle:
use tomcat dependency from spring-boot

This currently means 9.0.39, but will be upgraded automatically with next
spring boot update.


strehle:
@shamus  as followup from last dep. discussion we may omit tomcat version, Ok ?

strehle:
@shamus 
The bot UAA Identity Bot
could do the version bump similar to
cloudfoundry/uaa-release@e7b704a

PR #1472, opened 2020-12-09
more memory for gradle tests

strehle:
in pipeline I noticed often a failure because of memory
found here recommendations:
https://stackoverflow.com/questions/38967991/why-are-my-gradle-builds-dying-with-exit-code-137


strehle:
since the pipeline errors are network releated - more or less - I close this

PR #1491, opened 2021-01-13
Refactor jasmine-test.sh

strehle:
Refactor jasmine-test.sh after PR #1435 

The test run showed, the the needed docker image is used, but
the ./scripts/jasmine-test.sh script is executed locally in the worker.
The worker has installed locally node 8.x with npm 5.x. The errors shows this.
The docker image cfidentity/uaa-singular has node 12.x with npm 6.4.

Finally: error occurs now, because new jasmine depends on newer node, e.g. node 10 or high.
Before the local test was working also with a node 8.x

Ensure that jasmine-test.sh check if it runs in docker and if not, start itself in cfidentity/uaa-singular

npm ci should be used instead npm install therefore update package-lock.json


peterhaochen47:
Hi @strehle, thanks for looking out for the test failure caused by the jasmine bump.
However, we think the approach proposed in this PR would not work because there is some significant obstacles to running docker inside of a docker image. 

Instead, I think we can simply update the docker image where this jasmine test is run to have the correct versions of node & npm. For now, we are going to revert the PR that bumped jasmine (#1435) & and fix the test failure on that dependabot-initiated PR branch first.

strehle:
@peterhaochen47 thank you for reverting the node update, so that pipeline is green again.
Yes I am check also regulary status there to prevent situation that we cannot release.

However the node update for jasmine is something I would do , but the pipeline shows, that it needs an update of used node/npm, see https://hush-house.pivotal.io/teams/cf-uaa/pipelines/uaa-acceptance-gcp/jobs/unit-tests-javascript/builds/390
╭───────────────────────────────────────╮
   │                                       │
   │   Update available 5.5.1 → 6.14.11    │
   │      Run npm i -g npm to update       │
   │                                       │
   ╰───────────────────────────────────────╯

This version means outdated node, with node 12.x a npm version 6.x would be part, so either you update the local node version in the workers or you ensure that the used docker image cfidentity/uaa-singular is really used.
Test shows the pull of cfidentity/uaa-singular but later it is not used. If I manually pull and start cfidentity/uaa-singular then I see a recent node and npm version.

peterhaochen47:
Hi @strehle, thanks for identifying the source of the problem as the node/npm versions not matching what the new jasmine wants.
The reason why the CI job is not using the node/npm in the latest `cfidentity/uaa-singular` is because the `cfidentity/uaa-singular` image is tagged and the CI config specifies which tagged image it uses. We fixed it by updating that specification:
```
  type: docker-image
  source:
    repository: cfidentity/uaa-singular	  
    tag: 8.0.0	 # change to the newest => 92.77.0
```

Now the test is passing on #1493, which can now be merged.

strehle:
Ok thanks for the info

Issue #1529, opened 2021-02-23
Issue with JWT bearer token grant in case the initial JWT was issued by the UAA


torsten-sap:
### What version of UAA are you running?

75.0.0

### How are you deploying the UAA?

I am deploying the UAA using cf-deployment

### What did you do?

* cf login
* cf oauth-token
* POST to /oauth/token with JWT bearer token grant

### What did you expect to see? What goal are you trying to achieve with the UAA?

I expect to get a token of the user from which I initially retrieved the oauth token via "cf oauth-token".

### What did you see instead?

A new shadow user was created based on the "sub" claim of the JWT retrieved via "cf oauth-token". And the retrieved JWT was issued for that newly created shadow user.

**Details**
During JWT Bearer with an UAA-issued token, the UAA constructs an internal Identity Provider config that is used in the authentication of JWT Bearer. This internal config does not define any [attributeMappings](https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java#L174-L181).

The issue seems to appear after the merge of [PR#1330](https://github.com/cloudfoundry/uaa/pull/1330) which caused the user to be updated when the externalID changes, including the username. And as the attribute mapping is empty, the default is used, which is the "sub" claim of the token. But it should be "username" to "username".

We will provide a PR that fixes the issue.

peterhaochen47:
Thanks @torsten-sap, we started looking at this. May I ask you to clarify your description of the reproduction steps? Since these steps involve multiple tokens and multiple "retrieval" processes, it's not clear which tokens you are referring to at different points. Let me do some guess-work & try to rephrase the description:

**What did you do?**
- Run `cf login` with `user-a`, ~~a UAA internal user (aka not from an external IDP)~~ a user from an external IDP that us integrated with UAA.
- Run `cf oauth-token`, capture the outputted access token `access-token-foo`; When I decode `access-token-foo`, I can see its `user_name` claim is `user-a` and `sub` claim is `sub-user-a` (just an example value; the `sub` claim is usually the user id)
- POST to /oauth/token with [JWT bearer token grant](https://docs.cloudfoundry.org/api/uaa/version/75.0.0/index.html#jwt-bearer-token-grant), where the request param `assertion` is `access-token-foo`; UAA responds with an access token `access-token-bar`.
 
**What did you expect to see? What goal are you trying to achieve with the UAA?**
[...]

PR #1530, opened 2021-03-03
Fix: Username problem with JWT Bearer grant using UAA token

tack-sap:
Adresses #1529 

Fixes the described problem, that the JWT Bearer grant to exchange a token issued by the UAA against another token was not able to identify the user correctly, but was using a wrong username - resulting in a wrong username in the token and also in the updated database entry.

To fix the problem with the wrong username we add a mapping of the user_name claim to the internal representation of the UAA idp. With that the UAA will correctly identify the username in the given token as such. The other claims are already identified by default if contained in the token.

To prevent any updates to the database (e.g. by the changed externalId, which usually is not contained in the token and defaults to "sub") - any update of the entry is blocked, if this we recognize a that it is JWT Bearer with an UAA token (basically repeating the check that is already done earlier - unfortunately there is no possibility to pass this information forward, due to the interfaces implemented here)

The scenario was already covered by tests, but those tests were mocking the "sub" attribute of a UAA token as username, while "sub" is really the userid and only the user_name claim contains the username in UAA tokens. So it the problem was not visible in the tests. I adopted the tests to mock the uaa token correctly and now it is covered by the tests.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
@@ -406,21 +406,41 @@ public class ExternalOAuthAuthenticationManager extends ExternalLoginAuthenticat
         }
 
         //we must check and see if the email address has changed between authentications
-            if (haveUserAttributesChanged(userFromDb, userFromRequest)) {
-                logger.debug("User attributed have changed, updating them.");
-                userFromDb = userFromDb.modifyAttributes(email,
-                                                         userFromRequest.getGivenName(),
-                                                         userFromRequest.getFamilyName(),
-                                                         userFromRequest.getPhoneNumber(),
-                                                         userFromRequest.getExternalId(),
-                                                         userFromDb.isVerified() || userFromRequest.isVerified())
-                    .modifyUsername(userFromRequest.getUsername());
-                userModified = true;
-            }
+        if (haveUserAttributesChanged(userFromDb, userFromRequest) && !isIdTokenIssuedByUaaWithoutIdpRegistration(request)) {
+            logger.debug("User attributed have changed, updating them.");
+            userFromDb = userFromDb.modifyAttributes(email,
+                                                     userFromRequest.getGivenName(),
+                                                     userFromRequest.getFamilyName(),
+                                                     userFromRequest.getPhoneNumber(),
+                                                     userFromRequest.getExternalId(),
+                                                     userFromDb.isVerified() || userFromRequest.isVerified())
+                .modifyUsername(userFromRequest.getUsername());
+            userModified = true;
+        }
 
         ExternalGroupAuthorizationEvent event = new ExternalGroupAuthorizationEvent(userFromDb, userModified, userFromRequest.getAuthorities(), true);
         publish(event);
         return getUserDatabase().retrieveUserById(userFromDb.getId());
+    }
+
+    private boolean isIdTokenIssuedByUaaWithoutIdpRegistration(Authentication request) {
+        String idToken = ((ExternalOAuthCodeToken) request).getIdToken();
+        if (idToken == null) {
+            return false;
+        }
+        String claimsString = JwtHelper.decode(ofNullable(idToken).orElse("")).getClaims();
+        Map<String, Object> claims = JsonUtils.readValue(claimsString, new TypeReference<Map<String, Object>>() {});
+        String issuer = (String) claims.get(ClaimConstants.ISS);
+        if (idTokenWasIssuedByTheUaa(issuer)) {
+            try {
+                ((ExternalOAuthProviderConfigurator) getProviderProvisioning()).retrieveByIssuer(issuer, IdentityZoneHolder.get().getId());
+                return false;
+            } catch (IncorrectResultSizeDataAccessException e) {
+                return true;
+            }
+        } else {
+            return false;
+        }
     }
 
     @Override
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
@@ -140,15 +140,14 @@ public class ExternalOAuthAuthenticationManager extends ExternalLoginAuthenticat
 
     public IdentityProvider resolveOriginProvider(String idToken) throws AuthenticationException {
         try {
-            String claimsString = JwtHelper.decode(ofNullable(idToken).orElse("")).getClaims();
-            Map<String, Object> claims = JsonUtils.readValue(claimsString, new TypeReference<Map<String, Object>>() {});
+            Map<String, Object> claims = parseClaimsFromIdTokenString(idToken);
             String issuer = (String) claims.get(ClaimConstants.ISS);
             if (isEmpty(issuer)) {
                 throw new InsufficientAuthenticationException("Issuer is missing in id_token");
             }
             //1. Check if issuer is registered provider
             try {
-                return ((ExternalOAuthProviderConfigurator) getProviderProvisioning()).retrieveByIssuer(issuer, IdentityZoneHolder.get().getId());
+                return retrieveRegisteredIdentityProviderByIssuer(issuer);
             } catch (IncorrectResultSizeDataAccessException x) {
                 logger.debug("No registered identity provider found for given issuer. Checking for uaa.");
             }
@@ -165,6 +164,15 @@ public class ExternalOAuthAuthenticationManager extends ExternalLoginAuthenticat
         } catch (IllegalArgumentException | JsonUtils.JsonUtilException x) {
             throw new InsufficientAuthenticationException("Unable to decode expected id_token");
         }
+    }
+
+    private IdentityProvider retrieveRegisteredIdentityProviderByIssuer(String issuer) {
+        return ((ExternalOAuthProviderConfigurator) getProviderProvisioning()).retrieveByIssuer(issuer, IdentityZoneHolder.get().getId());
+    }
+
+    private Map<String, Object> parseClaimsFromIdTokenString(String idToken) {
+        String claimsString = JwtHelper.decode(ofNullable(idToken).orElse("")).getClaims();
+        return JsonUtils.readValue(claimsString, new TypeReference<Map<String, Object>>() {});
     }
 
     private boolean idTokenWasIssuedByTheUaa(String issuer) {
@@ -428,12 +436,12 @@ public class ExternalOAuthAuthenticationManager extends ExternalLoginAuthenticat
         if (idToken == null) {
             return true;
         }
-        String claimsString = JwtHelper.decode(ofNullable(idToken).orElse("")).getClaims();
-        Map<String, Object> claims = JsonUtils.readValue(claimsString, new TypeReference<Map<String, Object>>() {});
+        Map<String, Object> claims = parseClaimsFromIdTokenString(idToken);
         String issuer = (String) claims.get(ClaimConstants.ISS);
         if (idTokenWasIssuedByTheUaa(issuer)) {
             try {
-                ((ExternalOAuthProviderConfigurator) getProviderProvisioning()).retrieveByIssuer(issuer, IdentityZoneHolder.get().getId());
+                // check if the UAA Identity Zone is registered as an external Idp of itself
+                retrieveRegisteredIdentityProviderByIssuer(issuer);
                 return true;
             } catch (IncorrectResultSizeDataAccessException e) {
                 return false;
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
@@ -406,7 +406,7 @@ public class ExternalOAuthAuthenticationManager extends ExternalLoginAuthenticat
         }
 
         //we must check and see if the email address has changed between authentications
-        if (haveUserAttributesChanged(userFromDb, userFromRequest) && !isIdTokenIssuedByUaaWithoutIdpRegistration(request)) {
+        if (haveUserAttributesChanged(userFromDb, userFromRequest) && isRegisteredIdpAuthentication(request)) {
             logger.debug("User attributed have changed, updating them.");
             userFromDb = userFromDb.modifyAttributes(email,
                                                      userFromRequest.getGivenName(),
@@ -423,10 +423,10 @@ public class ExternalOAuthAuthenticationManager extends ExternalLoginAuthenticat
         return getUserDatabase().retrieveUserById(userFromDb.getId());
     }
 
-    private boolean isIdTokenIssuedByUaaWithoutIdpRegistration(Authentication request) {
+    private boolean isRegisteredIdpAuthentication(Authentication request) {
         String idToken = ((ExternalOAuthCodeToken) request).getIdToken();
         if (idToken == null) {
-            return false;
+            return true;
         }
         String claimsString = JwtHelper.decode(ofNullable(idToken).orElse("")).getClaims();
         Map<String, Object> claims = JsonUtils.readValue(claimsString, new TypeReference<Map<String, Object>>() {});
@@ -434,12 +434,12 @@ public class ExternalOAuthAuthenticationManager extends ExternalLoginAuthenticat
         if (idTokenWasIssuedByTheUaa(issuer)) {
             try {
                 ((ExternalOAuthProviderConfigurator) getProviderProvisioning()).retrieveByIssuer(issuer, IdentityZoneHolder.get().getId());
-                return false;
-            } catch (IncorrectResultSizeDataAccessException e) {
                 return true;
+            } catch (IncorrectResultSizeDataAccessException e) {
+                return false;
             }
         } else {
-            return false;
+            return true;
         }
     }
 
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
@@ -174,6 +174,8 @@ public class ExternalOAuthAuthenticationManager extends ExternalLoginAuthenticat
     private IdentityProvider buildInternalUaaIdpConfig(String issuer, String originKey) {
         OIDCIdentityProviderDefinition uaaOidcProviderConfig = new OIDCIdentityProviderDefinition();
         uaaOidcProviderConfig.setIssuer(issuer);
+        Map<String, Object> userNameMapping = Collections.singletonMap(USER_NAME_ATTRIBUTE_NAME, USER_NAME_ATTRIBUTE_NAME);
+        uaaOidcProviderConfig.setAttributeMappings(userNameMapping);
         IdentityProvider<OIDCIdentityProviderDefinition> uaaIdp = new IdentityProvider<>();
         uaaIdp.setOriginKey(originKey);
         uaaIdp.setConfig(uaaOidcProviderConfig);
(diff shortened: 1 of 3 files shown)


peterhaochen47:
@tack-sap, a few initial questions:
1. In issue #1529, your colleague says "The issue seems to appear after the merge of PR #1330". I don't understand how #1330 caused #1529. The issue, as they said, is ultimately caused by the [lack of attribute mapping](https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java#L174-L181) of the UAA IDP object: 
  * The lack of attribute mapping results in [using the `sub` claim of the incoming token as the userFromRequest's username](https://github.com/cloudfoundry/uaa/blob/b3d9fb1482f98d038dc654b41d5a08224a546bc2/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java#L222)
  * Then it uses userFromRequest's username (aka the `sub` of the incoming token) to [search UAA database](https://github.com/cloudfoundry/uaa/blob/b3d9fb1482f98d038dc654b41d5a08224a546bc2/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/manager/ExternalLoginAuthenticationManager.java#L127), which will yield no search result, since the correct shadow user has a different username.
  * When there's no search result, it [creates a new shadow user](https://github.com/cloudfoundry/uaa/blob/b3d9fb1482f98d038dc654b41d5a08224a546bc2/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/manager/ExternalLoginAuthenticationManager.java#L139) with the `sub` claim of the incoming token as the username. This new shadow user is an incorrect duplicate and what causes the issue in #1529.
The code path described above hasn't reached the code change in #1330 yet, so I wonder how #1330 directly caused this problem?

2. I find the commit message "Only update user if token not issued by UAA" inaccurate. Based on the definition of `isIdTokenIssuedByUaaWithoutIdpRegistration` (which I think means the token is issued by UAA for a UAA internal store user), doesn't the updated code mean that it WILL update user if the token is issue by UAA for a user from a UAA-integrated external IDP? Please correct me if I'm reading the function wrong. In that case, perhaps the commit message should be something like "Only update user if token is for a user from an external IDP"?

3. Separately, I also don't understand why [this commit](https://github.com/cloudfoundry/uaa/pull/1530/commits/46cb27430410b3839fba3c69c34d0e2e83d40113) is a necessary code change for #1529, for same reason as my comment 1) above. Do you still see the symptoms described in the issue without this commit?

Thanks!

tack-sap:
Hi, 

Let me try to answer your questions:

Regarding 1) You're right, that the problem is ultimately caused by the lack of attribute mapping of the UAA IdP Object. However the flow is a little different from what you describe:
Note that the username is defaulting to the claim `sub`, while other claims like email, given_name, family_name default to their corresponding values (so they are resolved correctly). In addition the `externalId` always resolves to the "sub" claim of the token. 

After searching for the username by username/sub claim, the UAA will [look for the shadow user by email](https://github.com/cloudfoundry/uaa/blob/b3d9fb1482f98d038dc654b41d5a08224a546bc2/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/manager/ExternalLoginAuthenticationManager.java#L131). As the email is resolved correctly, this will result in the correct shadow user - its just the username that is different.
Before #1330 this was no problem: The [username is not checked](https://github.com/cloudfoundry/uaa/blob/b3d9fb1482f98d038dc654b41d5a08224a546bc2/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/manager/ExternalLoginAuthenticationManager.java#L297), when check if the user attributes changed for a user. It is, however [updated](https://github.com/cloudfoundry/uaa/blob/b3d9fb1482f98d038dc654b41d5a08224a546bc2/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java#L415), when some of the other attributes changed. As the externalId always changes, the username is also updated. And before #1330 externalId was not included in the attributes to check for changes, and as the other attributes did not change there was no update done.

Regarding 2) The method checks, whether the token that we have at hand was issued by the UAA itself (basically the same check we do earlier, when we determine the IdP for the token). More precisely: We check whether the issuer is the UAA (same identity zone) AND we do not have a registration of this issuer as identity provider in this identity zone (which is something that is mainly used in the tests)
We only update the user here if the token was NOT issued by the UAA.

Regarding 3) The externalId will always change for a user, as this always maps to `sub`. As we do not want the externalId to change with JWT Bearer, we just don't update the user, if the token comes from the UAA itself. (Generally speaking: There cannot be any information in a token that was issued by the same UAA identity zone, that is used to update the same user in the database. So we just don't update in that case.)
The extensive check here is needed for the situation described above, where the same identity zone is registered as OIDC provider within itself (again: mainly used in tests) - so we need to do the same check we already did for determining the IdP (but we cannot just pass forward this information due to the interfaces that are used here, so we check it again)

Hope this answers your questions. If you have any further questions or proposals for this change, just let me know.

peterhaochen47:
Thanks @tack-sap,

1. You are right. I now see that I've missed the fact that searching for the existing shadow user in DB works like this:
- Then it uses userFromRequest's username (aka the `sub-user-a`) to [search in UAA user database](https://github.com/cloudfoundry/uaa/blob/b3d9fb1482f98d038dc654b41d5a08224a546bc2/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/manager/ExternalLoginAuthenticationManager.java#L127), which will yield no search result. As a fallback, it then [uses userFromRequest's email to search in UAA user database](https://github.com/cloudfoundry/uaa/blob/b3d9fb1482f98d038dc654b41d5a08224a546bc2/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/manager/ExternalLoginAuthenticationManager.java#L131), which will successfully yield the previously populated shadow user `user-a`. 

However, I am still fuzzy on what happens next, but I'm working through it - so l'll circle back on the precise flow tomorrow.

2.

I have a hard time connecting the statement `We only update the user here if the token was NOT issued by the UAA` with the code below:
```
if (haveUserAttributesChanged(userFromDb, userFromRequest) && !isIdTokenIssuedByUaaWithoutIdpRegistration(request)) {
 # UPDATE THE USER
}

    private boolean isIdTokenIssuedByUaaWithoutIdpRegistration(Authentication request) {
        String idToken = ((ExternalOAuthCodeToken) request).getIdToken();
        if (idToken == null) {
            return false;
        }
        String claimsString = JwtHelper.decode(ofNullable(idToken).orElse("")).getClaims();
        Map<String, Object> claims = JsonUtils.readValue(claimsString, new TypeReference<Map<String, Object>>() {});
        String issuer = (String) claims.get(ClaimConstants.ISS);
        if (idTokenWasIssuedByTheUaa(issuer)) {
            try {
                ((ExternalOAuthProviderConfigurator) getProviderProvisioning()).retrieveByIssuer(issuer, IdentityZoneHolder.get().getId());
                return false;
            } catch (IncorrectResultSizeDataAccessException e) {
                return true;
            }
        } else {
            return false;
        }
    }
```
Let's enumerate all the possibilities for `access-token-foo` (the incoming token described in the github issue):
- the token is issued by UAA:
 1. the uaa-issued token is for a user from UAA's internal user store
 2. the uaa-issued token is for a user from an external IDP that is registered with UAA (this is the case that the issue #1529 focuses on)
- the token is NOT issued by UAA:
 3. the externally-issued token is for a user from an external IDP that is registered with UAA
 4. the externally-issued token is for a user from an external IDP that is NOT registered with UAA

The existing code of [resolveOriginProvider](https://github.com/cloudfoundry/uaa/blob/46cb27430410b3839fba3c69c34d0e2e83d40113/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java#L141) is:
- first, checking if issuer is a registered provider, aka whether the token belongs to case 3; if so, returns the registered provider
- second, checking if issuer is UAA itself, aka whether the token belongs to case 1 or 2; if so, return the internal representation of the UAA IDP (which is what your PR focused on)

Going back to the definition of `isIdTokenIssuedByUaaWithoutIdpRegistration`, this function reads to me as:
- first checking if the token belongs to case 1 or 2 (`idTokenWasIssuedByTheUaa(issuer)`)
  - if so, checking if the token belongs to category 3 (`((ExternalOAuthProviderConfigurator) getProviderProvisioning()).retrieveByIssuer(issuer, IdentityZoneHolder.get().getId());`); if so, returns `false`, otherwise returns `true`

So, the second check doesn't quite make sense to me? Could you describe to me what the check `!isIdTokenIssuedByUaaWithoutIdpRegistration` intends to do in terms of those 4 cases?

A thought on readability: the presence of the negation `!` with `isIdTokenIssuedByUaaWithoutIdpRegistration` to signal when to NOT update the shadow user is perhaps harder to read than signaling when to update the user? Is there a way to write this positively? Such as `if (haveUserAttributesChanged(userFromDb, userFromRequest) && someOtherConditionIsTrue(request)) {# UPDATE THE USER}`?

Perhaps the difficulty in understanding why this check is needed (and what this check means) lies in that [the commit](https://github.com/cloudfoundry/uaa/pull/1530/commits/46cb27430410b3839fba3c69c34d0e2e83d40113) doesn't come with a new test / changed test? A path forward would be to add such a test.

tack-sap:
Hi,

Regarding 2) 
At the point where this method is called, 4 cannot be the case anymore, this is ruled out earlier in the flow.
With idTokenWasIssuedByTheUaa() we also rule out case 3 - if we have case 3 we will update the user if any of the attributes changed.

The cases 1 and 2 really don't make a difference at this point either. The difference we really check here is: Do we have a registration of an identity provider with this issuer?
This is mostly done in the tests, where e.g. identity zone A is used as OIDC identity provider in identity zone A itself. If we have such a registration, we have to treat is as external idp (so it is really case 3 and not case 1&2).
This is why we look for an identity provider registration with this issuer in this method - and if there is no such registration we know that we should not update the shadow user in the database.

The order of the checks is a little different here, as we don't really need to know the exact idp if we have a different issuer, so we can rule all of those cases out early (and save e.g. another request of the idp to the database, which is not very efficient by issuer)

If you want to we can also rename the method to something like isRegisteredIdpAuthentication() or similar. Would this name fit for you?

The whole check here was something I found due to failing tests without this commit. So basically this is already covered by existing tests.

peterhaochen47:
Hi Thanks! @tack-sap 
To test the necessity of [this commit](https://github.com/cloudfoundry/uaa/pull/1530/commits/46cb27430410b3839fba3c69c34d0e2e83d40113), I ran the tests & found that all tests passed without this commit. To demonstrate this, I created [this test PR](https://github.com/cloudfoundry/uaa/pull/1549) that only contains the first commit; and this test PR passed the CIs too. Could the `failing tests without this commit` be due to some other reasons?

tack-sap:
Hi @peterhaochen47,
Yes, you're right. Looks like I mixed up some of the changes there and the test failures I saw during development were not what caused the second commit, but something different. Sorry for the confusion.

Nevertheless I still think, that the change in the second commit is needed. I will update the PR with some additional tests that demonstrate the situations that are covered by the second commit.

tack-sap:
@peterhaochen47 I added some additional tests. 
The tests illustrate in which situations the shadow user from different in the UAA is updated during JWT Bearer.
You see that without the second commit an internal user will be updated (in this case: the externalId will be updated), when it should not.
The other test cases illustrate cases, that could behave wrong, when the check is done in a different way (e.g. if you just have a look at the issuer without looking at the registered idps or if you don't update the user at all during JWT Bearer).

I also added a renaming of the check to a positive meaning. (But did this in a separate commit, so you can cherry-pick and try the tests more easily)

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/token/JwtBearerGrantMockMvcTests.java
```diff
@@ -84,6 +89,85 @@ void default_zone_jwt_grant() throws Exception {
...
+        performJWTBearerGrantForJWT(targetZone, accessTokenForOriginZoneUser);
+
+        //Verify username and External ID not changed after this internal grant
+        ScimUser originShadowUserAfterExchange = getScimUser(originUser.getEmails().get(0).getValue(), originZoneOriginKey, targetZone.getId());
+        assertEquals(originShadowUser.getUserName(), originShadowUserAfterExchange.getUserName());
+        assertEquals(originShadowUser.getExternalId(), originShadowUserAfterExchange.getExternalId());
+    }
+
+    @Test
+    void default_zone_jwt_grant_user_update_same_zone_with_registration() throws Exception {
```
peterhaochen47: I'm confused about why this test is about `default_zone`? This test doesn't seem to engage with the default zone.

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/token/JwtBearerGrantMockMvcTests.java
```diff
@@ -84,6 +89,85 @@ void default_zone_jwt_grant() throws Exception {
...
+        assertEquals(targetZoneUser.getUserName(), targetUserAfterGrant.getUserName());
+        assertEquals(targetZoneUser.getExternalId(), targetUserAfterGrant.getExternalId());
+
+        //Check for user of registered IdP
+        String originZoneIdToken = getUaaIdToken(originZone.getIdentityZone(), originClient, originUser);
+        String accessTokenForOriginZoneUser = performJWTBearerGrantForJWT(targetZone, originZoneIdToken);
+        Map<String, Object> originUserClaims = JwtTokenUtils.getClaimsForToken(accessTokenForOriginZoneUser);
+
+        //Verify values for new shadow user set
+        ScimUser originShadowUser = getScimUser(originUser.getEmails().get(0).getValue(), originZoneOriginKey, targetZone.getId());
```
peterhaochen47: Could we rename this to just "shadowUser" perhaps? Since there is only one shadow user involved. Previously defined vars like "originUser" means "a user in the origin zone"; "originClient" means "client in origin zone". Based on that semantic, now "originShadowUser" might mean "shadow user in origin zone", which is not true, right? It actually means the "originUser's" shadow user in the target zone.

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/token/JwtBearerGrantMockMvcTests.java
peterhaochen47: If I'm reading this correctly, "same_zone_with_registration" means "registering a zone as an External IDP in itself", which seems like an unusual use case. Is this a documented use case? If not, is there a reason why we should make sure this use case works?

Isn't the JWT bearer grant that involves an external IDP already tested elsewhere?

peterhaochen47:
Thanks! Seeing the added tests, I have some basis for understanding the need for the check now, though I have a few questions, which I posted above. 

I also saw a few opportunities for refactors, which I wonder if it'll be okay to make commits on my end and add to this PR.

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/token/JwtBearerGrantMockMvcTests.java
tack-sap: Sorry, my bad. I started this test with the default zone and found that it was easier with the provided methods to use a non-default zone instead. I just forgot to rename the test after this change.

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/token/JwtBearerGrantMockMvcTests.java
tack-sap: Yes, the latter one is correct. It is the shadow user that exists in the target zone and represents the internal user from the origin zone (which is the idp in this case). Sure we can rename that variable.

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/token/JwtBearerGrantMockMvcTests.java
tack-sap: I agree that this is a rather unusual case - but it is handled when resolving the internal idp at the moment and also used in other tests, so I wanted to handle this case here too.

tack-sap:
Hi, I tried to answer your questions. 

If you see opportunities for a refactoring feel free to add them here - the important thing for us is just, that the bug gets fixed.

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/token/JwtBearerGrantMockMvcTests.java
peterhaochen47: Could you help me find where those other tests are? I wanna try to understand why they are written that way.

peterhaochen47:
@tack-sap Thanks for answering all the questions. Please feel free to make a commit to rename the test name and var name mentioned in the first 2 comments. And we'll go over this tomorrow & make the refactor commit & I think we'll be ready.

torsten-sap:
@peterhaochen47 Do we have any open points or can this PR be merged?

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
```diff
@@ -404,23 +406,43 @@ protected UaaUser userAuthenticated(Authentication request, UaaUser userFromRequ
...
 
         ExternalGroupAuthorizationEvent event = new ExternalGroupAuthorizationEvent(userFromDb, userModified, userFromRequest.getAuthorities(), true);
         publish(event);
         return getUserDatabase().retrieveUserById(userFromDb.getId());
     }
 
+    private boolean isRegisteredIdpAuthentication(Authentication request) {
+        String idToken = ((ExternalOAuthCodeToken) request).getIdToken();
+        if (idToken == null) {
+            return true;
```
peterhaochen47, line 437: When ID Token is somehow not present, shouldn't we default to "false" here? Aka not to update the user.

peterhaochen47:
@torsten-sap @tack-sap Thanks! The only points are:
1. my question about self-registering IDP: https://github.com/cloudfoundry/uaa/pull/1530#discussion_r606012928
2. my question about the case when ID token is not present: https://github.com/cloudfoundry/uaa/pull/1530#discussion_r610767194
3. I have a refactor commit that I have yet to push (and will push soon)

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/token/JwtBearerGrantMockMvcTests.java
tack-sap: Hi,

I know that at least the tests the [PasswordGrantIT](https://github.com/cloudfoundry/uaa/blob/develop/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/PasswordGrantIT.java) that is using it this way (though only for the OIDC password grant). Not sure if there are others also doing it this way.

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
tack-sap: If the idToken will be empty during OAuth Authorization Code Flow. But the user information we have at that point won't be empty. As the information in such a case always comes from the external idp, we should update the user in that case. If we return false here the user will not be updated during auth code flow at all.

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/token/JwtBearerGrantMockMvcTests.java
peterhaochen47: Interesting, I wonder if this is for the convenience of test setup or a real-life use case. But the current code to support this use case, though confusing to read (mitigated via a comment), likely won't cause any issue, so let's leave it be.

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
peterhaochen47: I see. If this empty idToken case is important (rather than just a "make-java-happy null check"), then a test would be helpful. Currently, all the tests still pass when I change it to "false". I'll leave it up to you to decide.

peterhaochen47:
Thanks! All open questions replied/addressed (please see above conversations). Please feel free to rebase & merge.

PR #1566, opened 2021-05-05
fix: passcode prompt for json 

tack-sap:
This PR fixes a bug that we saw in the latest release and that was introduced by #1441 

When you configure the account chooser and/or idp discovery (in our case both of them are configured) and then retrieve the login info as json (as it is done by cf cli), the passcode prompt will be missing from the returned information. This results in an empty prompt shown by the cf cli, which is a bad experience for using the cli.

The cause of this problem is, that the prompt for passcode will be filtered out when no external idp is passed to populate prompts. But with the optimization from #1441 the list is always empty, as it is not needed for the login page. For html the passcode is always filtered from the prompts, so it does not make any difference there, but for the json this now makes a difference.

The proposed approach to fix this is to skip the optimization in case the info is retrieved for json - and keep everything as it is for html. I also added new test cases that now check that the passcode prompt is still included in the response for json even if either discovery or account chooser are activated.


peterhaochen47:
LGTM

PR #1576, opened 2021-06-07
omit json-smart library

strehle:
fix CVE-2021-27568

Use SPI from https://github.com/json-path/JsonPath and set existing jackson library instead of minidev library
The used json-smart library brings uaa the CVE https://nvd.nist.gov/vuln/detail/CVE-2021-27568 but since json-path support other JSON parsers the solution should be to use exsiting library instead - without CVE


strehle:
@torsten-sap  FYI

strehle:
new spring boot fixed dependency

Issue #1584, opened 2021-06-17
Bug in test suite: Logging in against client that only allows SAML IDP should redirect


staylor14:
### What version of UAA are you running?
This is a test suite bug

### How are you deploying the UAA?
This is a test suite bug

### What did you do?
This line declares `SAML` is the only allowed IDP:
https://github.com/cloudfoundry/uaa/blob/db47c2b1c4a66da22f75a9159076887cfb5adf41/uaa/src/test/java/org/cloudfoundry/identity/uaa/login/LoginMockMvcTests.java#L303

This test passes:
https://github.com/cloudfoundry/uaa/blob/db47c2b1c4a66da22f75a9159076887cfb5adf41/uaa/src/test/java/org/cloudfoundry/identity/uaa/login/LoginMockMvcTests.java#L281-L287

### What did you expect to see? What goal are you trying to achieve with the UAA?
The test should fail.  UAA should redirect in this scenario.

### What did you see instead?
The test passes.

### Further details
[I've created a branch](https://github.com/cloudfoundry/uaa/tree/611c0b42ca1e61c3b7ff7b1730ba916b0dfd34b1) with a properly failing test to illustrate what I think should happen here.
https://github.com/cloudfoundry/uaa/blob/611c0b42ca1e61c3b7ff7b1730ba916b0dfd34b1/uaa/src/test/java/org/cloudfoundry/identity/uaa/login/LoginMockMvcTests.java#L309-L333

I suspect the same problem exists for a client that only allows `OIDC`.

I believe this subtle bug was introduced in ["Restructure login method to not read all IdentityProviders on login_hint"](https://github.com/cloudfoundry/uaa/commit/333954df0df9327cea2a75c906a63aa5fea3544e#diff-39666ac076d85b660eb868c942e21f648c6be235e4c829e5d2af7474b8fd617d)

strehle:
Solved

PR #1589, opened 2021-06-18
Fix redirect when only saml allowed

strehle:
fix for issue #1584


peterhaochen47:
cherry-picked to develop, no longer need this PR.

PR #1614, opened 2021-07-28
Change and update mail library

strehle:
Why: existing mail library has no update and this library needs own version definition

Solution: switch to spring boot maintained library, see https://github.com/spring-projects/spring-boot/blob/fa99ae774a1648e2978edaf61d155034370eaa82/spring-boot-project/spring-boot-dependencies/build.gradle#L1670


peterhaochen47:
Maybe add a note about why the library is changed? Is it because the old lib is deprecated? @strehle

strehle:
> Is it because the old lib is deprecated?

Yes see my description now https://github.com/cloudfoundry/uaa/pull/1614#issue-698724206 , ok?

PR #1615, opened 2021-07-29
Increase authorization code size

robl-sap:
Hi,

during internal tests we discovered that authorization codes issued by UAA don't meet the oAuth2.0 standard.
In particular, the [standard](https://datatracker.ietf.org/doc/html/rfc6749#section-10.10) requires 128 bits of randomness and suggests 160 bits of randomness, but UAA currently only uses 80 bits.

Therefore, we suggest to increase the size for the authorization codes from 10 to 20 characters.

Regards,
Jonas


strehle:
@robl-sap  Thanks, 2 points
a) can you sign CLA. Please click here to be authorized. For further assistance with EasyCLA, please submit a support request ticket.
b) can you please add one test with this condition and describe it about the standard, so that we have someting as regression test

robl-sap:
@strehle i signed CLA

strehle (approved):

peterhaochen47 (approved):

Issue #1620, opened 2021-07-30
SAML login fails with `InResponseToField` validation error


peterhaochen47:
### What version of UAA are you running?

v75.5.0

### How are you deploying the UAA?

I am deploying the UAA

- as part of a commercial Cloud Foundry distribution

### What did you do?

1. Configure UAA with a SAML external IDP
2. Log into UAA via the SAML provider using the latest Chrome browser

### What did you expect to see? What goal are you trying to achieve with the UAA?
See the UAA login success page after being redirected back from the SAML login page.

### What did you see instead?
See a UAA error page with the message "InResponseToField of the Response doesn't correspond to sent message"

peterhaochen47:
We decide to close the PR #1621 (This PR is only a short-term fix) for now, and instead try to come up with a long-term fix. If anyone is immediately blocked (which we haven't heard from any OSS users), they can request to reopen this PR #1621.

PR #1621, opened 2021-07-30
Fix SAML login flow by configuring JSESSIONID cookie

peterhaochen47:
- Explicitly set JSESSIONID cookie to `SameSite=None`, on which the expected SAML login relies:
After an end-user logs in via SAML, they are redirected back to UAA. At this point, UAA validates
that the SAML assertion originated from the original UAA login by examining the JSESSIONID cookie.
Currently, only `SameSite=None` would guarantee that the cookie would get sent to back UAA after
redirection from SAML login (see: https://web.dev/samesite-cookies-explained/).
- Previously, the SameSite is set to null (https://github.com/cloudfoundry/uaa/commit/8528a19f268fad41d8dd96e9).
We believe that this prior commit intends to set it to `None`, but mistakenly set it to null,
which worked at the time becaue browsers likely defaulted to `None` in the null case. However,
browsers have changed their default to `Lax`, causing our SAML login to fail. So the current
commit simply restores the prior intended cookie setting.
- Now browsers also require cookies to have the `Secure` attribute when `SameSite=None`
(see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite).
- Workaround a test that uses an implementation of httpclient that would not send
the `JSESSIONID` cookie that has `Secure` attribute, by manually attachinng the `JSESSIONID`
cookie to the request. Though this is "hacky", I deem this acceptable because there are enough
other tests that cover the login behavior; they use webdriver, which would make an exception
for HTTP on localhost (which is the standard [browser behavior](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies))

[Fixes Issue #1620]
[#178935973]

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/web/beans/UaaSessionConfig.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/web/beans/UaaSessionConfig.java
@@ -32,7 +32,8 @@ public class UaaSessionConfig {
             final @Value("${servlet.session-cookie.max-age:-1}") int cookieMaxAge
     ) {
         DefaultCookieSerializer cookieSerializer = new DefaultCookieSerializer();
-        cookieSerializer.setSameSite(null);
+        cookieSerializer.setSameSite("None");
+        cookieSerializer.setUseSecureCookie(true);
         cookieSerializer.setCookieMaxAge(cookieMaxAge);
         cookieSerializer.setCookieName("JSESSIONID");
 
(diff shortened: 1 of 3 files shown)


strehle:
> Hi @strehle, thank you for taking a look. Could you clarify what you mean? I do't have great context on what "changing the inpreponseto saving in DB similar to other codes" means.

Hi, I mean that we have other "codes", e.g. authorization codes (oauth2) or in response to (saml2) or passcode (internal sso). Some are stored in DB, e.g. authorization code or passcode, but the inreponse-code for SAML is only added to security session context and with that it dependens on cookie behaviour. I suggested to change this - store in DB - and make it independent from cookies

Another comment was, that we have a property for Secure flag already-> require_https

peterhaochen47:
@strehle, okay, we decide to close this PR for now, and instead try to come up with a long-term fix as you suggested. This PR is only a short-term fix. If anyone is immediately blocked (which we haven't heard from any OSS users), they can request to reopen this PR.

PR #1652, opened 2021-08-19
Add Vendor specific statements to retrieve authorities for user

### LINKED ISSUE #1473: Possible improvement: Usage of database specific SQL Statements in JdbcUaaUserDatabase
### What did you do?

In our monitoring of our database, we have still see that the dynamically generated SQL statements to resolve the group memberships of a user (that we have already optimized before: https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/user/JdbcUaaUserDatabase.java#L197-L228 ) are rather expensive and look confusing for a larger amount of groups that need to be resolved. We are also not so happy with the restriction of the number of parameters, that may cause that the statement needs to be split. 

### What goal are you trying to achieve?

We see a chance for further improvement, especially regarding the splitting of the statement, by altering the statement that is used depending on the underlying database. For example: All of the three currently supported databases (HSQL, PostgreSQL, MySQL) have a support for an array type as parameter in a database statement. But the syntax that needs to be used is not the same between those three, which is the reason why we currently cannot use it. 

So we would like to discuss, whether it would make sense to introduce a mechanism to use different SQL Statements here for the different databases - so each statement could be optimized for the database it is targeted on. (Currently resolving the groups is the only place where we would like to have such an improvement, but maybe we will find other places in the future)

Of course such a decision is not needed for most of the other statements, that are plain select/update/insert statements, but as this is a rather complex operation, we see potential in using the database specific operations in such a case. 

We would like to hear, what you think of such a mechanism.

tack-sap:
Adresses #1473 

With this change we introduce the possibility to use database vendor specific SQL Syntax for some SQL statements.
While this is not needed for the majority of SQL statements in the UAA (as they are rather simple and only access a single table), there are statements that we see as expensive and that could be optimized with different SQL syntax, but this is then vendor specific.

In this PR we address the statement to get the authorities of a user in the JdbcUaaUserDatabase, which is (even after several optimizations we already did) still one of the most expensive statement in our landscapes. Instead of using "IN" with a variable number of parameters we are using (for postgresql) "= ANY(?)" with an array-type parameter. Similar for HSQLDB, we are also using now a array-type parameter. (For MySQL we don't know such a optimization, but it could easily be implemented, if there is a similar way. Right now it will fall back to using the current statement)

As there could be other statements that can be optimized in a similar fashion in the future, I made the mechanism reusable: the bean could just be injected in other classes that make such a decision and they can then use different statements based on the vendor of the DB, as it is done in the JdbcUaaUserDatabase.

Currently this feature is always activated, but maybe this could also be done via some kind of switch. (However this would require new configuration options, and I tried to avoid this, so this is always activated now)

Regarding the tests: I noticed that the current tests for JdbcUaaUserDatabase are always using HSQLDB on my local machine, regardless of the DB that I provide for the run-unit-tests.sh script. So I only provided the unit tests for the added class. 
In the ITs, the database is however change with the different parameters. As there is no functional difference between the old and the new implementation, the coverage stays basically the same - only difference is that now each of the three db profiles is using a different statement.


server/src/main/java/org/cloudfoundry/identity/uaa/db/DatabaseVendorProvider.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.sql.Connection;
+import java.sql.DatabaseMetaData;
+import java.sql.SQLException;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.jdbc.core.JdbcTemplate;
+
+public class DatabaseVendorProvider {
```
strehle: This seems to be provided with class and bean from here

https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/db/DatabaseUrlModifier.java#L18-L43

server/src/main/java/org/cloudfoundry/identity/uaa/user/JdbcUaaUserDatabase.java
```diff
@@ -237,13 +242,40 @@ private String getAuthorities(final String userId) {
...
+        }
+
+        private List<Map<String,Object>> executeAuthoritiesQuery(List<String> memberList) {
+            Vendor dbVendor = databaseVendorProvider.getDatabaseVendor(jdbcTemplate);
+            if (Vendor.postgresql.equals(dbVendor)) {
+                return executeAuthoritiesQueryPostgresql(memberList);
+            } else if (Vendor.hsqldb.equals(dbVendor)) {
+                return executeAuthoritiesQueryHSQL(memberList);
+            } else {
+                return executeAuthoritiesQueryDefault(memberList);
```
strehle, line 271: This means, for MySQL the behavriour is not changed ?

uaa/src/main/webapp/WEB-INF/spring/oauth-endpoints.xml
```diff
@@ -642,6 +642,8 @@
 
     <bean id="userDatabase" class="org.cloudfoundry.identity.uaa.user.JdbcUaaUserDatabase"/>
 
+    <bean id="databaseVendorProvider" class="org.cloudfoundry.identity.uaa.db.DatabaseVendorProvider" />
```
strehle: pls check comment above, I think you dont this this bean

strehle (commented):
@tack-sap  did you have tested and forked this in SAPs UAA already ? so that we see how preformance improvements work ?

server/src/main/java/org/cloudfoundry/identity/uaa/db/DatabaseVendorProvider.java
tack-sap: Good finding, I will reuse the existing bean.

server/src/main/java/org/cloudfoundry/identity/uaa/user/JdbcUaaUserDatabase.java
tack-sap: Yes, MySQL does not support array as parameter type, according to what I found. So I decided to use the existing query.

strehle:
@peterhaochen47  currently blocked because of red  pipelines/uaa-acceptance-gcp so, please notify us if pipeline issues solved, thanks

server/src/main/java/org/cloudfoundry/identity/uaa/user/JdbcUaaUserDatabase.java
```diff
@@ -253,20 +285,23 @@ protected void getAuthorities(Set<String> authorities, final List<String> member
...
-                String displayName = (String) record.get("displayName");
-                String groupId = (String) record.get("id");
-                if (!authorities.contains(displayName)) {
-                    authorities.add(displayName);
-                    newMemberIdList.add(groupId);
-                }
-            }
-            getAuthorities(authorities, newMemberIdList);
+        private List<Map<String, Object>> executeAuthoritiesQueryPostgresql(List<String> memberList) {
+            String arrayAuthoritiesQuery = "select g.id,g.displayName from groups g, group_membership m where g.id = m.group_id  and g.identity_zone_id=? and m.member_id = ANY(?)";
```
strehle, line 294: The only comment here I have is.
Did you executed this via explain ? because of issue https://github.com/cloudfoundry/uaa/issues/1662 ?
I recommend to check for indexed query plan here and then also think about of LOWER for identity zone id

strehle (commented):
check the SQL with EXPLAIN ANALYZE on psql to be sure that we have indexed query plan

server/src/main/java/org/cloudfoundry/identity/uaa/user/JdbcUaaUserDatabase.java
tack-sap: Checked the statement via explain analyze:
the ANY(?) is using group_membership_unique_key to do an Index Only Scan
the groups id join is using groups_pkey to do an Index Scan and then filtering on identity_zone_id = 'uaa' (so no scan over identity zone id is happening here)

strehle: ok, thanks

PR #1668, opened 2021-09-18
support logout for OpenId proxy mode

strehle:
Feature description: https://openid.net/specs/openid-connect-session-1_0.html 

Support end_session_endpoint in own OIDC well-known configuration
Use end_session_endpoint from OIDC well-known configuration in Proxy Mode

Why: Support same logout (single logout) possibility SAML2 has. Needed for switching from SAML to OpenID connect scenarios. Almost all known Idp provide end_session_endpoint endpoint flow already.

Missing: id_token_hint in flow but use current parameter e.g. redirect_uri from zone and/or client and client_id


torsten-sap:
@tack-sap Could you do the review?

Issue #1480, opened 2020-12-16
Extending user model to store additional attributes?


bsekar:
I am deploying the UAA
- using a bosh release I downloaded from bosh.io

### What did you expect to see? What goal are you trying to achieve with the UAA?
Add app specific custom attributes to UAA user model.

Hi, I came across an old [issue](https://github.com/cloudfoundry/uaa/issues/554) regarding this. Was wondering if this has changed?

adarshasp87:
This is a much-needed feature. User model should allow to add custom attributes

strehle:
I know that this feature is often requested, therefore I simply answer, please propose it via PR and then we can discuss it, because I agree, community asks for this

bsekar:
Thanks @strehle . Created [PR](https://github.com/cloudfoundry/uaa/pull/1682)

PR #1682, opened 2021-09-22
Feature/custom attributes

bsekar:
Extending UAA schema to allow custom attributes based on discussion [here](https://github.com/cloudfoundry/uaa/issues/1480#issuecomment-878926436)

### DIFF
--- a/model/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUser.java
+++ b/model/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUser.java
@@ -15,13 +15,19 @@ package org.cloudfoundry.identity.uaa.scim;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.SerializerProvider;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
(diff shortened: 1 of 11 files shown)


bsekar:
We're aware of the missing CLA. We're working with our legal team to have it done

strehle:
I went through code and about users db enhancement. Why dont you use and extend user_info and info blob field which is also json as far as I see ?
To me it makes sense to (re-)use this schema and store custom attributes in info field as json similar then you do it

bsekar:
Thanks @strehle for reviewing this. Reusing the [userAttributes](https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/user/UserInfo.java#L30) in UserInfo was our initial option as well, but for the following reasons we decided to put it in ScimUser itself.


1. If one wants to update custom attributes on the user, currently there is no API to set the same. The only operation on userInfo is the OIDC standard `GET /userinfo`
1. UserInfo seems to be used when UAA is federating with external IdP (SAML etc) and we wanted this to be available natively as a first class property, when UAA acts as the user store/source of truth.

bsekar:
@strehle Did you get a chance to review the above comment?

strehle:
> Did you get a chance to review the above comment?

partly but uaa pipeline is red and release build is blocked because vmware colleagues are blocked.
So I need them for their thoughts.

Right now I can say that we (SAP) have often trouble because of high DB load and users table is one of the hotspots and thus I do not prefere to extend it, but rather have custom attributes in another table so that not all selects for user do a fetch of all values . Most of the scenarios wont need it therefore it should not be in main users db.
@br @peterhaochen47  your thoughts ?

strehle:
added other maintainers and request their feedback. 
So I wont block this if other say yes, but if nobody answer I would like to seperate the attribute persistency into another db and then I will accept it , as mentioned I recommend exiting user_info

torsten-sap (commented):
I agree with @strehle that custom attributes should not be persisted in the users table. The reasons were already mentioned.

PR #1716, opened 2021-11-09
Fix error mapping /error

strehle:
Call to /error
Resulted in http 500 but should map to mappings in HomeController

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/home/HomeController.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/home/HomeController.java
@@ -104,7 +104,7 @@ public class HomeController {
         return "error";
     }
 
-    @RequestMapping({"/error", "/error404"})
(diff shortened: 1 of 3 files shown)


strehle:
before get to http://localhost:8080/uaa/error resultet in 500 and generic error handler is not called....
after: any get to http://localhost:8080/uaa/error** is returning 200 and will reach homecontroller class

strehle:
@peterhaochen47 our monitoring has found mass of 500 which caused by errors, however calling /error should not return 500 to me. The current definition for /error is not matching / reaching HomeController but the 500 is withing spring mvc, therefore created this PR and IT to have a test for /error, see simple get

https://login.uaa-acceptance.cf-app.com/error vs. another working error handler: 
https://login.uaa-acceptance.cf-app.com/oauth_error

torsten-sap (approved):
LGTM

PR #1719, opened 2021-11-11
Add "FOR UPDATE SKIP LOCKED" DeleteExpiredQuery

### LINKED ISSUE #1705: Improve SQL for deleting revocable tokens

75.9.0

I am deploying the UAA

- locally only using gradlew with postgresql

### What did you do?

check SQL performance

no locking in SQL 

### What did you see instead?

Cleanup task from 
https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/resources/jdbc/PostgresLimitSqlAdapter.java
should be more efficient

strehle:
Postgresql 9.5 onwards supports this feature.
It prevents deadlocks in such scenarios were a delete is done with an inner select, see
https://stackoverflow.com/questions/68626800/deadlock-when-using-delete-on-postgresql

Assume that rely on the version 9.5 and higher is ok, since EOL
https://endoflife.software/applications/databases/postgresql
shows that you should have a newer DB

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/resources/jdbc/PostgresLimitSqlAdapter.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/resources/jdbc/PostgresLimitSqlAdapter.java
@@ -31,6 +31,6 @@ public class PostgresLimitSqlAdapter implements LimitSqlAdapter {
             " < ? " +
             " ORDER BY " +
             expiresColumn +
-            " LIMIT "+maxRows+"))";
+            " LIMIT "+maxRows+" FOR UPDATE SKIP LOCKED))";
     }
 }


strehle:
@peterhaochen47 @bruce-ricard question is, if you know tests and/or situation where postgresl < 9.5 is in use ?

We have seen lock situations in DB monitoring and found this statement enhancement which prevents locks.
Each login with authz will have entries in oauth_code and revocable_tokens and in similar time we have delete operations, thus this can lead to locks

strehle:
found release notes
https://github.com/cloudfoundry/uaa/releases/tag/v74.13.0
-> for postgresql we have condition 9.5. and higher already, so this change can be done

torsten-sap (approved):

PR #1737, opened 2021-12-05
feature: add group mapping mode AS_SCOPES for OIDC IdPs

### LINKED ISSUE #1444: Mapping group to organization role
Hopefully just a quick question. I haven't been able to find any docs on the subject other than an old email thread from 2015 mentioning that support was on the way.

Is it possible to give external groups to users from OIDC/LDAP which grants them organization roles in CF? Or the ability to give a role in CF to a UAA group rather than per user?

Thank you.

### LINKED ISSUE #1735: Mapping groups to scopes using OAuth (OIDC)
75.0.0

As part of a commercial Cloud Foundry distribution

### What did you do?
I have UAA configured to connect to an external IdP via OAuth(OIDC); however, I can't get the groups from the IdP's Id token to map to Scopes in the UAA token. I noticed there was a 'groupMappingMode: AS_SCOPES' that could be used for SAML, but I didn't see the same option for OAuth/OIDC.

I tried creating an attributeMapping to map external_groups to the groups claim like this:
       attributeMappings:
             external_groups: groups
and from the logs this appears to work as far as getting/saving the groups for the user, however when the UAA token is created these groups still aren't included.

Can this be done, or am I off on the approach?
Thanks.

The groups from the external IdP to appear in the Scopes of the UAA token.

### What did you see instead?
Groups are not included in the UAA token.

strehle:
Add new config option, similar to SAML configuration
Add default null, so that non existing or null entries in JSON config mapped to
existing default explicit mapping.
Only if config option AS_SCOPES is set this is visible in JSON config
Added JUnit for verification of authenticated authorities in both cases


strehle:
Example entries for  uaa.yml:
```
        attributeMappings:
          email: email
          external_groups:
            - roles
        groupMappingMode: AS_SCOPES
```

PR #1750, opened 2021-12-17
Fix issue #1713

### LINKED ISSUE #1713: Missing index for username and identity_zone on users table (Postgres)

75.9.0

I am deploying the UAA
- locally only using gradlew

### What did you do?

Executed a SQL manually because we had a postgres issue in bigger landscapes

Example:
```
select count(*) from users where ((LOWER(username) = LOWER('marissa') and LOWER(identity_zone_id) = LOWER('uaa')))
```

You may execute explain analyze.

Expected is an indexed scan in the query plan to ensure that we run with the index.

### What did you see instead?

Sequenced scan is done because we have no lower index on idz and origin.

### Reason found in git log

Second query that way adapted in the following commit: https://github.com/cloudfoundry/uaa/commit/dca40d968c42a2b5e09b84639787f9517ecc68b8

torsten-sap:


server/src/main/resources/org/cloudfoundry/identity/uaa/db/postgresql/V4_101_1639764160__Add_LowerIndex_To_Users_Wo_Origin.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX concurrently IF NOT EXISTS users_key_lower_wo_origin ON users (LOWER(username),LOWER(identity_zone_id));
```
strehle: only one remark, all postgresql relavat commds are in upper case exception concurrently. so you can think about this but no functional change

strehle (approved):

strehle (approved):

Issue #1817, opened 2022-03-23
Missing encoding of RelayState for SAML SLO


robl-sap:
### What version of UAA are you running?

75.16.0 or develop

### How are you deploying the UAA?

I am deploying the UAA

- locally only using gradlew

### What did you do?

Added SAML IdP, logged in, then triggered SP initiated SLO.

### What did you expect to see? What goal are you trying to achieve with the UAA?

SLO contains encoded RelayState.

### What did you see instead?

RelayState is not encoded.

Naive approach could be to urlencode when writing here:
https://github.com/cloudfoundry/uaa/blob/ad4a24861853348b52627c03541bd8b50d655e51/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/RedirectSavingSamlContextProvider.java#L43

Receiver side already decodes:
https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/SamlRedirectLogoutHandler.java#L29-L36

PR #1818, opened 2022-03-23
Properly urlencode RelayState for SAML

robl-sap:
Potential fix for #1817


strehle:
@robl-sap FYI I think you need to fix / decode also the receiver part, see https://github.com/cloudfoundry/uaa/pull/1819

<edit>
I  debugged the SLO scenario with 2 parties and the RelayState parameter in GET is urlencoded already. If we do here an additional URLEncode, then we would have to URLDecode it on reciever side twice, however double URLEn/Decode is not needed, but I dont see an issue here

Issue #1838, opened 2022-04-14
Checkmarx findings


torsten-sap:
### What version of UAA are you running?

75.18.0

### How are you deploying the UAA?

I am deploying the UAA using cf-deployment.

### What did you do?

Checkmarx scan.

### What did you expect to see? What goal are you trying to achieve with the UAA?

No findings.

### What did you see instead?

The scan reported some findings.

PR #1864, opened 2022-04-29
remove String Passwords variables from 3 classes:

litesoft:
Issue #1838

All Tests pass!

Used different techniques:
AutologinRequestConverter: switched local variables to method references / lambdas.
ExternalLoginAuthenticationManager: changed "String password" to "Object credentials" then inlined the conversion to string to pass on.
KeystoneAuthenticationManager: changed "String password" to "char[] password" in super class: converted to/from String for getter

Dropped unneeded Setters (making variables final)


litesoft:
I have done another EasyCLA, so I'm not sure what I have failed to do...

litesoft:
Bad email in commit!

PR #1881, opened 2022-05-04
correct object type of user_attributes

jdachman:


strehle:
Thanks for the PR. I only can image for what reasons (id_token mapping in proxy case?) but need more context why you need/want to change this. Best with a test.

jdachman:
Correct, this is for adding `user_attributes` to the access_token in a custom token enhancer. The presence of `user_attributes` in the list of claims already suggests someone anticipated as much. Just the choice of a String is odd, especially since in the id_token `user_attributes` is a `Map<String,List<String>>` (see https://github.com/cloudfoundry/uaa/blob/f8188564facddee35600a23580eb845194337753/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java#L621-L624). This PR then just brings what is here in line with what is there.

strehle:
I checked the creation with user_attributes and id_token contains the entries if the user has the scope, see
https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/openid/IdTokenCreator.java#L131

The access_token is not designed to have user attributes. The standard code would not serialize the user_attributes in access_token if Claims properties are corrected and you suggest, so please provide a test that it gets clear what you change should correct.

strehle:
@bruce-ricard  any thoughts about this

jdachman:
@strehle Please explain why you closed this.

strehle:
Because there is no progress... and until now I dont understand the business context for this, there is no test which could explain it and therefore I closed it now

jdachman:
The use case is, as I mentioned, for adding `user_attributes` to the `access_token`. In any event as I pointed out there is an inconsistency in the app since elsewhere user_attributes is of type `Map<String,List<String>>`.

jdachman:
Or, to put it a different way the whole reason why `user_attributes` is here is because someone anticipated adding `user_attributes` to the `access_token` claims. Yet that is not currently possible to implement because the type is wrong. So either delete `user_attributes` altogether from here or correct the type.

PR #1888, opened 2022-05-13
[PKCE] allowpublic client option for grant_type authorization_code using PKCE with S256 method

### LINKED ISSUE #1839: Receiving auth token in Authorization Code Grant with PKCE
75.18.0

```
curl 'https://example.com/uaa/oauth/token' \
  -H 'Accept: */*' \
  -H 'Cache-Control: no-cache' \
  -H 'Content-Type: application/x-www-form-urlencoded' \
  --data-raw 'grant_type=authorization_code&code=111...222&redirect_uri=example.com&code_verifier=1234...5678&client_id=my.client' \
  --compressed
  ```

```
{"error":"unauthorized","error_description":"Bad credentials"}
```

### What did you do?
Some time ago PKCE suport was added to UAA. I wanted to switch from oAuth2 Implicit Grant flow to Authorization Code Grant with PKCE. I'm able to get the `code` using browser flow, however i keep getting 401 when trying to request the token. As it's a browser flow, of course i'm not sending `Authorization` header, or `client_secret` in the POST request to `/oauth/token`. 
- Question - Is there even a way to request auth token in browser using flow with PKCE, without sending the `client_secret`? 
  - i'd expect it to work similar to spotify's implementation https://developer.spotify.com/documentation/general/guides/authorization/code-flow/
- API documentation describes receiving code using browser flow but don't suggest what to do with the received code - this could be improved.

How to exchange code for token in the browser?

strehle:
Info: allowpublic is an optional flag similar to autoapprove to define behaviour in oauth2 flow.
The option allow to omit client_secret parameter and/or client authentication in grant_type authorization_code.
Escpecially mobile scenarios showed the need for such option, because other OpenID providers allow in meanwhile similar
use cases.

The needed configuration for the clients can be passed via
- yaml
- uaac, see https://github.com/cloudfoundry/cf-uaac/pull/113 

Clients with allowpublic = true can omit client authentication completly (authorization or client_secret) but the PKCE flow with code_challange_method=S256 is need in the authorization_code flow


strehle:
Code Quality Gate:
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=1888

PR #1892, opened 2022-05-18
Map HTTP 500 RequestRejectedException into 400

strehle:
Spring Security Background
https://stackoverflow.com/questions/54489645/spring-the-request-was-rejected-because-the-url-was-not-normalized-how-to-tel

Spring Security 6 changes this also to 400, e.g. 
https://github.com/spring-projects/spring-security/issues/7568

Request:

https://login.uaa-acceptance.cf-app.com/%2etable?type=%3Cmasked%3E&textdomain=%3Cmasked%3E&default-language&lang=%3Cmasked%3E

OLD -> HTTP 500
image

NEW -> HTTP 400
image


torsten-sap:
@strehle With regards to the error message "Request from internal firewall rejected". Is it possible to add some more details on the why for the end user?

strehle:
> Is it possible to add some more details on the why for the end user?

pleases make a proposal , the blocking reason is logged as warning right now. therefore here you have the reason, e.g. 
The request was rejected because the URL contained a potentially malicious String "%2e"

For the end user I would not show the blocking reason 

But if you want we can do and show the exception text from spring security escaped in the page

@bruce-ricard  any thoughts from VMware side ?

torsten-sap:
@strehle I propose the following as message:
"The request was rejected because it contained a potentially malicious character"

strehle:
image

Ok now ? @torsten-sap 

Tests green
Sonar: https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=1892

torsten-sap:
@tack-sap Can you review technically? Thanks!

tack-sap (approved):
LGTM

PR #1913, opened 2022-06-10
fix NPE when calling /saml_error directly

robl-sap:
We saw that some customers tend to directly call the /saml_error page instead of being redirected.

This results in an 500 error code because there is no Authentication Exception present and thus we see a NPE.

Thus, we suggest to check whether an exception is present and otherwise default to an empty string.


strehle:
Result.
Sonar green:
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=1913

IT failed
SamlLoginWithLocalIdpIT > testWebSSOProfileWithArtifactInMetadataSamlIntegration FAILED

The failed scenario is in SAML Idp functionality and on first look I dont know why it fails, but in this state I cannot merge it because if would break our CI and release build functionality

torsten-sap (commented):
Change is fine for me

strehle:
> Change is fine for me

ok, change is small and I am also fine with it, but it breaks the CI currenlty so we or you have to invest why and fix this

PR #1926, opened 2022-06-25
Use login.url for password reset email template creation

### LINKED ISSUE #1908: unable to specify base URL in the reset password link using uaa.yml
We have UAA running behind an nginx proxy, something like:

```
# docker-compse.yml

uaa-service:
  image: cloudfoundry/uaa:75.20.0
  volumes: [ "./uaa.yml:uaa.yml" ]
  
nginx:
  image: ...
  ports: [ "8080:8080" ]
```

```
# nginx.conf

upstream uaa {
  server uaa-service:8080;
}

location /uaa {
  proxy_pass http://uaa;
}
```

```
# uaa.yml

uaa:
  url: "http://foo.com/"
  
login:
  url: "http://foo.com/"
  
...
```

UAA comes up fine at http://localhost:8080/uaa and I'm able to create a user. 

The problem is that the emails UAA generate contain the wrong links. The reset password email has a link to:

http://uaa/uaa/reset_password?code=zzzz

(where `uaa` is the host header it's receiving from nginx -- and not the `foo.com` we expect as configured in the uaa.yml file). 

I'm confused because the [spec](https://github.com/cloudfoundry/uaa-release/blob/develop/jobs/uaa/spec) says that `uaa.url` and `login.url` will be used for email links... although it does not appear to be respecting these when rendering the templates, and instead relying on spring to determine host/port/proto ...

please advise on how we can configure these links used by emails.

strehle:
fix issue 1908

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/account/ResetPasswordController.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/account/ResetPasswordController.java
@@ -15,6 +15,7 @@ import org.cloudfoundry.identity.uaa.zone.MergedZoneBrandingInformation;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.http.HttpStatus;
 import org.springframework.security.core.userdetails.UsernameNotFoundException;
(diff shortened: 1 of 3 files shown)


strehle:
@bruce-ricard  any concerns. The bug was reported external and it does not affect us (SAP) because dont offer this feature, but others might run into the reported problem and thus this should solve it and the question is, does it is related to your usage, ? if yes please check if OK with you

PR #1930, opened 2022-07-06
Add psql index for groups table on columns id, idz

phschon:


strehle:
Hi @phschon , thanks for the PR. Can you please create an issue tickert and describe in more detail from where you step into this, so that I can check around if there are other missing indexes

strehle:
Hi @phschon ,
I spend some time to check your request.

I could not found a query where the lower is the problem
The query to get groups is like

`select %s from %s where id=? and identity_zone_id=?`

see, https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupProvisioning.java#L59

6 lines below you see a query with LOWER in and where the LOWER index makes sense and this was contributed some years ago already, see
https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/resources/org/cloudfoundry/identity/uaa/db/postgresql/V4_9_0__Add_Index_To_Groups_Name.sql

The point is, it worked obviously before but now you have seen an issue. However I think it is not related to LOWER but due to the fact, that partial indexes (and here we have a partial index because of id) work but sometimes a DB gets under preasure and it can happen, that the DB says, "I am sick today, need to rest ;)" 

To you issue.
I agree that we can create an index so that DBs get full index access, but then this index should be added to all DBs.

But I request from you.
Please re-create the index on your DB without LOWER and check some queries with explain, see
https://www.postgresqltutorial.com/postgresql-indexes/postgresql-partial-index/
or here problem with partial index
https://dba.stackexchange.com/questions/295448/postgres-failing-to-consistently-use-best-partial-indexes-when-multiple-are-pres

For now I have a problem to add an index which is obviously also not fullly correct ( because we have no LOWER select ) but I see the need to solve it with a full index search

server/src/main/resources/org/cloudfoundry/identity/uaa/db/postgresql/V4_101_1657107575__Add_Groups_Lower_id_idz.sql
```diff
@@ -0,0 +1 @@
+create index concurrently IF NOT EXISTS id_identityzone_lower ON groups (lower(id), lower(identity_zone_id));
```
strehle, line 1: Please see my comment about lower
https://github.com/cloudfoundry/uaa/pull/1930#issuecomment-1179536073

strehle (changes requested):
please check on DB if index with lower is not even a better solution

PR #1931, opened 2022-07-06
doc: update sample curl output

swalchemist:
Ran the curl command for /uaa/login and copied in the current output for the README.

Co-authored-by: Ramkumar Vengadakrishnan <ramkumarv@vmware.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,14 +101,8 @@ First run the UAA server as described above:
 From another terminal you can use curl to verify that UAA has started by
 requesting system information:
 
-    $ curl -H "Accept: application/json" localhost:8080/uaa/login
-    {
-      "timestamp":"2012-03-28T18:25:49+0100",
-      "commit_id":"111274e",
-      "prompts":{"username":["text","Username"],
-        "password":["password","Password"]
-      }
-    }
+    $ curl --silent --show-error --head localhost:8080/uaa/login | head -1
+    HTTP/1.1 200
 
 For complex requests it is more convenient to interact with UAA using
 `uaac`, the [UAA Command Line Client](https://github.com/cloudfoundry/cf-uaac).


README.md
```diff
@@ -102,13 +102,7 @@ From another terminal you can use curl to verify that UAA has started by
...
 
     $ curl -H "Accept: application/json" localhost:8080/uaa/login
-    {
-      "timestamp":"2012-03-28T18:25:49+0100",
-      "commit_id":"111274e",
-      "prompts":{"username":["text","Username"],
-        "password":["password","Password"]
-      }
-    }
+    {"app":{"version":"0.0.0"},"links":{"uaa":"http://localhost:8080/uaa","passwd":"/forgot_password","login":"http://localhost:8080/uaa","register":"/create_account"},"zone_name":"uaa","entityID":"cloudfoundry-saml-login","commit_id":"b257883","idpDefinitions":{},"prompts":{"username":["text","Email"],"password":["password","Password"]},"timestamp":"2022-06-29T03:58:05-0700"}
```
bruce-ricard: I like better how the previous one was formatted, it was easier to read with one key per line. Do you think you could pipe the output into `jq .` to make it look pretty?

bruce-ricard: We could update the command to be:
```
curl -H "Accept: application/json" localhost:8080/uaa/login | jq .
```

To be display exactly what the command outputs.

peterhaochen47 (approved):

README.md
swalchemist: @bruce-ricard That's an admirable readability improvement, but given that the purpose of the command is to show whether the server is running or not, I think the extra screen real estate would actually detract from this purpose. Do you like this better?

```
→ curl -sSH "Accept: application/json" localhost:8081/uaa/login > /dev/null && echo "RUNNING"
curl: (7) Failed to connect to localhost port 8081 after 7 ms: Connection refused

→ curl -sSH "Accept: application/json" localhost:8080/uaa/login > /dev/null && echo "RUNNING"
RUNNING
```

swalchemist: Cc: @ram-pivot.

swalchemist: Further improvement:
```
curl -sS localhost:8080/uaa/login > /dev/null && echo "RUNNING"
RUNNING
```

@bruce-ricard Looks ok?

swalchemist: Cc: @pabloarodas

bruce-ricard:
Interesting idea. 

I have a follow up idea:
to replace the `> dev/null` you can use `--head` to only fetch the headers, and then you can display the actual HTTP code with

```
curl --silent --head http://www.example.org
```

which outputs something like:
>HTTP/1.1 200 OK
Content-Encoding: gzip
Accept-Ranges: bytes
Age: 556901
Cache-Control: max-age=604800
Content-Type: text/html; charset=UTF-8
Date: Fri, 08 Jul 2022 22:11:45 GMT
Etag: "3147526947"
Expires: Fri, 15 Jul 2022 22:11:45 GMT
Last-Modified: Thu, 17 Oct 2019 07:18:26 GMT
Server: ECS (mic/9A9C)
X-Cache: HIT
Content-Length: 648


And if you want just the http code
```
curl --silent --head http://www.example.org | head -1
```

> HTTP/1.1 200 OK

does the trick.

Maybe that's better than `echo RUNNING` ?

README.md
swalchemist: Okay, adopted your suggestion below, also added --show-error so there's output in case of error.

swalchemist:
@bruce-ricard Okay, adopted your suggestion, also added --show-error so there's output in case of error.

PR #1936, opened 2022-07-12
feat: Added dependabot support for 74.5.x branch

ram-pivot:


peterhaochen47 (approved):
LGTM.
Will merge without all the GitHub action checks complete because this is not a src code change.

PR #1947, opened 2022-07-14
feat: Port over changes in release v74.5.26 into OSS UAA

pabloarodas:
fix: SPRING_SESSION_ATTRIBUTES.ATTRIBUTE_BYTES column in MySQL is not large enough

- Backport the OSS change that makes the column type from blob (64kb) to mediumblob (16mb)
- Add test for Spring Session Attributes table

[#178915188][PKS-1096]
[#182723091] Open sourcing UAA tracker story


peterhaochen47:
@pabloarodas This port is only the test file. I assume this is because the real code change is on develop already?

peterhaochen47 (approved):

PR #1966, opened 2022-07-26
Feat/port 74.5.35

pabloarodas:
* Add validation to corsPolicy's allowed URIs when creating/updating identity zones #179727364
* fix: flyway mysql migration SQL script failure when using utf8mb4 character set #181291079

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/zone/GeneralIdentityZoneConfigurationValidator.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/zone/GeneralIdentityZoneConfigurationValidator.java
@@ -91,6 +91,7 @@ public class GeneralIdentityZoneConfigurationValidator implements IdentityZoneCo
         return config;
     }
 
+    @SuppressWarnings("javasecurity:S2631")
     private void validateRegexStrings(List<String> uris, String fieldName) throws InvalidIdentityZoneConfigurationException {
         if (uris != null) {
             for (String uri : uris) {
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/cloudfoundry/identity/uaa/zone/GeneralIdentityZoneConfigurationValidator.java
```diff
@@ -83,6 +91,18 @@ public IdentityZoneConfiguration validate(IdentityZone zone, IdentityZoneValidat
         return config;
     }
 
+    private void validateRegexStrings(List<String> uris, String fieldName) throws InvalidIdentityZoneConfigurationException {
+        if (uris != null) {
+            for (String uri : uris) {
+                try {
+                    Pattern.compile(uri);
```
strehle, line 99: sonar suggests that you do Pattern.compile(Pattern.quote(uri));

https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&types=VULNERABILITY&pullRequest=1966&id=cloudfoundry-identity-parent&open=AYI7Se-6BHqZnT1RNIiH

pabloarodas: We tried doing that change but tests started failing after that. Since this is complaining about a value that is not a user input, what we did was ignore the specific validation for this method.

swalchemist: Yeah, the tests probably require that these strings actually work with regex metacharacters, which is what Sonar is complaining about.

swalchemist (approved):

PR #1968, opened 2022-07-26
Fix postgres migration test failure

strehle:
see former commit
https://github.com/cloudfoundry/uaa/commit/480760e909aab6de77e4bb41da5401aa34e87565


ram-pivot (approved):
Thanks. Looks good to me.

PR #1971, opened 2022-07-28
fix sonar findings

strehle:


strehle:
no logic changed, but followed local sonar lint plugin recommendations to solve
image

Issue #1983, opened 2022-08-03
Update of last_logon_success_time consuming DB resources


tack-sap:
### What version of UAA are you running?

v75.21.0

### How are you deploying the UAA?

The UAA (uaa-release) is deployed via bosh as part of the complete cf landscape installation. We are running the UAA against a Postgresql Database.

### What did you do?

We saw that some users are requesting tokens for the same user in parallel in multiple threads. This is probably done in some automated job (not by us, but someone using our UAA). 

### What did you expect to see? What goal are you trying to achieve with the UAA?

We were expecting that the UAA would be able to handle the parallel requests without any problems.

### What did you see instead?

We saw that the UAA is having problems to update the last_logon_success_time for the parallel requests. While the UAA logs show no problems here as eventually the update is done, we see in the DB insights, that some of the update commands to update the timestamp take several seconds. Most of this time is used to wait for other db processes that are accessing the same row at the same time. If the requests are kept at a constant rate this will lead to more and more DB resources consumed by this usually trivial update command. We actually got alerts that the DB was having problems to accept new connections because of such a situation lately.

In our own UAA-based service we saw this problem already some time ago and resolved this by changing the update command for the logon time to a version where we would try to get a lock for the row without waiting during the update. With this new statement the problem disappeared. Therefore we want to contribute this updated statement to the UAA so we won't see this in the future in the UAA.

As we only saw the problem in our postgresql based deployment and don't know about the behavior of the other DBMS in such a situation, the PR will only target Postgresql. The Syntax used in the new statement is Postgresql specific anyways, we need to differentiate between different DBMS anyways. If similar problems would appear with other vendors, it would be easy to implement similar mechanism for those, using their specific SQL dialect.

strehle:
I see you mention a PR and in this fix and here you mention only Postgresql
So do you known that SKIP Locked is support from MYSQL too ?
https://dev.mysql.com/blog-archive/mysql-8-0-1-using-skip-locked-and-nowait-to-handle-hot-rows/

To me such an extension to have skip locked is fine, but please improve here UAA in general not only the postgresql case

PR #1984, opened 2022-08-03
prevent lock situation during update of last_logon_success_time on postgres

tack-sap:
PR for #1983 

As mentioned in the Issue description, this is how we resolved the lock situation in our own UAA-based service.

In a normal situation the behavior of this code is identical. Only in case the lock cannot be granted during the update statement, the update would be skipped. But this would mean another process is currently trying an update on the same row, maybe with a few milliseconds difference in last logon time, but still there is an update. As we still have all the audit events of all those parallel requests, we concluded that the stability of the UAA should be more important than keeping the last logon time redundantly in the DB for every user.

Like announced in the issue description this is only changing the behavior for postgresql, as this is where we saw the problem and this is also where we already were able to validate that the fix is working and not breaking stuff.

As the behavior from the outside is the same in a normal use case, the current ITs are already covering the new code (when executed against postgres).


tack-sap:
Failing IT seems to be unrelated to my changes, as I see the same errors locally against a clean develop branch now

strehle (commented):
in past similar vendor sql related problems were solved with LimitSqlAdapter, so why not rename this to something like VendorSqlAdapter and add the update SQL into this interface next to limit SQLs 

Sonar should be green before merge
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=1984

strehle:
@torsten-sap FYI please check if you agree on a solution where we use an interface class instead of a custom SQL inside the call

You want stay with our code, ok but Sonar needs to be Green that is what we ensure before merge

torsten-sap:
@tack-sap, @strehle and myself decided to go for PR https://github.com/cloudfoundry/uaa/pull/1993 to solve issue https://github.com/cloudfoundry/uaa/issues/1983. Thus, closing this PR.

PR #1986, opened 2022-08-03
Feat port 74.5.33

pabloarodas:
fix: Improve efficiency of SimpleSearchQueryConverter's parameter prefix generation
- Move AlphanumericRandomValueStringGenerator from test to main
- Use AlphanumericRandomValueStringGenerator when SimpleSearchQueryConverter generates parameter prefix instead of regenerating the string until it does not contain "-" using RandomValueStringGenerator

[#179824012]


hsinn0 (approved):
Looks good.  Approving pending the failed CI check to pass.

server/src/main/java/org/cloudfoundry/identity/uaa/util/AlphanumericRandomValueStringGenerator.java
```diff
@@ -0,0 +1,44 @@
+package org.cloudfoundry.identity.uaa.util;
+
+import java.security.SecureRandom;
+import java.util.Random;
+
+public class AlphanumericRandomValueStringGenerator {
```
hsinn0, line 6: In 74.5.x, `AlphanumericRandomValueStringGenerator` extends `org.springframework.security.oauth2.common.util.RandomValueStringGenerator`.  Any specific reason not doing the same?

ram-pivot: @hsinn0 Sonar gave warnings to not extend a class that is Deprecated and Sonar checks failed.

hsinn0: I see.  In fact, we know that the whole `spring-security-oauth2` is deprecated and have a story to replace with different one.  Because the whole UAA project has dependency to the EOL-ed `spring-security-oauth2` package, it practically does not really matter for this one class to depend on a deprecated class.
Anyhow, this change will do the job as it is, so I am fine with it then.

ram-pivot: Yes agree. Since we modified this file, sonar was scanning and complained about this specific class usage.

PR #1990, opened 2022-08-10
Bump mariadb version 2.7.6

strehle:
I think you reverted back to 2.7.4 because in spring boot 2.7.x the version is pinned to 3.x
So I am fine with 2.7.x , but I think we should then define  latest version there ?

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -4,7 +4,7 @@ ext {
 }
 
 // Versions we're overriding from the Spring Boot Bom
-ext["mariadb.version"] = "2.7.4"
+ext["mariadb.version"] = "2.7.6"
 ext["flyway.version"] = "7.15.0"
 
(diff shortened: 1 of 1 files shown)


ram-pivot (approved):
Looks good to me.

ram-pivot:
> I think you reverted back to 2.7.4 because in spring boot 2.7.x the version is pinned to 3.x So I am fine with 2.7.x , but I think we should then define latest version there ?

We ran into issues when using 3.x version of mariadb and had to revert back to using 2.7.4. Hopefully using 2.7.x latest should be fine.  Thanks for the PR.

strehle:
> Hopefully using 2.7.x latest should be fine.

I dont know if you need another checks. The PR passed the pipeline, so can we continue with this version? if you face issues with your tests, please come back, then I ll revert

PR #1993, opened 2022-08-11
fix https://github.com/cloudfoundry/uaa/issues/1983

### LINKED ISSUE #1983: Update of last_logon_success_time consuming DB resources

strehle:
alternative fix for issue #1983 

Because of existing paramter maxSqlParameters and https://github.com/cloudfoundry/uaa/pull/1143
The meaning of the parameter was mainly for postgresql but in implementation there was no reference to a DB , same done in this PR

Sonar is green because with the property we can test of switching the behaviour, similar to maxSqlParameter
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=1993

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/user/JdbcUaaUserDatabase.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/user/JdbcUaaUserDatabase.java
@@ -20,6 +20,7 @@ import org.springframework.security.core.authority.AuthorityUtils;
 import org.springframework.security.core.userdetails.UsernameNotFoundException;
 import org.springframework.util.StringUtils;
 
+import javax.annotation.PostConstruct;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


torsten-sap (approved):

torsten-sap:
@tack-sap, @strehle and myself decided to go for this PR to solve issue https://github.com/cloudfoundry/uaa/issues/1983. Thus, approving it.

PR #2024, opened 2022-09-14
docs: Added MFA deprecated notice and updated support text

ram-pivot:


peterhaochen47 (approved):
@ram-pivot  LGTM. Please add it to the bosh release spec file as well: https://github.com/cloudfoundry/uaa-release/blob/57a364fdfa07db5cf2f11d714008019dd1f856e0/jobs/uaa/spec#L956

strehle (approved):

PR #2034, opened 2022-09-21
fix issue with broken copy button

### LINKED ISSUE #2031: bosh login - copy button is broke and getting authentication failure with UAA

Version: 74.5.4

- concourse pipeline

### What did you do?

When using `bosh login` and getting redirected to the cloud foundry site to get the one time passcode, we get the following when we click the copy button:
[copy button.pdf](https://github.com/cloudfoundry/uaa/files/9611327/copy.button.pdf)

```
15:25 $ bosh login
Using environment '<REDACTED>'

Email ():
Password ():
Temporary Authentication Code ( Get one at https://<REDACTED>:8443/passcode ) ():

Failed to authenticate with UAA

Email ():
Asking for text:
  EOF

Exit code 1
```

After looking at UAA logs, restarting UAA processes via `monit`, my teammate selected the 1 time passcode and copied it / pasted it into the terminal and things work fine.  The copy button is broke

We tried:
- Chrome (regular & incognito)
- Safari
- Firefox

I also checked with our InfoSec team to see if they deployed any kind of new network policies, they did not.

I expected the copy button to copy, and paste into the terminal when I paste.  The copy button does not copy at all.

### What did you see instead?

```
Failed to authenticate with UAA

Email ():
Asking for text:
  EOF

Exit code 1
```

Here is some info from the browser dev console:

> Refused to execute inline event handler because it violates the following Content Security Policy directive: "script-src 'self'". Either the 'unsafe-inline' keyword, a hash ('sha256-...'), or a nonce ('nonce-...') is required to enable inline execution. Note that hashes do not apply to event handlers, style attributes and javascript: navigations unless the 'unsafe-hashes' keyword is present.

### Workaround

We do have a workaround in place, however it would be nice if the copy button working again.

strehle:
extract clip board java script into file
add event handler , which is described in many CSP guidelines

### DIFF
--- a/server/src/main/resources/templates/web/passcode.html
+++ b/server/src/main/resources/templates/web/passcode.html
@@ -4,24 +4,14 @@
       layout:decorate="~{layouts/main}">
 <head>
     <th:block layout:insert="~{nav :: head}"></th:block>
+    <script type="application/javascript" src="resources/javascripts/passcode/copy_clipboard.js"></script>
 </head>
 <div layout:fragment="page-nav">
     <th:block layout:replace="~{nav :: nav}"></th:block>
 </div>
 <div class="island" layout:fragment="page-content">
-    <script type="text/javascript" th:inline="javascript">
-        function CopyToClipboard(id)
-        {
-            var r = document.createRange();
-            r.selectNode(document.getElementById(id));
-            window.getSelection().removeAllRanges();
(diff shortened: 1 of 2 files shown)


peterhaochen47 (approved):
LGTM. Okay to merge given the urgency of this bug. We could backfill test later.

PR #2051, opened 2022-10-05
Add test for copy-to-clipboard functionality in `passcode` page

hsinn0:
[#183422195]

### DIFF
--- a/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/LoginIT.java
+++ b/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/LoginIT.java
@@ -269,6 +269,13 @@ public class LoginIT {
         attemptLogin(testAccounts.getUserName(), testAccounts.getPassword());
 
         assertThat(webDriver.findElement(By.cssSelector("h1")).getText(), Matchers.containsString("Temporary Authentication Code"));
+
+        // Verify that the CopyToClipboard function can be executed
+        String passcode = webDriver.findElement(By.id("passcode")).getText();
+        ((JavascriptExecutor)webDriver).executeScript("CopyToClipboard",
+                passcode);
+        // Verify that the copybutton can be clicked
+        webDriver.findElement(By.id("copybutton")).click();
     }
 
     @Test


strehle (approved):
Thanks

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/LoginIT.java
```diff
@@ -269,6 +269,13 @@ public void testPasscodeRedirect() {
         attemptLogin(testAccounts.getUserName(), testAccounts.getPassword());
 
         assertThat(webDriver.findElement(By.cssSelector("h1")).getText(), Matchers.containsString("Temporary Authentication Code"));
+
+        // Verify that the CopyToClipboard function can be executed
+        String passcode = webDriver.findElement(By.id("passcode")).getText();
+        ((JavascriptExecutor)webDriver).executeScript("CopyToClipboard",
+                passcode);
+        // Verify that the copybutton can be clicked
+        webDriver.findElement(By.id("copybutton")).click();
```
bruce-ricard, line 278: Not knowing what the semantics of the `click()` method is, from its name it looks to me that it just clicks on the element. But you can always click on anything, even if it's not per say clickable. It seems to me like what we should be testing is that the clicking actually has some impact. Maybe could we test that passcode is indeed in the clipboard?

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/LoginIT.java
```diff
@@ -269,6 +269,13 @@ public void testPasscodeRedirect() {
         attemptLogin(testAccounts.getUserName(), testAccounts.getPassword());
 
         assertThat(webDriver.findElement(By.cssSelector("h1")).getText(), Matchers.containsString("Temporary Authentication Code"));
+
+        // Verify that the CopyToClipboard function can be executed
```
bruce-ricard, line 273: It seems to me like this test is mixing 2 different layers of abstraction, by testing the javascript method, and then testing the UI.

This is an integration test, it seems to me like we could just test that the functionality is working, when we click on the button.

bruce-ricard (changes requested):

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/LoginIT.java
hsinn0: The button click will copy the passcode to the OS clipboard.  Not something I want to check in Java code.  I also checked if I can check for some kind of event, but was not feasible without changing the involved javascript function.
This case, at least, will test the existence of the 'copybutton'.

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/LoginIT.java
hsinn0: By testing both the button's existence and it is clickable, and also the Java function is loaded and can be executed, we at least test both components involved in the action.  I do understand that we miss the part verifying that the function is being executed when you click the button.

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/LoginIT.java
bruce-ricard: It looks like it's fairly easy to get the contents of the clipboard: https://stackoverflow.com/a/7106086/1296753

How would you feel about testing the contents of the clipboard just after the click, to validate that it contains the passcode?

PR #2054, opened 2022-10-08
Update UAA server debug mode run options

hsinn0:
- Use `-agentlib` option instead of `-Xrunjdwp`
- Add gradle run option to suspend the server till the debugger is attached
- Update the related section in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,21 @@ To load JDWP agent for UAA jvm debugging, start the server as follows:
 ```sh
 ./gradlew run -Dxdebug=true
 ```
+or
+```sh
+./gradlew -Dspring.profiles.active=default,hsqldb,debug run
+```
 You can then attach your debugger to port 5005 of the jvm process.
+
+To suspend the server start-up until the debugger is attached (useful for
+debugging start-up code), start the server as follows:
+```sh
+./gradlew run -Dxdebugs=true
+```
+or
+```sh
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -117,8 +117,22 @@ To load JDWP agent for UAA jvm debugging, start the server as follows:
 ```sh
```
swalchemist, line 117: How about removing the old option so we can simplify the document?

README.md
```diff
@@ -117,8 +117,22 @@ To load JDWP agent for UAA jvm debugging, start the server as follows:
...
 ```
+or
+```sh
+./gradlew -Dspring.profiles.active=default,hsqldb,debug run
+```
 You can then attach your debugger to port 5005 of the jvm process.
 
+To suspend the server start-up until the debugger is attached (useful for
+debugging start-up code), start the server as follows:
+```sh
```
swalchemist, line 128: Just need one option here, too, if they both do the same thing.

swalchemist (changes requested):

README.md
hsinn0: Old option meaning using `spring.profiles.active` property? I want to keep it so one can just use spring profile with all the desired options instead.

README.md
hsinn0: The same. I want to keep both options so one can also use the spring profile with all the desired options instead.

PR #2071, opened 2022-10-23
Refactor UaaTokenServices

strehle:
remove bean setter and use final variables everywhere it was easy to change
other places change the scope of the method to proteced

[sonar recommendations](https://plugins.jetbrains.com/plugin/7973-sonarlint)

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java
@@ -73,7 +73,6 @@ import org.springframework.util.MultiValueMap;
 import org.springframework.util.StringUtils;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java
```diff
@@ -852,9 +842,8 @@ public OAuth2AccessToken readAccessToken(String accessToken) {
         // Expiry is verified by check_token
         CompositeToken token = new CompositeToken(accessToken);
         token.setTokenType(OAuth2AccessToken.BEARER_TYPE);
-        token.setExpiration(new Date(Long.valueOf(claims.get(EXPIRY_IN_SECONDS).toString()) * 1000L));
+        token.setExpiration(new Date(Long.valueOf(claims.get(EXPIRY_IN_SECONDS).toString()) * ONE_SECOND));
```
swalchemist: It's great to get rid of the magic number. Maybe a more descriptive name would be `MILLIS_PER_SECOND`. Otherwise it looks like we're multiplying by 1.

swalchemist (changes requested):
Thanks for continuing the refactor!

It would have been nice to have smaller commits with more descriptive commit messages, but I was able to get the idea.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java
strehle: thanks, done

swalchemist (approved):
Sorry it took me a few days!

Issue #1698, opened 2021-10-15
UAA authorization codes still too short


meeque:
This is a follow-up to PR #1615. The fix there is incomplete, UAA still uses authorization codes that are only **120 bits** strong. That is better than before the PR (which was 60 bits, though erroneously reported as 80 bits), but still [less than suggested by the oAuth2 spec](https://datatracker.ietf.org/doc/html/rfc6749#section-10.10). The standard mandates **128 bits** of randomness and suggests **160 bits** of randomness.

I'm not reporting this to security@cloudfoundry.org, because imho the security risk is fairly low. Afaik UAA authorization codes are short-lived, single-use, and require an additional client secret to obtain an access token. So 120 bits might just be enough here. Still why not bump it up to **160 bits** and be on the safe side? Feel free to reject this issue, if you disagree.

### What version of UAA are you running?

Unknown. Sorry, couldn't find that info endpoint.

### How are you deploying the UAA?

I am **not** deploying UAA. I'm just a user.

### What did you do?

I signed in with the authorization code grant. I observed the network requests in my browser's developer tools.

### What did you expect to see? What goal are you trying to achieve with the UAA?

I expected to see UAA pass an authorization code based on **160 bits** of randomness to the application callback.

### What did you see instead?

I observed an authorization code of **20 characters** `length`. That was 20 url-safe, printable characters, mostly alpha-numeric, plus 2 punctuation, 64 possible characters altogether. Such code can only hold up to **120 bits** of randomness:
```
64^20 = (2^6)^20 = 2^(6*20) = 2^120
```

Here is an example authorization code:
```
IgXbFDZ6aV4LVmdbyBeR
```

### Root cause:

After PR #1615, UAA uses Spring's [RandomValueStringGenerator](https://github.com/spring-projects/spring-security-oauth/blob/main/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/RandomValueStringGenerator.java) with `length` 20. However, that `length` specifies the length of the resulting strings, not the number of random bytes encoded in those strings.

[...]

PR #2072, opened 2022-10-23
increase randomness of passcode

strehle:
addresses #1698


peterhaochen47:
LGTM, better if you could squash the commits since first commit breaks tests.

PR #2101, opened 2022-11-16
Fix error "rawPassword cannot be null" and prevent null in password encoder

### LINKED ISSUE #2100: UAA cannot start with multiple secrets in cf client

76.3.0, but with develop same issue

- using cf-deployment

### What did you do?

setup  in cf the well known empty secret but do this after the cf client has multiple secrets already.

UAA cannot start because of stack

Caused by: java.lang.IllegalArgumentException: rawPassword cannot be null
at org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder.matches(BCryptPasswordEncoder.java:123) ~[spring-security-crypto-5.7.4.jar:5.7.4]
at org.cloudfoundry.identity.uaa.util.beans.BackwardsCompatibleDelegatingPasswordEncoder.matches(BackwardsCompatibleDelegatingPasswordEncoder.java:30) ~[cloudfoundry-identity-server-76.3.0.jar:?]
at org.cloudfoundry.identity.uaa.client.ClientAdminBootstrap.updatePasswordsIfChanged(ClientAdminBootstrap.java:275) ~[cloudfoundry-identity-server-76.3.0.jar:?]
at org.cloudfoundry.identity.uaa.client.ClientAdminBootstrap.addNewClients(ClientAdminBootstrap.java:221) ~[cloudfoundry-identity-server-76.3.0.jar:?]
at org.cloudfoundry.identity.uaa.client.ClientAdminBootstrap.afterPropertiesSet(ClientAdminBootstrap.java:100) ~[cloudfoundry-identity-server-76.3.0.jar:?]

### What did you see instead?

strehle:
Issue is, that a client update in UAA start (bootstrap) overwrites the secrets from YAML or oauth-clients.xml Especially for cf client.
If the cf client has 2 secrets in DB already, then the update with only one secret fails with used by: java.lang.IllegalArgumentException: rawPassword cannot be null

Prevent this in BackwardsCompatibleDelegatingPasswordEncoder

### DIFF
--- a/server/src/test/java/org/cloudfoundry/identity/uaa/client/ClientAdminBootstrapTests.java
+++ b/server/src/test/java/org/cloudfoundry/identity/uaa/client/ClientAdminBootstrapTests.java
@@ -562,7 +562,7 @@ class ClientAdminBootstrapTests {
         );
     }
 
-    protected static ClientDetails doSimpleTest(
+    static ClientDetails doSimpleTest(
             final Map<String, Object> map,
             final ClientAdminBootstrap clientAdminBootstrap,
             final MultitenantJdbcClientDetailsService clientRegistrationService,
@@ -605,7 +605,7 @@ class ClientAdminBootstrapTests {
         return created;
     }
 
-    protected static void assertSet(
+    static void assertSet(
             final String expectedValue,
             final Collection defaultValueIfNull,
             final Collection actualValue,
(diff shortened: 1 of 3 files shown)


strehle:
sonar 
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2101

server/src/test/java/org/cloudfoundry/identity/uaa/client/ClientAdminBootstrapTests.java
```diff
@@ -562,7 +562,7 @@ void clientWithoutGrantTypeFails() {
         );
     }
 
-    private static ClientDetails doSimpleTest(
+    protected static ClientDetails doSimpleTest(
```
hsinn0: It appears that the access modifier of this method and the other two methods are changed to 'protected' so they can be accessed from 'ClientAdminBootstrapMultipleSecretsUpdateTests' class, which is in the same package.  In that case, since they only need  package level access, changing them to have no access modifier would be better.

hsinn0 (approved):

PR #2106, opened 2022-11-22
Feature: Add private_key_jwt for OIDC Identity Providers

### LINKED ISSUE #1933: Support private_key_jwt for OIDC Proxy mode

develop version / 75.21.0

I am deploying the UAA

- locally only using gradlew
- using cf-deployment

### What did you do?

Configured an external OAuth2 / OpenID connect provider to proxy the authentication through UAA to external/corporate IdP
If I do this, I have set the external secret into DB but this is not encrypted but even with enccryption there is the problem of key rotation.

So the issue is, because UAA does not support private_key_jwt  method from OIDC for password grant. For authorization_code grant UAA would support PKCE with PUBLIC client already, so for this grant type UAA would not need the credential , but for password grant it is still needed anf this private_key_jwt  should be implemented.

This method is support by corporates already, e.g. https://docs.microsoft.com/en-us/power-apps/maker/portals/configure/configure-openid-provider

No secrets from external OIDC providers in DB table

### What did you see instead?

Plain secret of external OIDC provider in DB table

strehle:
Goal: >>> Allow to omit relyingPartySecret setting <<<

Support private_key_jwt for OIDC servers which provide this, e.g. okta, connect2d, SAP IAM.
[Microsoft Azure AD is not working yet, because of x5t claim. UAA needs to understand and work with X509 for JWT.](https://github.com/cloudfoundry/uaa/issues/2109)

Currently suport OIDC and OAuth2 standards.
Allow to omit relyingPartySecret setting but activate private_key_jwt with jwtClientAuthentication configuration.

If empty, then following OIDC rules

https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication if defined, allow to have own iss and aud claims 

as they are described in RFC 7523.


strehle:
sonar

https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2106

strehle:
@torsten-sap tested this with https://help.sap.com/docs/IDENTITY_AUTHENTICATION/6d6d63354d1242d185ab4830fc04feb1/1bdc7291de0e4e529648bfe05856029a.html?locale=en-US

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/jwt/JwtClientAuthentication.java
```diff
@@ -0,0 +1,62 @@
...
+  public static final String GRANT_TYPE = "urn:ietf:params:oauth:client-assertion-type:jwt-bearer";
+
+  private final KeyInfoService keyInfoService;
+
+  public JwtClientAuthentication(
+      KeyInfoService keyInfoService) {
+    this.keyInfoService = keyInfoService;
+  }
+
+  public String getClientAssetion(OIDCIdentityProviderDefinition config) {
```
swalchemist: Typo - seems that "Assertion" is misspelled in the method name.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/jwt/JwtClientAuthentication.java
```diff
@@ -0,0 +1,62 @@
+package org.cloudfoundry.identity.uaa.oauth.jwt;
```
swalchemist, line 1: Why no unit test coverage for this class from any of the tests in the server module?

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/OauthIDPWrapperFactoryBean.java
```diff
@@ -61,6 +61,16 @@ else if(OIDC10.equalsIgnoreCase(type)) {
                         oidcIdentityProviderDefinition.setPasswordGrantEnabled(idpDefinitionMap.get("passwordGrantEnabled") == null ? false : (boolean) idpDefinitionMap.get("passwordGrantEnabled"));
                         oidcIdentityProviderDefinition.setSetForwardHeader(idpDefinitionMap.get("setForwardHeader") == null ? false : (boolean) idpDefinitionMap.get("passwordGrantEnabled"));
                         oidcIdentityProviderDefinition.setPrompts((List<Prompt>) idpDefinitionMap.get("prompts"));
+                        if (idpDefinitionMap.get("jwtclientAuthentication") != null) {
```
swalchemist: The logic in this method is already too deeply nested. Please extract some methods to simplify the code.

server/src/test/java/org/cloudfoundry/identity/uaa/authentication/manager/PasswordGrantAuthenticationManagerTest.java
```diff
@@ -354,7 +360,47 @@ void testOIDCPasswordGrantProviderNoRelyingPartyCredentials() {
             instance.authenticate(auth);
             fail();
         } catch (ProviderConfigurationException e) {
-            assertEquals("External OpenID Connect provider configuration is missing relyingPartyId or relyingPartySecret.", e.getMessage());
+            assertEquals("External OpenID Connect provider configuration is missing relyingPartyId.", e.getMessage());
+        }
+    }
+
+    @Test
+    void testOIDCPasswordGrantProviderJwtClientCredentials() throws MalformedURLException {
```
swalchemist: Any way to simplify the rather large setup for this test?

swalchemist (changes requested):

server/src/test/java/org/cloudfoundry/identity/uaa/authentication/manager/PasswordGrantAuthenticationManagerTest.java
strehle: refactored this. other tests could also now re-use the mock methods, but in all other tests similar setup is done always from scratch

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/jwt/JwtClientAuthentication.java
strehle: created test

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/OauthIDPWrapperFactoryBean.java
strehle: done

strehle:
@swalchemist Thanks for your review. The label was set in order I wanted to have a review from @torsten-sap  or @tack-sap because this PR solves an issue we have with our setup

During refactoring I fould a bug, but now it should work as expected

Sonar:
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2106

Expectation is:
We use OIDC provider in our setup of UAA (SAP IAM / Azure ) but we need to get rid of the secrets in the configuration.
OIDC defines private_key_jwt mechanism as replacement for secrets.

strehle:
Hi @swalchemist do have any further concerns here ?

swalchemist (approved):

torsten-sap (approved):

Issue #2081, opened 2022-10-29
LimiterExpirationProcess error repeated infinitely when a UAA server goes idle


swalchemist:
### What version of UAA are you running?

develop branch, commit f96b007af

### How are you deploying the UAA?

- locally only using gradlew using the default config files

### What did you do?

When I have the uaa server running on my Macbook and leave the computer idle for an hour or more, several times I've returned to find this error spewing in the uaa log several times per second: `ERROR --- ExpirationBuckets: Second requested outside of current BucketRing window`. Once it starts, it seems to continue indefinitely as long as the server runs. It doesn't seem to have a noticeable effect on performance or functionality, but it does clog the log a great deal.

I've seen this happen three times so far. It seems to happen every time the laptop goes to sleep for more than a few minutes. The last time, however, I'm not sure if the laptop went to sleep at all. Here's a log snippet from the onset of the error, apparently not triggered by a hit to the back end:

```
[2022-10-28T19:36:18.339237Z] uaa - 94573 [http-nio-8080-exec-6] ....  INFO --- RateLimitingFilter: path '/uaa/login.do' (84000ns) -- NOT limited
[2022-10-28T19:36:18.349408Z] uaa - 94573 [http-nio-8080-exec-6] ....  INFO --- SamlKeyManagerFactory: Loaded service provider certificate legacy-saml-key
[2022-10-28T19:36:18.351051Z] uaa - 94573 [http-nio-8080-exec-6] ....  INFO --- NonSnarlMetadataManager: Initialized local service provider for entityID: cloudfoundry-saml-login
[2022-10-28T19:36:18.445025Z] uaa - 94573 [http-nio-8080-exec-6] ....  INFO --- Audit: IdentityProviderAuthenticationSuccess ('marissa'): principal=67a4ec51-7bcc-438c-9d5d-90512a841e9b, origin=[remoteAddress=0:0:0:0:0:0:0:1, sessionId=<SESSION>], identityZoneId=[uaa], authenticationType=[uaa]
[2022-10-28T19:36:18.446906Z] uaa - 94573 [http-nio-8080-exec-6] ....  INFO --- Audit: UserAuthenticationSuccess ('marissa'): principal=67a4ec51-7bcc-438c-9d5d-90512a841e9b, origin=[remoteAddress=0:0:0:0:0:0:0:1, sessionId=<SESSION>], identityZoneId=[uaa]
[...]

PR #2112, opened 2022-11-25
Rate limit: protected status page

strehle:
- with that needed to change IT
- enhanced IT block check to have the numbers in test
- protection right now is high: admin only
- uaac curl /RateLimitingStatus


swalchemist (approved):
Looks good. FYI, in the browser when someone either cancels the basic auth dialog or enters non-admin client credentials, raw XML is returned which renders oddly:

`An Authentication object was not found in the SecurityContextunauthorized`

or

`Access is deniedaccess_denied`

Probably not something to worry about.

swalchemist:
I reproduced #2081 while running the code in this branch. I presume that's not really related to the changes in the branch?

strehle:
I have rebase this branch so 
https://github.com/cloudfoundry/uaa/issues/2081 
should be fixed now, @swalchemist  please check again to ensure it, thanks.

the error is same as for all protected endpoints in UAA, e.g.
http://localhost:8080/uaa/oauth/token or $base/userinfo , etc.

tack-sap (approved):

PR #2120, opened 2022-11-30
Revert "Create sonarcloudpr.yml"

strehle:
Reverts cloudfoundry/uaa#2117


strehle:
was a test only but thats it, therefore revert

PR #2125, opened 2022-12-03
No response for non-supported HTTP methods

strehle:
But return simply 405

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/home/HomeController.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/home/HomeController.java
@@ -8,12 +8,15 @@ import javax.servlet.RequestDispatcher;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import java.io.IOException;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 import java.security.Principal;
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


hsinn0 (commented):
It appears that we can accomplish the same, i.e. not including the html body in the response by just setting DispatcherServlet's dispatchTraceRequest property to false.  It can be done by removing the following part from the servlet element in webapp/WEB-INF/web.xml as its default value is false:

```
        <init-param>
            <param-name>dispatchTraceRequest</param-name>
            <param-value>true</param-value>
        </init-param>
```
With my local UAA server, I tried the following:
```
% curl 'login.uaa-acceptance.cf-app.com/info' -i -X TRACE
```
And got responses as follows.
- with this PR change
  ```
  HTTP/1.1 405
  Transfer-Encoding: chunked
  Date: Thu, 08 Dec 2022 07:07:03 GMT
  ```
- with the web.xml change
  ```
  HTTP/1.1 405
  Allow: HEAD, DELETE, POST, GET, OPTIONS, PUT
  Content-Type: message/http
  Content-Length: 87
  Date: Thu, 08 Dec 2022 07:26:49 GMT

  TRACE /uaa/error HTTP/1.1
  host: localhost:8080
  user-agent: curl/7.79.1
  accept: */*
  ```
IMHO, the response with the web.xml change looks better.

strehle:
Created https://github.com/cloudfoundry/uaa/pull/2129 as alternative PR

PR #2134, opened 2022-12-09
refactor: rename TokenValidation

bruce-ricard:
* We believe that "TokenValidation" was very confusing, and that this class represents a jwt token that has had its signature verified.

* We understand the that in order for the signature to be verified, the checkSignature() method must be called, but in practice all of the class' factory methods do indeed call that method.

Co-Authored-by: Bruce Ricard <bricard@vmware.com>


strehle (approved):
in general ok

please check if we can ignore smell and findings from
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2134

swalchemist:
@strehle Strange that Sonar triggered on a pre-existing issue just because we renamed a method. In any case, we'll fit it. Almost all of the code smells are because of the deprecated spring-security-oauth2 library, which we do hope to replace soon. We'll look at the handful of other code smells for this class.

strehle:
I have closed the issues so sonar is green

swalchemist:
I have fixed the dead code bug found by Sonar and also all the code smells except for the ones about deprecated code.

Tallicia (approved):
New class name is much more clear. Cleaned up sonar lint issues are good to see. Looking forward to seeing more smells cleaned up in future refactoring.

bruce-ricard:
I just rebased from `develop`, mostly to get the fix that makes the tests not flake on HH.

PR #2158, opened 2023-01-10
Fix SAML config: convert SamlConfig toLower

### LINKED ISSUE #2157: Store saml key toLower to meet JKS requirements
Found issue by accident in 
https://uaa.uaa-acceptance-k8s.cf-app.com/

The usage of SAML keys in upper case causes "key not found issue" because KeyStore alias is converted to lower.
See
https://stackoverflow.com/questions/27070724/how-is-the-alias-name-used-in-java-keystore-class

Develop , e.g. https://uaa.uaa-acceptance-k8s.cf-app.com/

I am deploying the UAA

- locally only using gradlew
- using cf-release
- using cf-deployment
- K8s https://uaa.uaa-acceptance-k8s.cf-app.com/

### What did you do?

Login endpont callsed
https://uaa.uaa-acceptance-k8s.cf-app.com/

No error

### What did you see instead?

Exception in ZoneAwareMetadataGenerator

Thought I can omit the issue with https://github.com/cloudfoundry/uaa/pull/2156 but after same tests I think it is better to 
meet the current requirements from KeyStore
Why:
It worked some years ago but now not longer, so if someone performs an upgrade from an old UAA to this version, then this will run into the issue therefore this should be fixed

strehle:
Configuration from https://uaa.uaa-acceptance-k8s.cf-app.com uses UPPER Case Key ID
This causes an exception and non usable UAA for SAML.
Since it must have worked before this fix here should fix a regression


server/src/main/java/org/cloudfoundry/identity/uaa/impl/config/IdentityZoneConfigurationBootstrap.java
```diff
@@ -97,7 +98,7 @@ public void afterPropertiesSet() throws InvalidIdentityZoneDetailsException {
         samlKeys = ofNullable(samlKeys).orElse(EMPTY_MAP);
         for (Map.Entry<String, Map<String,String>> entry : samlKeys.entrySet()) {
             SamlKey samlKey = new SamlKey(entry.getValue().get("key"), entry.getValue().get("passphrase"), entry.getValue().get("certificate"));
-            definition.getSamlConfig().addKey(entry.getKey(), samlKey);
+            definition.getSamlConfig().addKey(ofNullable(entry.getKey()).orElseThrow(() -> new InvalidIdentityZoneDetailsException("SAML key id must not be null.", null)).toLowerCase(Locale.US), samlKey);
```
hsinn0: Are we sure that `Locale.US` is the right choice for the `String.toLowerCase()` call?  Should it not be `Locale.ROOT` instead?

hsinn0 (commented):
Are we sure that

server/src/main/java/org/cloudfoundry/identity/uaa/impl/config/IdentityZoneConfigurationBootstrap.java
strehle: I simply have seen that Locale.US is used on other places and therefore, but you are right

server/src/main/java/org/cloudfoundry/identity/uaa/impl/config/IdentityZoneConfigurationBootstrap.java
```diff
@@ -165,7 +166,7 @@ public IdentityZoneConfigurationBootstrap setSamlKeys(Map<String, Map<String, St
     }
 
     public IdentityZoneConfigurationBootstrap setActiveKeyId(String activeKeyId) {
-        this.activeKeyId = activeKeyId;
+        this.activeKeyId = activeKeyId != null ? activeKeyId.toLowerCase(Locale.US) : null;
```
hsinn0: This should be also changed to `Locale.ROOT` as done for line 101?

hsinn0 (approved):

PR #2168, opened 2023-01-18
Configure Content Security Policy script-src

### LINKED ISSUE #2086: Allow overriding Content security header
76.0.0 contains https://github.com/cloudfoundry/uaa/pull/1981/commits/72565fb56cd1f90af499119d32c891937f3c5a76 which disallows inline scripts.

On [GOV.UK PaaS ](https://login.cloud.service.gov.uk/login)we use javascript to improve some of the accessibility and on our customised passcode page to insert a copy code button that uses the clipboard api.

With the introduced change both of those things are no longer functional. 

Consider enabling a way to override the CSP settings.

jameshochadel:
https://github.com/cloudfoundry/uaa/commit/72565fb56cd1f90af499119d32c891937f3c5a76 disallowed inline scripts, but some users load scripts from other sources to add functionality to, for example, the login page. This change adds a new optional section to `uaa.yml` that allows custom script sources.

* The general approach is modeled on the `CorsFilter` and related `CorsConfiguration`.
* I wasn't sure where to put the new config class, so I co-located it with the `ContentSecurityPolicyFilter` itself.
* I couldn't find end-to-end tests for configuring filters from `uaa.yml`, so I just added a unit test for the new functionality. Let me know if integration tests for loading config exist and I'll add a test for the new configuration.

Addresses https://github.com/cloudfoundry/uaa/issues/2086.


uaa/src/main/webapp/WEB-INF/spring-servlet.xml
```diff
@@ -248,7 +248,13 @@
     </bean>
 
     <bean id="contentSecurityPolicyFilter"
-          class="org.cloudfoundry.identity.uaa.security.web.ContentSecurityPolicyFilter"/>
+          class="org.cloudfoundry.identity.uaa.security.web.ContentSecurityPolicyFilter">
+        <property name="cspAllowedScriptSrc"
+                  value="#{@config['contentSecurityPolicy']==null ? T(java.util.Arrays).asList('self') :
```
strehle: if you simply set self then CSP does not work. 'self' is wanted - I assume
-> ('''self''')

uaa/src/main/resources/uaa.yml
```diff
@@ -576,4 +576,10 @@ ldap:
     userDn: 'cn=admin,dc=test,dc=com'
     password: 'password'
     searchBase: 'dc=test,dc=com'
-    searchFilter: 'cn={0}'
\ No newline at end of file
+    searchFilter: 'cn={0}'
+
+# Set a custom content security policy
+# contentSecurityPolicy:
```
strehle: please remove it from here, examples should be placed in other places, e.g. cargo yml or uaa-release https://github.com/cloudfoundry/uaa-release/blob/develop/spec/compare/all-properties-set-uaa.yml

strehle (commented):
do you consider a setting in uaa-release ?
https://github.com/cloudfoundry/uaa-release/pull/84

uaa/src/main/webapp/WEB-INF/spring-servlet.xml
```diff
@@ -248,7 +248,13 @@
...
+                           @config['contentSecurityPolicy']['allowed']==null ? T(java.util.Arrays).asList('self') :
+                           @config['contentSecurityPolicy']['allowed']['scriptSrc']==null ? T(java.util.Arrays).asList('self') :
```
strehle: I recommend to simply use csp and omit allowed e.g. uaa.yml from 

```
uaa.csp.script-src:
  - 'self'
  - 'unsafe-inline'
```

would place csp section in uaa-release
https://github.com/cloudfoundry/uaa-release/blob/develop/jobs/uaa/templates/config/uaa.yml.erb#L845

csp defines different section .e.g. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy so currently we need only script-src , later we could add other in this section

jameshochadel:
@strehle Thank you for the review! I've addressed the feedback. I matched the `csp` section to the `cors` section structure. And I will make a separate PR to `uaa-release` as you suggest.

jameshochadel:
Sorry @strehle — I just re-requested your review, and it bumped all of the other reviewers off the PR for some reason. I don't have permission to add them back myself.

server/src/main/java/org/cloudfoundry/identity/uaa/security/web/ContentSecurityPolicyConfiguration.java
```diff
@@ -0,0 +1,27 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
+import java.util.Collections;
+import java.util.Set;
+
+/**
+ * By default, UAA only allows script-src 'self'. Custom script sources can be
+ * configured here.
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
```
strehle: is this needed , I dont see a reason, we dont read json but this is in bean defined

server/src/main/java/org/cloudfoundry/identity/uaa/security/web/ContentSecurityPolicyFilter.java
```diff
@@ -8,15 +8,27 @@
...
-                                 HttpServletResponse response, FilterChain chain)
+            HttpServletResponse response, FilterChain chain)
             throws IOException, ServletException {
 
-        response.setHeader("Content-Security-Policy", "script-src 'self'");
+        // If cspHeader were set when the class was instantiated, it would use
+        // the values in `configuration` before they're possibly overwritten
+        // by spring-servlet.xml, which itself reads values from uaa.yml. To
+        // avoid this, we set it lazily here instead.
+        if (this.cspHeader == null) {
```
strehle: you can also move the setting in the bean constructor, which is the best approach, so then the string can be final

server/src/main/java/org/cloudfoundry/identity/uaa/security/web/ContentSecurityPolicyFilter.java
```diff
@@ -31,4 +43,15 @@ protected boolean shouldNotFilter(HttpServletRequest request) {
         return pathsWithHtmlInlineScripts.stream()
                 .anyMatch(requestPath::startsWith);
     }
+
+    private String cspHeaderValue() {
```
strehle, line 48: can be called from construtor

uaa/src/main/webapp/WEB-INF/spring-servlet.xml
```diff
@@ -248,7 +248,13 @@
     </bean>
 
     <bean id="contentSecurityPolicyFilter"
-          class="org.cloudfoundry.identity.uaa.security.web.ContentSecurityPolicyFilter"/>
+          class="org.cloudfoundry.identity.uaa.security.web.ContentSecurityPolicyFilter">
```
strehle, line 251: here you can define a constructor and set your property simply as input for constructor

uaa/src/main/webapp/WEB-INF/spring-servlet.xml
```diff
@@ -248,7 +248,13 @@
...
+        <property name="cspAllowedScriptSrc"
+                  value="#{@config['csp']==null ? T(java.util.Arrays).asList(&quot;'self'&quot;) :
+                           @config['csp']==null ? T(java.util.Arrays).asList(&quot;'self'&quot;) :
+                           @config['csp']['scriptSrc']==null ? T(java.util.Arrays).asList(&quot;'self'&quot;) :
```
strehle: do you want to stay with scriptSrc ? I recommnd script-src to re-use same names from 
https://content-security-policy.com/script-src/

strehle (commented):
for me change is OK , other maintainer should also check it

strehle:
@jameshochadel did you see the comments, only wanted to know if you want this in next release. We currently have issues with the build pipeline but if these issues are fixed we create a new release

jameshochadel:
@strehle Sorry for the delay; I saw your comments but was waiting for another maintainer review like you mentioned. I'll go through them and address them now!

uaa/src/main/webapp/WEB-INF/spring-servlet.xml
jameshochadel: I'm happy to do either. I was matching the CORS configuration casing, but it looks like other sections of `uaa.yml` use a variety of casing (camelCase in `ratelimit`, snake_case in `encryption`...). I'll match it to the hyphenated case as you suggest.

server/src/main/java/org/cloudfoundry/identity/uaa/security/web/ContentSecurityPolicyConfiguration.java
jameshochadel: I brought it over from the CORS configuration, so I'm not sure if it's needed. I'll remove it.

uaa/src/main/webapp/WEB-INF/spring-servlet.xml
strehle: ok you can  stay with it, I know about these mixes... that if if you have open source, however I thought script-src might be easiert to understand - because of standard behind CSP - but I have no problem with the name.

jameshochadel:
@strehle I believe I've addressed your feedback. Thank you again.

strehle:
Ok, here I provided you a fix for the test and I think we should add a log , see my 2 commits from
https://github.com/cloudfoundry/uaa/commits/configure-csp-script-src

I will do log later, but please fix test in this PR and then we can merge it 

uaa-release is something you can do now

strehle (approved):
please fix test but LGTM

jameshochadel:
@strehle The test has been updated. Sorry about that and thank you again.

strehle:
sonar ok:
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2168

PR #2195, opened 2023-02-08
Refactor: YAML processor from spring

strehle:
Discussion about https://github.com/spring-projects/spring-boot/issues/33457

UAA uses own constructor but better rely on spring


bruce-ricard (approved):
Looks good to me. Deleting code is always a good thing.

Tallicia:
I'd merge this if I had permissions to.

PR #2208, opened 2023-02-27
[WIP only] Fix KeyInfo

strehle:
Problems: 
* Duplicated code ,[ Sonar rated 25% duplicated code ](https://sonarcloud.io/component_measures?metric=duplicated_lines_density&selected=cloudfoundry-identity-parent%3Aserver%2Fsrc%2Fmain%2Fjava%2Forg%2Fcloudfoundry%2Fidentity%2Fuaa%2Foauth%2FKeyInfo.java&view=list&id=cloudfoundry-identity-parent)
* Remove abstract class and combine it into one - remove dublication
* MAC for kty is not RFC complaint https://www.rfc-editor.org/rfc/rfc7518#section-6.1 . To ensure that MAC still works, support both, MAC and oct 
* Add support for EC verification - needed if OIDC Identity Provider is configured with ES??? alg. This might be the case because ES256 is recommended in [OIDC additional standards](https://openid.bitbucket.io/fapi/fapi-2_0-security-profile.html#name-cryptography-and-secrets)


model/src/test/java/org/cloudfoundry/identity/uaa/oauth/jwk/JsonWebKeyTest.java
```diff
@@ -0,0 +1,102 @@
+package org.cloudfoundry.identity.uaa.oauth.jwk;
```
strehle, line 1: renamed this class in https://github.com/cloudfoundry/uaa/pull/2116
So depending on what PR comes in first we can change it

strehle:
We should finish https://github.com/cloudfoundry/uaa/pull/2116 and then we can work again on this PR

PR #2245, opened 2023-03-16
feat: Increase the client redirect_uri length limit

hsinn0:
- Change the database data type from varchar to text type as mysql varchar type is subject to the table's column size limit

[#183903458]

### DIFF
--- a/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/ClientAdminEndpointsIntegrationTests.java
+++ b/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/ClientAdminEndpointsIntegrationTests.java
@@ -55,8 +55,10 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.Date;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import static junit.framework.Assert.assertEquals;
 import static junit.framework.Assert.assertFalse;
@@ -69,7 +71,6 @@ import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
-
 
 /**
  * @author Dave Syer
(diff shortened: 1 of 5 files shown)


peterhaochen47 (approved):

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/ClientAdminEndpointsIntegrationTests.java
```diff
@@ -50,13 +50,7 @@
 import org.springframework.util.MultiValueMap;
 import org.springframework.web.client.RestTemplate;
 
-import java.util.ArrayList;
```
strehle: Can you please revert import changes.
There is an ongoning discussion about clean code but we have different view on that but therefore please exclude such changes where they are not needed

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/clients/ClientAdminEndpointsMockMvcTests.java
```diff
@@ -64,15 +64,7 @@
 import org.springframework.web.context.WebApplicationContext;
 
 import javax.servlet.http.HttpServletResponse;
-import java.util.ArrayList;
```
strehle: See other comments about import changes

strehle (changes requested):
please revert import changes

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/ClientAdminEndpointsIntegrationTests.java
hsinn0: Done

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/clients/ClientAdminEndpointsMockMvcTests.java
hsinn0: Done

peterhaochen47:
Merged as all comments have been addressed.

PR #2253, opened 2023-03-21
Add additional logging if we pass the tests after passing retry of th…

hsinn0:
…e failed tests

- Because we did not realize that retry plugin was used and wondered why the test suite passed when there was a failed test in the log

[#184739970]


Tallicia:
The additional log message looks good.

PR #2263, opened 2023-03-28
Fix missing whitelist check for authority mapping

strehle:
IdP groups and/or claims are used for authority assignment These elements can be filtered with config option externalGroupsWhitelist This is useful especially for direct groupMappingMode = AS_SCOPES.


hsinn0:
Looks good to me

hsinn0 (approved):

PR #2268, opened 2023-04-03
Fix indexOf Checks in KeystoneAuthenticationManager

adrianhoelzl-sap:


bruce-ricard (approved):
Nice refactor. Thank you!

server/src/main/java/org/cloudfoundry/identity/uaa/authentication/manager/KeystoneAuthenticationManager.java
```diff
@@ -52,9 +52,9 @@ protected boolean evaluateResponse(Authentication authentication, ResponseEntity
 
     @Override
     protected KeystoneAuthenticationRequest getParameters(String username, String password) {
-        if (getRemoteUrl()!=null && getRemoteUrl().indexOf("/v2.0")>0) {
```
bruce-ricard, line 55: I think that you are actually fixing a bug. This should have been `>= 0`. If the URL happens to start with `/v2.0`, `indexOf` will return 0, and the conditional will fail.

Although in practice I guess it's impossible for a URL to start with `/`

adrianhoelzl-sap: The usage of `> 0` instead of `>= 0` was actually the Sonar finding I wanted to fix here. However, my IDE then recommended using `.contains(...)` instead. Let me know if I should change it back to `indexOf(...) >= 0` :)

strehle (approved):
LGTM

PR #2269, opened 2023-04-03
Remove unnecessary References to "apache.directory" Library

adrianhoelzl-sap:


model/src/main/java/org/cloudfoundry/identity/uaa/util/UaaStringUtils.java
```diff
@@ -47,6 +47,7 @@ public final class UaaStringUtils {
     private static final Pattern CTRL_PATTERN = Pattern.compile("[\n\r\t]");
 
     public static final String EMPTY_STRING = "";
+    public static final String[] EMPTY_STRING_ARRAY = {};
```
bruce-ricard: I don't understand the point of these 2 constants. I would rather see `""` and `{}` in the code than that.

Not having [magic values](https://en.wikipedia.org/wiki/Magic_number_(programming)) in the code is great. But replacing a magic `2` in the code by having `public final int TWO = 2` doesn't solve the problem. It just creates a useless redirect that makes the code more difficult to understand.

server/src/main/java/org/cloudfoundry/identity/uaa/cypto/EncryptionKeyService.java
```diff
@@ -18,13 +17,16 @@ public class EncryptionKeyService {
     public EncryptionKeyService(
             final @Value("${encryption.active_key_label}") String activeKeyLabel,
             final @Value("#{@config['encryption']['encryption_keys']}") List<EncryptionKey> encryptionKeys) {
-        if (Strings.isEmpty(activeKeyLabel)) {
+        if (activeKeyLabel == null || activeKeyLabel.isEmpty()) {
```
bruce-ricard: This change makes the code worse IMO. The `null` check was nicely hidden in the util method, and now it pops out.

server/src/main/java/org/cloudfoundry/identity/uaa/cypto/EncryptionKeyService.java
```diff
@@ -18,13 +17,16 @@ public class EncryptionKeyService {
...
             throw new NoActiveEncryptionKeyProvided(
               "UAA cannot be started without encryption key value uaa.encryption.active_key_label"
             );
         }
 
-        List<EncryptionKey> keysWithoutPassphrase = encryptionKeys.stream().filter(encryptionKey -> Strings.isEmpty(encryptionKey.getPassphrase())).collect(Collectors.toList());
+        List<EncryptionKey> keysWithoutPassphrase = encryptionKeys.stream().filter(encryptionKey -> {
+            final String passphrase = encryptionKey.getPassphrase();
+            return passphrase == null || passphrase.isEmpty();
+        }).collect(Collectors.toList());
```
bruce-ricard: This change makes the code worse IMO. The null check was nicely hidden in the util method, and now it pops out.

server/src/main/java/org/cloudfoundry/identity/uaa/security/ScimUserUpdateDiff.java
```diff
@@ -40,7 +39,8 @@ public boolean isAnythingOtherThanNameDifferent(String userId, ScimUser scimUser
 
     private boolean originsEquivalent(ScimUser scimUserFromRequest, ScimUser scimUserFromDb) {
         return scimUserFromDb.getOrigin().equals(scimUserFromRequest.getOrigin()) ||
-                (scimUserFromDb.getOrigin().equals(OriginKeys.UAA) && Strings.isEmpty(scimUserFromRequest.getOrigin()));
+                (scimUserFromDb.getOrigin().equals(OriginKeys.UAA) && (scimUserFromRequest.getOrigin() == null || scimUserFromRequest.getOrigin()
+                        .isEmpty()));
```
bruce-ricard: This change makes the code worse IMO. The null check was nicely hidden in the util method, and now it pops out.

server/src/test/java/org/cloudfoundry/identity/uaa/util/UaaUrlUtilsTest.java
```diff
@@ -133,12 +132,12 @@ void getParameterMapFromInvalidQueryString() {
         assertNotNull(map);
         assertEquals("code", map.get("response_type")[0]);
         assertEquals("", map.get("redirect_uri")[0]);
-        assertEquals(Strings.EMPTY_STRING_ARRAY, map.get("client_id"));
+        assertEquals(UaaStringUtils.EMPTY_STRING_ARRAY, map.get("client_id"));
```
bruce-ricard: I'd rather see `{}` here.

bruce-ricard (changes requested):
I disagree with the new null checks. 
I disagree in general with this kind of constants that don't bring any value. But I agree that you are not making the code any worse than it is right now. Only making it better by getting rid of the library calls.

But thank you for cleaning up the code, and attempting to get rid of a dependency.

server/src/main/java/org/cloudfoundry/identity/uaa/cypto/EncryptionKeyService.java
adrianhoelzl-sap: In the latest commit, I have added a method `isNullOrEmpty` in `UaaStringUtils` for these checks.

server/src/test/java/org/cloudfoundry/identity/uaa/util/UaaUrlUtilsTest.java
adrianhoelzl-sap: Unfortunately, we cannot use an array initializer at this location. Instead, we would need to use `new String[0]` here. IMHO, `UaaStringUtils.EMPTY_STRING_ARRAY` is easier to read than that. What do you think?

model/src/main/java/org/cloudfoundry/identity/uaa/util/UaaStringUtils.java
adrianhoelzl-sap: IMHO, using the `EMPTY_STRING` constant instead of `""` makes it more obvious that the empty string is actually intended and was not just forgotten.

adrianhoelzl-sap: I also just removed the `EMPTY_STRING_ARRAY` constant.

strehle (approved):
@bruce-ricard please check, your comments if solved

from my side ok

strehle:
@bruce-ricard are you ok to merge this. 
We use this code in our forks and we remove LDAP completely and therefore we clean-up this code. For us it means this code here should not depend on LDAP libraries anymore.

Anothther non-ldap related change:
isEmpty is now deprecated in Spring framework, therefore created own utils method in UaaStringUtils

PR #2270, opened 2023-04-03
Remove unused Variable in LoginInfoEndpoint.evaluateLoginHint

adrianhoelzl-sap:


bruce-ricard:
I tracked down the usage of the variable to this [commit](https://github.com/cloudfoundry/uaa/commit/333954df0df9327cea2a75c906a63aa5fea3544e) from 2 years ago.

It appears that the variable stoped being used, and that a new method parameter is now used for the following conditional.

@tack-sap as the author of this commit, could you please confirm that you meant to stop using that variable?

tack-sap (approved):
Yes, I can confirm that changing from resolving this inside the method to passing the List as request parameter was intended in that change.
Looks like I forgot to remove this variable then.

strehle:
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2270

PR #2271, opened 2023-04-03
refactor UAA token store: use good time types

bruce-ricard:
I started looking at this class because of a bug we found. I wanted to refactor first, so that I can fix the issue after that.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
@@ -77,7 +77,7 @@ public class UaaTokenStore implements AuthorizationCodeServices {
     private static final String SQL_CLEAN_STATEMENT = "delete from oauth_code where created < ? and expiresat = 0";
 
     private final DataSource dataSource;
-    private final long expirationTime;
+    private final long expirationTimeInMilliseconds;
     private final RandomValueStringGenerator generator = new RandomValueStringGenerator(32);
     private final RowMapper rowMapper = new TokenCodeRowMapper();
 
@@ -87,9 +87,9 @@ public class UaaTokenStore implements AuthorizationCodeServices {
         this(dataSource, DEFAULT_EXPIRATION_TIME_IN_MILLISECONDS);
     }
 
-    public UaaTokenStore(DataSource dataSource, long expirationTime) {
+    public UaaTokenStore(DataSource dataSource, long expirationTimeInMilliseconds) {
         this.dataSource = dataSource;
-        this.expirationTime = expirationTime;
+        this.expirationTimeInMilliseconds = expirationTimeInMilliseconds;
     }
 
     @Override
@@ -101,7 +101,7 @@ public class UaaTokenStore implements AuthorizationCodeServices {
         while ((tries++)<=max_tries) {
             try {
                 String code = generator.generate();
-                long expiresAt = System.currentTimeMillis()+getExpirationTime();
+                long expiresAt = System.currentTimeMillis()+ getExpirationTimeInMilliseconds();
                 String userId = authentication.getUserAuthentication()==null ? null : ((UaaPrincipal)authentication.getUserAuthentication().getPrincipal()).getId();
                 String clientId = authentication.getOAuth2Request().getClientId();
                 SqlLobValue data = new SqlLobValue(serializeOauth2Authentication(authentication));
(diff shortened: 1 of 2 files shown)


peterhaochen47 (approved):

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
```diff
@@ -245,26 +245,31 @@ public TokenCode mapRow(ResultSet rs, int rowNum) throws SQLException {
...
+                    (expiresat == 0) ?
+                            Optional.empty() :
+                            Optional.of(Instant.ofEpochMilli(expiresat));
+
+            return createTokenCode(code, userid, client_id, expirationInstant, created, authentication);
         }
     }
 
-    public TokenCode createTokenCode(String code, String userId, String clientId, long expiresAt, Timestamp created, byte[] authentication) {
+    public TokenCode createTokenCode(String code, String userId, String clientId, Optional<Instant> expiresAt, Instant created, byte[] authentication) {
```
strehle: According our clean code guide we should not use Optional in method parameters, but Optional should be used for return values only, because then you can control the rule of https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Optional.html with Optional.ofNullable(xxx) , but if you allow Optional in method variables of PUBLIC methods you need to check also that Optional is not null

strehle (dismissed):
Please see and follow sonar to fix all in section issues 

https://sonarcloud.io/project/issues?resolved=false&pullRequest=2271&id=cloudfoundry-identity-parent

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
```diff
@@ -298,10 +303,10 @@ public String getUserId() {
         }
 
         public boolean isExpired() {
-            if (getExpiresAt()==0) {
-                return new Timestamp(System.currentTimeMillis()-getExpirationTime()).after(getCreated());
+            if (getExpiresAt().isEmpty()) {
+                return Instant.now().minus(getExpirationTime()).isAfter(getCreated());
             } else {
-                return getExpiresAt() < System.currentTimeMillis();
+                return getExpiresAt().get().isBefore(Instant.now());
```
strehle: Sonar is happy if you do

        public boolean isExpired() {
            Optional<Instant> optionalExpiresAt = getExpiresAt();
            if (optionalExpiresAt.isEmpty()) {
                return Instant.now().minus(getExpirationTime()).isAfter(getCreated());
            } else {
                return optionalExpiresAt.get().isBefore(Instant.now());
            }
        }

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
bruce-ricard: I agree. I was planning an improvement and I forgot to do it. Coming up!

Thanks for catching this.

bruce-ricard:
I believe my latest changes address both issues.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
bruce-ricard: I fully acknowledge that I didn't technically fix this, though the method is only for testing purposes. So it seems to me that it is OK in this case. Let me know what you think.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
```diff
@@ -245,31 +243,35 @@ public TokenCode mapRow(ResultSet rs, int rowNum) throws SQLException {
...
+                return new LegacyTokenCode(code, userid, created, client_id, authentication);
+            } else {
+                return new NewTokenCode(code, userid, Instant.ofEpochMilli(expiresat), client_id, authentication);
+            }
         }
     }
 
-    public TokenCode createTokenCode(String code, String userId, String clientId, long expiresAt, Timestamp created, byte[] authentication) {
-        return new TokenCode(code,userId,clientId,expiresAt,created,authentication);
+    public TokenCode createTokenCodeForTesting(String code, String userId, String clientId, Optional<Instant> expiresAt, Instant created, byte[] authentication) {
```
strehle, line 257: do want still go with Optional in a public method ?

strehle (commented):
clean sonar, thanks
https://sonarcloud.io/project/issues?resolved=false&pullRequest=2271&id=cloudfoundry-identity-parent

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
bruce-ricard: I don't love it, but as I said in my other message, it's a public method only for testing purposes. Actually I may just change it to package private, that should help a bit.

PR #2272, opened 2023-04-05
Fix: Fixing performance while calling containsAll over list

saurabh-rahate:
Improving performance in case of 
```java
list.containsAll(anotherList)
```

by using `HashSet` wrapper around one of the lists
```java
new HashSet<>(list).containsAll(anotherList)
```

This improves time complexity.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/user/UserInfo.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/user/UserInfo.java
@@ -67,12 +67,10 @@ public class UserInfo {
     }
 
     protected boolean compareRoles(List<String> l1, List<String> l2) {
-        if (l1==null && l2==null) {
-            return true;
-        } else if (l1==null || l2==null) {
-            return false;
+        if (l1 == null || l2 == null) {
+            return l1 == l2;
         }
-        return new HashSet<>(l1).containsAll(l2) && new HashSet<>(l2).containsAll(l1);
+        return new HashSet<>(l1).equals(new HashSet<>(l2));
     }
 
     @Override
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/user/UserInfo.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/user/UserInfo.java
@@ -20,6 +20,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import org.springframework.util.LinkedMultiValueMap;
 import org.springframework.util.MultiValueMap;
 
+import java.util.HashSet;
 import java.util.List;
 
 public class UserInfo {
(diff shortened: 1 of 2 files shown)


strehle (approved):
ok for me

strehle:
@Tallicia  @bruce-ricard  any concerns

Tallicia:
The 8 test look good and the optimization does as well.

Tallicia (approved):
Nice improvement.

Tallicia:
Would be nice to see performance tests to see if sets are of large enough size in practice to know the real world performance improvement.

server/src/main/java/org/cloudfoundry/identity/uaa/user/UserInfo.java
```diff
@@ -71,7 +72,7 @@ protected boolean compareRoles(List<String> l1, List<String> l2) {
         } else if (l1==null || l2==null) {
             return false;
         }
-        return l1.containsAll(l2) && l2.containsAll(l1);
+        return new HashSet<>(l1).containsAll(l2) && new HashSet<>(l2).containsAll(l1);
```
bruce-ricard: It seems to me like we should be able to leverage the `Set` `equals` [method](https://docs.oracle.com/javase/7/docs/api/java/util/Set.html#equals(java.lang.Object))

The linked documentation states:

>Returns true if the specified object is also a set, the two sets have the same size, and every member of the specified set is contained in this set

We should be able to write
```
return new HashSet<>(l1).equals(new HashSet<>(l2))
```

bruce-ricard (changes requested):
Thank you for making performance improvements. 

To be perfectly frank, I am not convinced that this change will improve performance in the general practical world. I don't know if most objects don't have one or two roles. And this change forces the creation of 2 new hashsets every single time.

Also, we cannot find any usage of that `equals` method in the UAA code base. I don't think that this change will have any impact at all. 

Additionally, I am not convinced that Java doesn't already use HashSets internally to resolve `containsAll` calls on `List` object.

@Tallicia and me

server/src/main/java/org/cloudfoundry/identity/uaa/user/UserInfo.java
saurabh-rahate: Changed implementation to use `Set` `equals` method.

strehle:
I have tested this and created an extra PR https://github.com/cloudfoundry/uaa/pull/2294
With a test of role about 10000 the difference is 300 ms compared to 30 ms , in case of more e.g. 100000xxx it sometimes runs into many seconds, so that I decided to stay only with 10000
So it brings the expected result from times described here:
https://www.baeldung.com/java-hashset-arraylist-contains-performance
List -> O(mn)
HashSet -> O(n)

Finally, I will merge this, so thanks for your PR

PR #2273, opened 2023-04-06
Fix for ldap.ssl.skipverification

cache-sk:
If you are using directory server with ssl (ldaps), without tls (ldap.ssl.tls: none) and with, for example, self-signed certificate, ldap.ssl.skipverification property is ignored and connection fails with InternalAuthenticationServiceException with root exception SSLHandshakeException: PKIX path building failed ... etc.

When tls is set to none, [SimpleDirContextAuthenticationStrategy](https://github.com/cloudfoundry/uaa/blob/a1fff3d1a2f5969f1c25bfc80e253f3f5a567505/server/src/main/java/org/cloudfoundry/identity/uaa/provider/ldap/ProcessLdapProperties.java#L98) is used, so SocketFactory is not instanced in code (as for tls allowed a couple of lines lower), but relies on JNDI value of `java.naming.ldap.factory.socket`. However, even if the value is correctly set to `SkipSslLdapSocketFactory`, instance of `LdapSocketFactory` is created.

The reason is that `SkipSslLdapSocketFactory` is extended from `BaseLdapSocketFactory`, which creates an instance of `LdapSocketFactory` in `getDefault()` method. And according to the [documentation](https://docs.oracle.com/javase/jndi/tutorial/ldap/security/ssl.html), this method is used to create instace of SocketFactory  defined in `java.naming.ldap.factory.socket`.

The simplest solution is to override `getDefault()` method in `SkipSslLdapSocketFactory` (in same manner as in `BaseLdapSocketFactory` in this pull request).

Maybe this three classes should be refactored in future.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/security/SkipSslLdapSocketFactory.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/security/SkipSslLdapSocketFactory.java
@@ -19,8 +19,18 @@ import javax.net.ssl.X509TrustManager;
 import java.security.SecureRandom;
 import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
+import javax.net.SocketFactory;
 
 public class SkipSslLdapSocketFactory extends BaseLdapSocketFactory {
+
+    private static SocketFactory instance;
+    public static SocketFactory getDefault() {
+        if (instance == null) {
+            instance = new SkipSslLdapSocketFactory();
+        }
+
+        return instance;
+    }
 
     public SkipSslLdapSocketFactory() {
         try {


strehle:
From text I understand this , but from Code I am not sure and since there is no test it gets hard for me to understand what you fix with this PR.

Sonar also wants some test coverage: https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2273

cache-sk:
Hello, I don't know if I fully understand what I have to test.

Changed code is almost exact copy of lines [10](https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/security/BaseLdapSocketFactory.java#L10) and 13-19 from BaseLdapSocketFactory.java (which are also not covered by tests), just returned instance is changed to SkipSslLdapSocketFactory.

Should i write a test, which check if getDefault() in SkipSslLdapSocketFactory returns instance of SkipSslLdapSocketFactory instead of LdapSocketFactory (which is obvious), or should i write a test, which starts embeded ldap server with generated self signed certificate to check, if java runtime implementation do what is written in java documentation, when is `java.naming.ldap.factory.socket` set to custom SocketFactory?

bruce-ricard:
Ideally you'd right a functional test that explains what your change is fixing. In this case it looks like it would be an integration test with an LDAP server.

What practical problem are you trying to solve? Are you unable to use the UAA server with an LDAP server without SSL?

cache-sk:
UAA server with an LDAP server With SSL, self signed cert, and without TLS, as i wrote in first sentence.

How to reproduce:

Setup local LDAP server, with SSL, with self signed cert, without TLS and DO NOT ad that cert to local trust store. 

NOTE: I've never done it, I use an existing server to which I have access to and a third person gave me the access data (working in Apache Directory Studio).

Use configuration like this:
```
spring_profiles: default,ldap
ldap:
  enabled: true
  override: true
  profile:
    file: ldap/ldap-search-and-bind.xml
  base:
    url: 'ldaps://localhost:636/'
    userDn: cn=****
    password: ****
    searchBase: ou=****
    searchFilter: cn={0}
    mailAttributeName: mail
    mailSubstitute: '{0}@****'
    mailSubstituteOverridesLdap: true
    referral: follow
  addShadowUserOnLogin: true
  storeCustomAttributes: true
  groups:
    file: ldap/ldap-groups-map-to-scopes.xml
    searchBase: ou=****
    groupSearchFilter: member={0}
    searchSubtree: true
    maxSearchDepth: 10
    autoAdd: true
    ignorePartialResultException: true
    groupRoleAttribute: spring.security.ldap.dn
  ssl:
    skipverification: true
    tls: none 
```

Run UAA from sources (i.e. gradlew run), try to log on in UAA. You will get ssl crt error in log.
It doesn't matter if ldap.ssl.skipverification is set to true.

Reason is described in first comment - if TLS is none, `LdapSocketFactory` is used instead of `SkipSslLdapSocketFactory`, because both is inherited from `BaseLdapSocketFactory`, which implementation of `getDefault()` returns `LdapSocketFactory`.

After using my "patch", skipverification is honored.

strehle:
also no LDAP expert but found an existing test for expired certificate so forked this and here I see the issue in develop branch and thus created https://github.com/cloudfoundry/uaa/pull/2313

strehle (approved):
sonar happy
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2313

created in another PR the functional / regression tests

PR #2281, opened 2023-04-12
fix: incorrect elapsed time determination

bruce-ricard:
* A recent previous refactor commit was using the `getNano()` method, thinking that it would return the total number of nanoseconds in the duration, which is does not. It returns the number of nanoseconds within one second, which is not at all what we wanted.

* We decided to go ahead and just use the `getSeconds` method, which this time actually returns the total number of seconds in the duration object. We understand that previously the precision of this was to the millisecond level. We believe that it's OK if we change to the second level. The cleanup will still happen every n minutes, just not at the precise millisecond.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
@@ -214,7 +214,7 @@ public class UaaTokenStore implements AuthorizationCodeServices {
         Instant last = lastClean.get();
         //check if we should expire again
         Instant now = Instant.now();
-        if (Duration.between(last, now).getSeconds() > getExpirationTime().getSeconds()) {
+        if (enoughTimeHasPassedSinceLastExpirationClean(last, now)) {
             //avoid concurrent deletes from the same UAA - performance improvement
             if (lastClean.compareAndSet(last, last.plus(getExpirationTime()))) {
                 try {
@@ -228,6 +228,10 @@ public class UaaTokenStore implements AuthorizationCodeServices {
                 }
             }
         }
+    }
+
+    private boolean enoughTimeHasPassedSinceLastExpirationClean(Instant last, Instant now) {
+        return Duration.between(last, now).getSeconds() > getExpirationTime().getSeconds();
     }
 
     public Duration getExpirationTime() {
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
@@ -231,7 +231,7 @@ public class UaaTokenStore implements AuthorizationCodeServices {
     }
 
     private boolean enoughTimeHasPassedSinceLastExpirationClean(Instant last, Instant now) {
-        return Duration.between(last, now).getSeconds() > getExpirationTime().getSeconds();
+        return Duration.between(last, now).toMillis() > getExpirationTime().toMillis();
     }
 
     public Duration getExpirationTime() {
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
@@ -214,7 +214,7 @@ public class UaaTokenStore implements AuthorizationCodeServices {
         Instant last = lastClean.get();
         //check if we should expire again
         Instant now = Instant.now();
-        if (Duration.between(last, now).getNano() > getExpirationTime().getNano()) {
+        if (Duration.between(last, now).getSeconds() > getExpirationTime().getSeconds()) {
             //avoid concurrent deletes from the same UAA - performance improvement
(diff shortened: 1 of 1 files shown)


peterhaochen47:
> ...getNano() method...returns the number of nanoseconds within one second

So `<some-duration>.getNano()` returns a constant? That's surprising...

Anyway, LGTM

peterhaochen47 (approved):

bruce-ricard:
> > ...getNano() method...returns the number of nanoseconds within one second
> 
> So `<some-duration>.getNano()` returns a constant? That's surprising...
> 

Argh, I didn't explain properly. What I meant was that `getNano()` returns the number of nanoseconds with the last second of the duration. Darn that's really hard to explain. Even the [official doc](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#:~:text=getNano,-public%20int%20getNano&text=int%20getNano()-,Gets%20the%20number%20of%20nanoseconds%20within%20the%20second%20in%20this,to%20the%20length%20in%20seconds.) struggles to give a good explanation.

Tallicia:
@strehle is fix @bruce-ricard and I worked on good to merge?

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
```diff
@@ -214,7 +214,7 @@ protected void performExpirationClean() {
         Instant last = lastClean.get();
         //check if we should expire again
         Instant now = Instant.now();
-        if (Duration.between(last, now).getNano() > getExpirationTime().getNano()) {
+        if (Duration.between(last, now).getSeconds() > getExpirationTime().getSeconds()) {
```
strehle: why not using toMillis for both, then we are on same level as before with "currentTimeMillis"

bruce-ricard: I didn't realize that there was a [toMillis](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#toMillis--) method. Good to know! 

There's also a `toNanos()`, in addition to `getNano()`, and they don't return at all the same thing. Yay.

`toMillis()` would indeed make it a true refactor. We thought that using seconds was good enough of a precision for this particular use case. 

Originally I tried to use the [compareTo](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#compareTo-java.time.Duration-) method, but I can never figure out which way it goes.

strehle: ok , so we have other usages of Duration in combination always with toMillis, 

However I did some debugging of the tests and for me there is still the bug that the cleanup is called to often.
I dont see an update of lastClean.set(Instant.now());

See tests cleanUpLegacyCodesCodesWithoutExpiresAtAfter3Days and cleanUpExpiredTokensBasedOnExpiresField
They expect 0 because cleanup is done ALWAYS, but with default 5 min the existing codes stay in DB.

strehle: see here from my debugging I updated lastCleanup

https://github.com/strehle/uaa/commit/7a701143d2b3ac140705d5e9bccbe09ae0615641

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
```diff
@@ -214,7 +214,7 @@ protected void performExpirationClean() {
         Instant last = lastClean.get();
```
strehle, line 214: This lastClean instance - before it was only a value needs to be set after first usage, 
To me this caused the issue, that the check in line 217 always was true and cleanup is executed always.
Even with lastClean.compareAndSet(.....) the lastClean is not set to a recent Instant but still in year 1970....

strehle (changes requested):
you can stay with nano, but use always to toXXX method

However the performance issue will be only solved if lastClean is corrected after first usage, see my proposal 
https://github.com/strehle/uaa/commit/7a701143d2b3ac140705d5e9bccbe09ae0615641

bruce-ricard:
I'm aware that these changes don't fix the issue. They were only refactoring changes.

strehle:
> I'm aware that these changes don't fix the issue. They were only refactoring changes.

ok, but then please either use toMillis (because internally we deal always with milli-seconds) or now you have an Instant Object and you could also refactor line 217 to something like

`if (getExpirationTime().minus(Duration.between(last, now)).isNegative()) {`

bruce-ricard:
> `if (getExpirationTime().minus(Duration.between(last, now)).isNegative()) {`

I like this! It's not great, but I think it's the best we'll get with the available interface.

bruce-ricard:
Actually I changed my mind. 

```
getExpirationTime().minus(Duration.between(last, now)).isNegative()
```

Is really not that easy to read. I decided to use the milliseconds as originally stated.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
bruce-ricard: I'll fix that in my next PR.

strehle (approved):

PR #2291, opened 2023-04-19
Fix auth code cleanup function

bruce-ricard:
It will now properly only clean once every 5 minutes

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
@@ -84,7 +84,7 @@ public class UaaTokenStore implements AuthorizationCodeServices {
     private final RandomValueStringGenerator generator = new RandomValueStringGenerator(32);
     private final RowMapper rowMapper = new TokenCodeRowMapper();
 
-    private AtomicReference<Instant> lastClean = new AtomicReference<>(Instant.EPOCH);
+    private final AtomicReference<Instant> lastClean = new AtomicReference<>(Instant.EPOCH);
 
     private final Clock clock;
 
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
@@ -228,7 +228,7 @@ public class UaaTokenStore implements AuthorizationCodeServices {
         Instant now = Instant.now(clock);
         if (enoughTimeHasPassedSinceLastExpirationClean(last, now)) {
             //avoid concurrent deletes from the same UAA - performance improvement
-            if (lastClean.compareAndSet(last, last.plus(getExpirationTime()))) {
+            if (lastClean.compareAndSet(last, now)) {
                 try {
                     JdbcTemplate template = new JdbcTemplate(dataSource);
                     int expired = template.update(SQL_EXPIRE_STATEMENT, now.toEpochMilli());
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
@@ -112,7 +112,8 @@ public class UaaTokenStore implements AuthorizationCodeServices {
         performExpirationClean();
         JdbcTemplate template = new JdbcTemplate(dataSource);
         int tries = 0;
-        while ((tries++)<=max_tries) {
+        while (true) {
+            tries++;
             try {
                 String code = generator.generate();
                 Instant expiresAt = Instant.now().plus(getExpirationTime());
@@ -132,7 +133,6 @@ public class UaaTokenStore implements AuthorizationCodeServices {
                 if (tries>=max_tries) throw exists;
             }
         }
-        return null;
     }
 
     @Override
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
@@ -45,6 +45,7 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Timestamp;
 import java.sql.Types;
+import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.Collection;
@@ -85,13 +86,24 @@ public class UaaTokenStore implements AuthorizationCodeServices {
 
     private AtomicReference<Instant> lastClean = new AtomicReference<>(Instant.EPOCH);
 
+    private final Clock clock;
+
     public UaaTokenStore(DataSource dataSource) {
         this(dataSource, DEFAULT_EXPIRATION_TIME);
     }
 
     public UaaTokenStore(DataSource dataSource, Duration expirationTime) {
+        this(dataSource, expirationTime, Clock.systemDefaultZone());
+    }
+
+    private UaaTokenStore(DataSource dataSource, Duration expirationTime, Clock clock) {
         this.dataSource = dataSource;
         this.expirationTime = expirationTime;
+        this.clock = clock;
+    }
+
+    static UaaTokenStore constructorForTesting(DataSource dataSource, Duration expirationTime, Clock clock) {
+        return new UaaTokenStore(dataSource, expirationTime, clock);
     }
 
     @Override
@@ -213,7 +225,7 @@ public class UaaTokenStore implements AuthorizationCodeServices {
     protected void performExpirationClean() {
         Instant last = lastClean.get();
         //check if we should expire again
-        Instant now = Instant.now();
+        Instant now = Instant.now(clock);
         if (enoughTimeHasPassedSinceLastExpirationClean(last, now)) {
             //avoid concurrent deletes from the same UAA - performance improvement
             if (lastClean.compareAndSet(last, last.plus(getExpirationTime()))) {
@@ -314,7 +326,7 @@ public class UaaTokenStore implements AuthorizationCodeServices {
 
         @Override
         boolean isExpired() {
-            return expiresAt.isBefore(Instant.now());
+            return expiresAt.isBefore(Instant.now(clock));
         }
 
         @Override
@@ -343,7 +355,7 @@ public class UaaTokenStore implements AuthorizationCodeServices {
 
         @Override
         boolean isExpired() {
-            return Instant.now().minus(getExpirationTime()).isAfter(created);
+            return Instant.now(clock).minus(getExpirationTime()).isAfter(created);
         }
 
         @Override
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
@@ -111,9 +111,9 @@ public class UaaTokenStore implements AuthorizationCodeServices {
         final int max_tries = 3;
         performExpirationClean();
         JdbcTemplate template = new JdbcTemplate(dataSource);
-        int tries = 0;
+        int attempt = 0;
         while (true) {
-            tries++;
+            attempt++;
             try {
                 String code = generator.generate();
                 Instant expiresAt = Instant.now().plus(getExpirationTime());
@@ -130,7 +130,7 @@ public class UaaTokenStore implements AuthorizationCodeServices {
                 }
                 return code;
             } catch (DataIntegrityViolationException exists) {
-                if (tries>=max_tries) throw exists;
+                if (attempt>=max_tries) throw exists;
             }
         }
     }
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
@@ -112,8 +112,7 @@ public class UaaTokenStore implements AuthorizationCodeServices {
         performExpirationClean();
         JdbcTemplate template = new JdbcTemplate(dataSource);
         int attempt = 0;
-        while (true) {
-            attempt++;
+        while ((attempt++)<=maxAttempts) {
             try {
                 String code = generator.generate();
                 Instant expiresAt = Instant.now().plus(getExpirationTime());
@@ -133,6 +132,7 @@ public class UaaTokenStore implements AuthorizationCodeServices {
                 if (attempt>=maxAttempts) throw exists;
             }
         }
+        return null;
     }
 
     @Override
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
@@ -108,7 +108,7 @@ public class UaaTokenStore implements AuthorizationCodeServices {
 
     @Override
     public String createAuthorizationCode(OAuth2Authentication authentication) {
-        final int max_tries = 3;
+        final int maxAttempts = 3;
         performExpirationClean();
         JdbcTemplate template = new JdbcTemplate(dataSource);
         int attempt = 0;
@@ -130,7 +130,7 @@ public class UaaTokenStore implements AuthorizationCodeServices {
                 }
                 return code;
             } catch (DataIntegrityViolationException exists) {
-                if (attempt>=max_tries) throw exists;
+                if (attempt>=maxAttempts) throw exists;
             }
         }
     }
--- a/server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
+++ b/server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
@@ -236,19 +236,23 @@ class UaaTokenStoreTests {
         ConstantClock clock = new ConstantClock();
         store = UaaTokenStore.constructorForTesting(dataSource, Duration.ofMinutes(1), clock);
 
-
         int count = 10;
         long oneday = 1000 * 60 * 60 * 24;
         for (int i = 0; i < count; i++) {
             legacyCodeServices.createAuthorizationCode(clientAuthentication);
         }
+
         assertThat(jdbcTemplate.queryForObject("SELECT count(*) FROM oauth_code", Integer.class), is(count));
+
         jdbcTemplate.update("UPDATE oauth_code SET created = ?", new Timestamp(System.currentTimeMillis() - (2 * oneday)));
         clock.timeItIs = Instant.now();
+
         assertThrows(InvalidGrantException.class, () -> store.consumeAuthorizationCode("non-existent"));
         assertThat(jdbcTemplate.queryForObject("SELECT count(*) FROM oauth_code", Integer.class), is(count));
+
         jdbcTemplate.update("UPDATE oauth_code SET created = ?", new Timestamp(System.currentTimeMillis() - (4 * oneday)));
         clock.timeItIs = Instant.now().plus(Duration.ofMinutes(10));
+
         assertThrows(InvalidGrantException.class, () -> store.consumeAuthorizationCode("non-existent"));
         assertThat(jdbcTemplate.queryForObject("SELECT count(*) FROM oauth_code", Integer.class), is(0));
     }
--- a/server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
+++ b/server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
@@ -210,14 +210,14 @@ class UaaTokenStoreTests {
 
     @Test
     void cleanUpExpiredTokensBasedOnExpiresField() {
-        TestClock clock = new TestClock();
+        ConstantClock clock = new ConstantClock();
         Duration expirationTime = Duration.ofHours(1);
         store = UaaTokenStore.constructorForTesting(dataSource, expirationTime, clock);
 
         int count = 10;
         String lastCode = null;
         for (int i = 0; i < count; i++) {
-            clock.instants.add(Instant.now());
+            clock.timeItIs = Instant.now();
             lastCode = store.createAuthorizationCode(clientAuthentication);
         }
         assertThat(jdbcTemplate.queryForObject("SELECT count(*) FROM oauth_code", Integer.class), is(count));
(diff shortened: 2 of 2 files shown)


server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
```diff
@@ -450,5 +468,24 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
...
+        public Clock withZone(ZoneId zone) {
+            return this;
+        }
+
+        @Override
+        public Instant instant() {
+            return instants.remove();
+        }
+    }
+
```
bruce-ricard: That's the only way I found to do what I wanted to do. Please advice if you know a better way.

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
```diff
@@ -450,5 +468,24 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
         }
     }
 
+    public class TestClock extends Clock {
```
swalchemist: The class name is a bit confusing - sounds like this is a unit test class.

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
```diff
@@ -206,32 +210,46 @@ void retrieveNonExistentToken() {
...
+        store = UaaTokenStore.constructorForTesting(dataSource, Duration.ofMinutes(1), clock);
+
+
         int count = 10;
         long oneday = 1000 * 60 * 60 * 24;
         for (int i = 0; i < count; i++) {
             legacyCodeServices.createAuthorizationCode(clientAuthentication);
         }
         assertThat(jdbcTemplate.queryForObject("SELECT count(*) FROM oauth_code", Integer.class), is(count));
         jdbcTemplate.update("UPDATE oauth_code SET created = ?", new Timestamp(System.currentTimeMillis() - (2 * oneday)));
```
swalchemist: If we're mocking the clock for this test, it would be best to choose a fixed point in time instead of starting it based on the system clock. Otherwise, you'll get clock ticks that you can't control for.

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
```diff
@@ -206,32 +210,46 @@ void retrieveNonExistentToken() {
...
+        clock.instants.add(Instant.now());
         assertThrows(InvalidGrantException.class, () -> store.consumeAuthorizationCode("non-existent"));
```
swalchemist, line 249: The three blocks of assertions make this test a bit hard to follow. Maybe it wants to be two or three different tests?

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
```diff
@@ -206,32 +210,46 @@ void retrieveNonExistentToken() {
...
         final String finalLastCode = lastCode;
+        clock.instants.add(Instant.now().plus(expirationTime));
+        clock.instants.add(Instant.now().plus(expirationTime));
+
         assertThrows(InvalidGrantException.class, () -> store.consumeAuthorizationCode(finalLastCode));
         assertThat(jdbcTemplate.queryForObject("SELECT count(*) FROM oauth_code", Integer.class), is(0));
     }
 
     @Test
     void cleanUpLegacyCodesCodesWithoutExpiresAtAfter3Days() {
```
swalchemist: This test method name is unparseable. Could it be reduced to English?

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
```diff
@@ -450,5 +468,24 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
         }
     }
 
+    public class TestClock extends Clock {
+        public Queue<Instant> instants = new LinkedList<Instant>();
```
swalchemist: Could you improve the name here to make it more clear that this is a list of instances that will be returned by the class?

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
```diff
@@ -450,5 +468,24 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
...
+            return ZoneId.systemDefault();
+        }
+
+        @Override
+        public Clock withZone(ZoneId zone) {
+            return this;
+        }
+
+        @Override
+        public Instant instant() {
```
swalchemist, line 486: Safe to assume that this is called when now() is invoked? Why not modify now() directly?

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
```diff
@@ -450,5 +468,24 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
         }
     }
 
+    public class TestClock extends Clock {
+        public Queue<Instant> instants = new LinkedList<Instant>();
+
+        @Override
+        public ZoneId getZone() {
+            return ZoneId.systemDefault();
```
swalchemist, line 477: Using the system default zone wouldn't be good in some situations, but probably works fine for this test.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
```diff
@@ -83,24 +84,36 @@ public class UaaTokenStore implements AuthorizationCodeServices {
...
     @Override
     public String createAuthorizationCode(OAuth2Authentication authentication) {
-        final int max_tries = 3;
+        final int maxAttempts = 3;
         performExpirationClean();
         JdbcTemplate template = new JdbcTemplate(dataSource);
-        int tries = 0;
-        while ((tries++)<=max_tries) {
+        int attempt = 0;
+        while (true) {
```
swalchemist: The check on maxTries has moved inside a catch block. Is there a potential infinite loop here?

bruce-ricard: It hasn't moved. It disappeared. As I explain in the commit message, this expression is always `true`, as explained by IntelliJ.

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
bruce-ricard: I don't disagree. I hadn't read it until now.

It's a bit awkward to read a comment about an old line that I didn't touch. Are you asking me to fix that in this PR?

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
bruce-ricard: It won't tick though. I'm creating a static clock that will return that exact time.

I don't think that I understand what you mean.

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
bruce-ricard: Same as the other comment about code I didn't change.

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
bruce-ricard: OK, interesting. IMO `ClockTest` is a test class. `TestClock` is a clock for tests.

But actually I think that I can improve its name, so I'll try that.

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
swalchemist: I was referring to things like new `Timestamp(System.currentTimeMillis() - (2 * oneday))` that are still using the system clock.

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
swalchemist: If I write a unit test for the Clock class, I would name it ClockTest, which sounds like TestClock. So "Test" tends to be overused as a name.

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
swalchemist: Yes, though I wouldn't insist on it since you didn't introduce that name in this PR.

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
bruce-ricard: `now()` is not a `Clock` method, it's an `Instant` method, that calls the `Clock#instant()` method.

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
swalchemist: You did make the code blocks longer, though. An easy fix might be adding some whitespace.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
```diff
@@ -83,24 +84,36 @@ public class UaaTokenStore implements AuthorizationCodeServices {
...
+    private final AtomicReference<Instant> lastClean = new AtomicReference<>(Instant.EPOCH);
+
+    private final Clock clock;
 
     public UaaTokenStore(DataSource dataSource) {
         this(dataSource, DEFAULT_EXPIRATION_TIME);
     }
 
     public UaaTokenStore(DataSource dataSource, Duration expirationTime) {
+        this(dataSource, expirationTime, Clock.systemDefaultZone());
```
hsinn0, line 96: Methods such as `Instant.now()` returns systemUTC instant.  To avoid potential timezone mix-up, we'd better use `Clock.systemUTC()` here.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
hsinn0: ~~Although the change is not going to cause an infinite loop because of the `throw` in the `catch` block, this change does modify the code behavior.
The old code will gracefully exit the while loop when `tries` becomes 3.  The new code will throw `DataIntegrityViolationException`.   Notice that the post-increment operator (`++`) was used in the condition in the old code.~~
No that is not it.  Let me re-read this.  I see.  Part of the method is not well-written.  So the logic is:
- return immediately if `template.update()` was successful but try couple of more times in if the update was not successful.

The throw and rethrow of `DataIntegrityViolationException` is unnecessary.  The code can just count the number of failed updates then throw exception once if the `max_tries` go over the `tries`.  This code should be changed for code efficiency and performance reason.

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
hsinn0: > ClockTest is a test class. TestClock is a clock for tests

Yep, that conforms with the Java test class naming pattern.

hsinn0 (changes requested):

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
bruce-ricard: That's a good point. I don't think that I understand very much the difference between the two.

As far as I can tell, `Instant.now()` will return the same value, regardless of the clock. But then it may get mixed up if we ask for the date of the instant? 

I'm gonna keep reading a bit on it.

bruce-ricard: I'm super confused.

To convert an instant to a date, you are forced to pass the zone again, which makes complete sense. That I understand.
```
Instant.now().atZone(ZoneId.systemDefault()).getDayOfMonth();
```

But then why is there a `Clock#getZone()` method. I quickly took a look at the library code, the zone doesn't appear to be used for anything in `Instant`. I'm wondering if we can just pass `null` here, as we are not going to use the Zone.

bruce-ricard: I meant passing `null` for the ZoneId.

bruce-ricard: Alright my quick investigation found that `Instant` doesn't use the Clock's zone for anything that we care about, so any System Clock will result in the same behavior.

I would like to do `Clock.system(null)` but there's a null check on the zone parameter, so we can't do that.

With all that information, I lean towards just leaving the code the way it is. Let me know what you think.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
bruce-ricard: >Part of the method is not well-written

What was your first clue?

bruce-ricard: >The throw and rethrow of DataIntegrityViolationException is unnecessary.


I agree, this is pretty crazy.

Are you saying that you want this changed as part of this PR?

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
hsinn0: Java 11 Instant.now() source code:
>     public static Instant now() {
>         return Clock.systemUTC().instant();
>     }
>

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
hsinn0: > > Part of the method is not well-written
> 
> What was your first clue?

I could only see what the old code was doing after expanding the hidden code block in this PR as that code was not part of this PR change.

hsinn0: > > The throw and rethrow of DataIntegrityViolationException is unnecessary.
> 
> I agree, this is pretty crazy.
> 
> Are you saying that you want this changed as part of this PR?

It can be a separate PR if you prefer that way.

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
bruce-ricard: I renamed the class to something more specific.

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
bruce-ricard: Right. That's old code that writes stuff to the DB in the middle of the test.

I would have never written this like that in the first place, but again I don't think that it's that important to fix this now. Another commit/PR can do that.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
swalchemist: Bruce is right, this was already in the catch block.

But I do not see how `(tries++)<=max_tries` is always true, and the refactor that my IntelliJ suggests that the logic move inside the loop, not that it be removed entirely. I'm still concerned about a potential infinite loop, and I'm curious why there isn't a test that catches it.

strehle:
Hi,
can you please at least have a short look at https://github.com/cloudfoundry/uaa/pull/2292 because to me this fix is small enough and we dont need to manipulate the clock with mocking the expiration

strehle (changes requested):
please check
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2291
Missing conditions in code coverage

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
bruce-ricard: I like new lines. Let me know what you think of this.

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
bruce-ricard: OK the test was indeed nasty. I refactored it, and renamed the test.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
bruce-ricard: I'm gonna makes more PRs on this code anyways.

bruce-ricard:
> please check https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2291 Missing conditions in code coverage

This is a false positive. I didn't add any true new code. It's just complaining because it doesn't realize I refactored code.

I'll try to see if I can add coverage regardless.

strehle:
> I'll try to see if I can add coverage regardless.

yes and it shows that there is new complexity with the while(true) and exit if

`
} catch (DataIntegrityViolationException exists) {
                if (attempt>=maxAttempts) throw exists;
            }
`

So is this change really needed?

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
strehle: my advise is , use mockito for the expiration, see https://github.com/cloudfoundry/uaa/pull/2292

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
```diff
@@ -213,10 +225,10 @@ else if (map.get(USER_AUTHENTICATION_UAA_PRINCIPAL)!=null) {
     protected void performExpirationClean() {
         Instant last = lastClean.get();
         //check if we should expire again
-        Instant now = Instant.now();
+        Instant now = Instant.now(clock);
         if (enoughTimeHasPassedSinceLastExpirationClean(last, now)) {
             //avoid concurrent deletes from the same UAA - performance improvement
-            if (lastClean.compareAndSet(last, last.plus(getExpirationTime()))) {
+            if (lastClean.compareAndSet(last, now)) {
```
strehle, line 231: to me this is the only line which need to be fixed, so compareAndSet needs the correct new value which is now

strehle:
only mention this for completeness : Another way to fix this would be to remove the internal cleanup and the setup of a 
task scheduler, e.g. https://www.concretepage.com/spring/spring-task-scheduler in order to regular cleanup this (and maybe the other, e.g. revocable_token) table(s) . We have this cleanup in our fork

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
bruce-ricard: > the refactor that my IntelliJ suggests that the logic move inside the loop, not that it be removed entirely.

That's pretty amazing that our IntelliJs suggest 2 different refactors.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
bruce-ricard: Which is unbelievably confusing, because the `.instant()` call doesn't use the Clock's zone.

bruce-ricard:
> only mention this for completeness : Another way to fix this would be to remove the internal cleanup and the setup of a task scheduler, e.g. https://www.concretepage.com/spring/spring-task-scheduler in order to regular cleanup this (and maybe the other, e.g. revocable_token) table(s) . We have this cleanup in our fork

I agree with that. I don't like the way the cleanup is currently happening. That would be a much larger code change though. If you think that you can easily bring it from your fork, that would be great.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
bruce-ricard: I reverted that change. I'll make another short PR for it.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
hsinn0: > to me this is the only line which need to be fixed

I agree, now especially as the other alternate PR does not have sleep() in the unit test code anymore.  My preference now is going with https://github.com/cloudfoundry/uaa/pull/2292.

strehle:
I agree that we can / should refactor this class, but to get a fixed release, can we agree on a minimal fix ?
https://github.com/cloudfoundry/uaa/pull/2311 because I think there we are all the same opinion ?

strehle:
@bruce-ricard I checked JdbcRevocableTokenProvisioning or JdbcExpiringCodeStore, and then realized, that Filip has introduced a time interface for such situations already and since I would stay with UAA currenty "style" I would prefere to have the TimeService as layer to abstract the time and not introducing a new variant with Clock

See
https://github.com/cloudfoundry/uaa/blob/develop/uaa/src/main/webapp/WEB-INF/spring/oauth-endpoints.xml#L535
or
https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/codestore/JdbcExpiringCodeStore.java#L167-L174

So I vote for replacing clock with timeservice and then of course we can test similar situations because we can mock the time, e.g. https://github.com/cloudfoundry/uaa/blob/develop/server/src/test/java/org/cloudfoundry/identity/uaa/oauth/TokenValidityResolverTest.java#L24-L26

swalchemist:
Are we still going to use the change in this PR?

hsinn0:
> Are we still going to use the change in this PR?

This PR should be closed.  For the immediate fix, we already merged #2292.  For the follow-up refactoring, we have #2315. This was discussed during OSS sync-up meeting on 5/3/2023.

Tallicia:
I'm closing this per discussion above. Please, if it needs to be closed, explain why and close it.

PR #2292, opened 2023-04-19
Minimal Fix auth code cleanup function

strehle:
Set lastClean depending on expirationTime, so that check in performExpirationClean
Adapted the tests so that we wait instead of manually set it to expired.

For short expiring tokens we will call performExpirationClean and cleanup more often, therefore rely on this mechanism for the tests where we expect that performExpirationClean removes entries


strehle:
sonar

https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2292

strehle:
@bruce-ricard Removed the sleep, because use the expiration with Mockito, so that I can manipulate this values also very without extra Sleep and without Clock object
The fix does same as https://github.com/cloudfoundry/uaa/pull/2291, please check

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
```diff
@@ -453,5 +456,12 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
         }
     }
 
+    private static Duration givenMockedExpiration() {
+        Duration durationMock = mock(Duration.class);
+        doReturn(Instant.EPOCH).when(durationMock).subtractFrom(any(Temporal.class));
+        doReturn(Instant.now().plus(Duration.ofMinutes(5))).when(durationMock).addTo(any(Temporal.class));
+        return durationMock;
+    }
+
```
bruce-ricard: Don't you find this complicated and convoluted compared to the simplicity and elegance of injecting the Clock, and setting the time in the test?

Look at [this change](https://github.com/cloudfoundry/uaa/pull/2291/commits/cffb1d6774ba2e673426770a2f5746c1043f1fdb) for example. We were able to get rid of the SQL statements just by pushing the clock forward 4 days.

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
```diff
@@ -206,27 +209,27 @@ void retrieveNonExistentToken() {
...
         int count = 10;
-        store = new UaaTokenStore(dataSource, Duration.ofSeconds(3));
+        store = new UaaTokenStore(dataSource, givenMockedExpiration());
         String lastCode = null;
         for (int i = 0; i < count; i++) {
             lastCode = store.createAuthorizationCode(clientAuthentication);
         }
         assertThat(jdbcTemplate.queryForObject("SELECT count(*) FROM oauth_code", Integer.class), is(count));
 
+        jdbcTemplate.update("UPDATE oauth_code SET expiresat = ?", System.currentTimeMillis() - 60000);
```
bruce-ricard: And you have to put back this statement that you were happy to have deleted.

bruce-ricard (commented):
I largely prefer the option of injecting the Clock. 

I think that it makes the tests a lot easier to understand.

I do not understand what `givenMockedExpiration()` does. It looks like you had to deeply look at the code to figure out what to mock, and how to mock it.

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
strehle: yes because your argument yestday , but here I dont see the cleaner fullfilled

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
strehle: This mock - to me - is excatly what we need , often mock a bean but here we can use mockito also for our own data
https://www.baeldung.com/mockito-annotations, 
Changed to only 3 lines

strehle:
Closed comments before because code changed.

@bruce-ricard you mentioned yesterday this rule: https://deviq.com/principles/boy-scout-rule but to me cleaner is also the principal of "least code as possible" and the real fix is only the line 219 from 
https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java#L219
The problem is compareAndSet and the incorrect 2 parameter which should be now. With this we have the correction.
I hope, you agree.

-
Then then we have to discuss to test the fix. What was missed before was a test see extra PR https://github.com/cloudfoundry/uaa/pull/2293 to ensure that we dont have that many SQL statements, therefore created a test that performs this check. The error message currently is from mockito "Wanted at most 1 time but was 10"
Mocks and mockito framework helps to check what we want and to me that is, ensure that the delete statement is not executed always, see https://www.baeldung.com/mockito-verify #2293 

The enhancement of clock is simply not needed to fix and to test then this fix. It makes the current code not cleaner and therefore in PR https://github.com/cloudfoundry/uaa/pull/2291 my concern is that code gets complexer

bruce-ricard:
> the real fix is only the line 219 from https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java#L219 The problem is compareAndSet and the incorrect 2 parameter which should be now. With this we have the correction. I hope, you agree.

I absolutely agree with that, of course.

> you mentioned yesterday this rule: https://deviq.com/principles/boy-scout-rule but to me cleaner is also the principal of "least code as possible"

I very much disagree with "least code as possible". What we should thrive for is code as clean and readable as possible. Often, that means less code. Not always. 
Injecting a dependency shouldn't be considered a complicated change that adds more useless changes and increases the complexity of the code. I don't often use the word "design pattern", because I often disagree with the patterns, but dependency injection is an excellent pattern that we should almost always follow. In this specific example, the practicality and elegance that it brings to the tests is IMHO a great example of how good dependency injection is.


> The enhancement of clock is simply not needed to fix and to test then this fix. 

I don't get the argument "is not needed". No refactor is ever "needed". But we do them because they make the code better.

> It makes the current code not cleaner and therefore in PR #2291 my concern is that code gets complexer

I argue that it makes the code a lot better. As mentioned above, extracting a dependency doesn't make the code more complex. It makes it simpler, by abstracting away the complexity of the dependency.
The complexity is there regardless. I think that it's incorrect to say that adding a `private final Clock` to the class makes the code more complex. All I'm doing is telling the truth: we do have a dependency on time. The previous code pretends like there isn't.
 
> Then then we have to discuss to test the fix. What was missed before was a test see extra PR #2293 to ensure that we dont have that many SQL statements, therefore created a test that performs this check. The error message currently is from mockito "Wanted at most 1 time but was 10" Mocks and mockito framework helps to check what we want and to me that is, ensure that the delete statement is not executed always, see https://www.baeldung.com/mockito-verify #2293

I like to try to do as few "white box" tests as possible. In this case, I argue that we should validate that the code is no longer in the table after x amount of time. I dislike making raw SQL statements in the tests (as we discussed yesterday), but I think that I dislike even more verifying the mocks. They are both low level, implementation detail oriented, white box tests.

strehle:
@bruce-ricard I checked JdbcRevocableTokenProvisioning or JdbcExpiringCodeStore, and then realized, that Filip has introduced a time interface for such situations already and since I would stay with UAA currenty "style" I would prefere to have the TimeService as layer to abstract the time and not introducing a new variant with Clock

See
https://github.com/cloudfoundry/uaa/blob/develop/uaa/src/main/webapp/WEB-INF/spring/oauth-endpoints.xml#L535
or
https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/codestore/JdbcExpiringCodeStore.java#L167-L174

So I vote for replacing clock with timeservice and then of course we can test similar situations because we can mock the time, e.g. https://github.com/cloudfoundry/uaa/blob/develop/server/src/test/java/org/cloudfoundry/identity/uaa/oauth/TokenValidityResolverTest.java#L24-L26

strehle:
@hsinn0 @Tallicia  Can I merge this? or do you wanna start a review on it? no sure if what we agreed on

hsinn0 (approved):

hsinn0:
> @hsinn0 @Tallicia Can I merge this? or do you wanna start a review on it? no sure if what we agreed on

Yes, merging this is fine.

PR #2293, opened 2023-04-20
TDD for UaaTokenStore: count DB usage in performExpirationClean

strehle:
Add test to check the problem which is that performExpirationClean is called to often


server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
```diff
@@ -332,6 +337,32 @@ void beOAuth2StandardCompliant() {
         assertTrue(code.length() >= 32);
     }
 
+    @Test
+    void testCountingTheExecutedSqlDeleteStatements() throws SQLException {
+        // Given, mocked data source to count how often it is used, call performExpirationClean 10 times
+        DataSource mockedDataSource = mock(DataSource.class);
+        Connection mockedConnection = mock(Connection.class);
+        doReturn(mockedConnection).when(mockedDataSource).getConnection();
+        store = new UaaTokenStore(mockedDataSource);
```
bruce-ricard: I spent 15 minutes staring at this code to understand what it does, but I just can't. I don't understand why you are mocking the `getConnection()` method. I can't even see it used in the `UaaTokenStore` code. 

It looks to me like you are doing some quite low level white box testing. I would much rather try to write high level tests.

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
```diff
@@ -332,6 +337,32 @@ void beOAuth2StandardCompliant() {
...
+        for (int i = 0; i < 9; i++) {
+            // now start to cleanup in loop
+            try {
+                store.performExpirationClean();
+            } catch (Exception sqlException) {
+                // ignore
+            }
+        }
+        // Then
+        verify(mockedConnection, atMost(1)).prepareStatement("delete from oauth_code where expiresat > 0 AND expiresat < ?");
```
bruce-ricard: One idea: it seems to me like we could for example change the signature of the `performExpirationClean` method, to return a boolean. It would be `true` if the cleanup happened, and `false` otherwise. 
That way, the assertions would be much easier to write, wouldn't they?

bruce-ricard (changes requested):
I'm not comfortable accepting this, as I don't understand the code. But thank you very much for writing a test for this code.

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
strehle: see now, just close this comment as it is obsolete

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
strehle: see now, just close this comment as it is obsolete

PR #2311, opened 2023-04-28
Minimal Fix auth code cleanup function

strehle:
Problem. Cleanup should not be called with each Token creation, but periodically to not stress DB too much


strehle:
please use https://github.com/cloudfoundry/uaa/pull/2292

PR #2315, opened 2023-05-05
Refactor: Add Instant to TimeService interface and use TimeService in UaaTokenStore

strehle:
Add TimeService to UaaTokenStore and replace all now usage to timeService object

Allows to change expiration time without changes in DB expiration.

Follow-up from discussion of https://github.com/cloudfoundry/uaa/pull/2291

Sonar results
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2315


server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
```diff
@@ -362,6 +353,20 @@ void testCountingTheExecutedSqlDeleteStatements() throws SQLException {
         assertTrue(after.compareTo(before) < Duration.ofMinutes(5).toNanos());
         // Expect us to call the DB only once within 5 minutes. Check this when using the data source object
         verify(mockedDataSource, atMost(1)).getConnection();
+        // When
+        doReturn(Instant.now().plus(Duration.ofDays(1))).when(timeService).getCurrentInstant();
+        for (int i = 0; i < 10; i++) {
+            performExpirationClean(store);
+            verify(mockedDataSource, atMost(i + 2)).getConnection();
+        }
+    }
```
hsinn0, line 366: It is hard to understand the logic here.  Shouldn't we always expect `atMost(2).getConnection()` here instead?  It seems to me that the duration will be below 5 minutes for the whole for loop iterations.

strehle: yes was a loop before, but now simplified it. important only . If time is gone the DB cleanup is expected again

hsinn0 (approved):

server/src/main/java/org/cloudfoundry/identity/uaa/util/TimeService.java
```diff
@@ -16,12 +16,15 @@
...
 public interface TimeService {
     default long getCurrentTimeMillis() {
-        return System.currentTimeMillis();
+        return getCurrentInstant().toEpochMilli();
     }
 
-    default Date getCurrentDate() { return new Date(getCurrentTimeMillis()); }
+    default Date getCurrentDate() { return Date.from(getCurrentInstant()); }
+
+    default Instant getCurrentInstant() { return Instant.now(); }
```
bruce-ricard, line 29: I would like to understand why you think it is better to reuse this interface. Now it's an interface with 3 methods, for which every caller will most likely use exactly one method, making the 2 other ones useless.

IMO these should be 3 different interfaces. The one returning the `Instant` being the one we want to use in the future, and the 2 other ones being marked as `@Deprecated` in the short term.

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/ApprovalServiceTest.java
```diff
@@ -68,7 +71,7 @@ public void ensureRequiredApprovals_throwsWhenApprovalsExpired() {
         approval.setScope("foo.read");
         approval.setStatus(Approval.ApprovalStatus.APPROVED);
         approval.setExpiresAt(new Date(approvalExpiry));
-        when(timeService.getCurrentTimeMillis()).thenReturn(approvalExpiry + 1L);
+        when(timeService.getCurrentInstant()).thenReturn(Instant.EPOCH.plus(approvalExpiry + 1L, ChronoUnit.MILLIS));
```
bruce-ricard: There is another `plus()` method which is IMO is better: it adds [a duration](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#plus-java.time.Duration-) and looks like this:

```
time.plus(Duration.ofMillis(42))
```

but I notice that there is an even better one when you want to add a specific simple amount of time, you can directly call [plusMillis(42)](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#plusMillis-long-). And the code becomes

```
Instant.EPOCH.plusMillis(approvalExpiry + 1L)
```

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
```diff
@@ -200,7 +201,7 @@ void retrieveToken() {
     void retrieveExpiredToken() {
         String code = store.createAuthorizationCode(clientAuthentication);
         assertThat(jdbcTemplate.queryForObject("SELECT count(*) FROM oauth_code WHERE code = ?", new Object[]{code}, Integer.class), is(1));
-        jdbcTemplate.update("update oauth_code set expiresat = 1");
+        doReturn(Instant.now().plus(UaaTokenStore.DEFAULT_EXPIRATION_TIME)).when(timeService).getCurrentInstant();
```
bruce-ricard, line 204: I see that sometimes you mock by doing 

```
when(something).return(something else)
```

and other times like here, you do

```
doReturn(value).when(something)
```

Is there a reason you switch the order?

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
```diff
@@ -482,10 +487,10 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
...
     }
 
-    private static Duration givenMockedExpiration() {
-        Duration durationMock = mock(Duration.class);
-        doReturn(Instant.now().plus(UaaTokenStore.DEFAULT_EXPIRATION_TIME)).when(durationMock).addTo(any(Temporal.class));
-        return durationMock;
+    private static TimeService givenMockedTime() {
+        TimeServiceImpl timeService = mock(TimeServiceImpl.class);
+        doReturn(Instant.now()).when(timeService).getCurrentInstant();
+        return timeService;
```
bruce-ricard, line 497: I don't understand the point of this method. It now just returns a normal TimeService that returns the actual current time.

bruce-ricard (changes requested):
I don't understand the reason why it's better to reuse another interface that is already used by the code.

IMO this is very much the opposite of what dependency injection needs to be: you want to inject what you need, not what is used elsewhere in the code.
Each class or file needs to create their own interfaces specifying what they need, what their dependencies are. Even if 2 classes end up having the exact same dependency at a give point in time, we should NOT reuse the interface, because it creates a binding between the 2 users. If some one class now requires the dependency to be more powerful, that same power will be given to the other class, which doesn't need or want it.

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
swalchemist: The way they work is different. The doReturn() version is required in some situations, like using a spy. There are some cases where they do the same thing.

server/src/main/java/org/cloudfoundry/identity/uaa/util/TimeService.java
strehle: reverted other changes

PR #2326, opened 2023-05-16
fix: Correcting mysql 5 backup to mysql 8 restore to align collation …

Tallicia:
…shared in both mysql 5 and 8.

[#184908751]


hsinn0 (approved):

strehle (approved):
cannot really judge but ok for us

MrDXY:
Please hold on to backporting this pr to v74, as we have discovered that this fix may introduce new issues for tkgi.

PR #2339, opened 2023-05-19
fix: cve resolution for 74-5 and consistency for internal subproject …

Tallicia:
…build dependencies. 

Directing subproject gradle builds to use the latest working versions of dependency libraries with identified CVEs.

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -1,6 +1,7 @@
 // Versions we're overriding from the Spring Boot Bom
 ext["flyway.version"] = "7.15.0" // flyway 8+ drops support for mysql 5.7
 ext["mariadb.version"] = "2.7.7" // Bumping to v3 breaks some pipeline jobs (and compatibility with Amazon Aurora MySQL), so pinning to v2 for now. v2 (current version) is stable and will be supported until about September 2025 (https://mariadb.com/kb/en/about-mariadb-connector-j/).
+ext["snakeyaml.version"] = "1.33"
 
 def versions = [:]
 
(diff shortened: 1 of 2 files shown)


Tallicia:
@peterhaochen47 Thank you for the feedback! :) Refactor is included now.

dependencies.gradle
```diff
@@ -16,6 +18,9 @@ versions.springSecuritySamlVersion = "1.0.10.RELEASE" // this lib has reached EO
 versions.tomcatVersion = "9.0.75"
 versions.statsdClientVersion = "3.1.0" // statsd is not included in spring-boot 2.2 and later, so we need to specify the dependency explicitly
 versions.junitVersion = "4.13.2" // Needed to support junit 4 with spring-boot 2.4
+versions.snakeyamlVersion = "1.33" // Needed to resolve CVEs in internal spring-boot 2.7.12 inclusion of snakeyaml
```
peterhaochen47: Why does this (plus the usage of this var below) need to be defined in addition to `ext["snakeyaml.version"] = "1.33"`. I've tried to remove this var & did not notice any difference in our final dependencies (as shown in intellij) after removing.

peterhaochen47: Other than ^, LGTM

Tallicia: ~~It is redundant in the this cve only set of 4 libraries. I'll remove it.~~

Tallicia: Correction, removing it results in a build failure: 
`/Users/yalicia/src/workspace/uaa/server/src/main/java/org/cloudfoundry/identity/uaa/impl/config/UaaConfiguration.java:21: error: package org.yaml.snakeyaml does not exist
import org.yaml.snakeyaml.TypeDescription;
                         ^`

peterhaochen47: I've tried removing this var `versions.snakeyamlVersion = "1.33"` and its usage (aka changing `libraries.snakeyaml = "org.yaml:snakeyaml:${versions.snakeyamlVersion}"` back to libraries.snakeyaml = "org.yaml:snakeyaml" which is the existing code) and it seems to be work. Did you experiment with the same thing?

peterhaochen47: I've tried removing this var `versions.snakeyamlVersion = "1.33"` and its usage (aka changing `libraries.snakeyaml = "org.yaml:snakeyaml:${versions.snakeyamlVersion}"` back to `libraries.snakeyaml = "org.yaml:snakeyaml"` which is the existing code) and it seems to be work (with `./gradlew build -x test`) and does not result in warnings in "Dependency Analyzer" tool in intellij. Did you experiment with the same thing?

Tallicia: The redundant snakeyaml versioning has been removed.

peterhaochen47 (approved):

PR #2341, opened 2023-05-19
Pr/bruce/refactor while condition

bruce-ricard:


server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
```diff
@@ -117,10 +118,9 @@ public String createAuthorizationCode(OAuth2Authentication authentication) {
                 }
                 return code;
             } catch (DataIntegrityViolationException exists) {
-                if (tries>=max_tries) throw exists;
+                if (attempt>=maxAttempts) throw exists;
```
hsinn0, line 121: The throw of `DataIntegrityViolationException` and catch is unnecessary and inefficeint.  Assuming no other lines above in this try block throws `DataIntegrityViolationException`, it can be all done in the same if block, that is:
```
if (updated==0 && attemps>maxAttempt) {
    throw new DataIntegrityViolationException("[oauth_code] Failed to insert code. Result was 0");
}
```

hsinn0: Actually, you need to do more than that as you also need to `return code` when `update != 0` but you get the idea.

swalchemist: Agreed, the catch and rethrow could be simplified. It wasn't added by the PR, though.

hsinn0: I take it back.  The `template.update()` call throws `DataAccessException`, which is a super class of `DataIntegrityViolationException`, so it is possible to get  `DataIntegrityViolationException` thrown from that call, too.  With that said, I am approving this PR as it is.

hsinn0 (approved):

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
hsinn0: > It wasn't added by the PR, though.

Right, I do know that.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
```diff
@@ -100,7 +100,8 @@ public String createAuthorizationCode(OAuth2Authentication authentication) {
         performExpirationClean();
         JdbcTemplate template = new JdbcTemplate(dataSource);
         int tries = 0;
-        while ((tries++)<=max_tries) {
+        while (true) {
+            tries++;
```
swalchemist: Interesting. I'm able to comment on a line of code in an individual commit. Here I'm seeing the name of the variable as "tries", which was later refactored.

swalchemist (approved):
I had to really study the original code to convince myself that this refactor doesn't change behavior, and now I'm convinced that it does not (I hope we have very solid unit test coverage). The fact that I had to review it so carefully is a testament to why this refactor is needed.

PR #2342, opened 2023-05-19
Fix cleanup multiple parallel runs issue

bruce-ricard:
The cleanup now cannot be run multiple times in parallel due to the new Mutex protection.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
@@ -97,7 +97,7 @@ public class UaaTokenStore implements AuthorizationCodeServices {
     @Override
     public String createAuthorizationCode(OAuth2Authentication authentication) {
         final int maxAttempts = 3;
-        performExpirationClean();
+        performExpirationCleanIfEnoughTimeHasElapsed();
         JdbcTemplate template = new JdbcTemplate(dataSource);
         int attempt = 0;
         while (true) {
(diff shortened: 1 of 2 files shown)


server/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStoreTests.java
```diff
@@ -350,7 +350,7 @@ void testCountingTheExecutedSqlDeleteStatements() throws SQLException {
         // When
         for (int i = 0; i < 10; i++) {
             try {
-                store.performExpirationClean();
+                store.performExpirationCleanIfEnoughTimeHasElapsed();
```
swalchemist, line 353: Are you confident that there is sufficient test coverage for the parallel use case?

Some questionable things in this test - 
 - Doesn't seem to call the method concurrently.
 - With "atMost(1)" it may be non-deterministic. Why not expect exactly 1 call?
 - Could it really run successfully for close to 5 minutes?
 - Why ignore SQL exceptions?

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
```diff
@@ -210,23 +212,28 @@ else if (map.get(USER_AUTHENTICATION_UAA_PRINCIPAL)!=null) {
...
+    protected void performExpirationCleanIfEnoughTimeHasElapsed() {
+        if (cleanMutex.tryAcquire()) {
+            //check if we should expire again
+            Instant now = Instant.now();
+            if (enoughTimeHasPassedSinceLastExpirationClean(lastClean, now)) {
+                //avoid concurrent deletes from the same UAA - performance improvement
+                lastClean = now;
+                actuallyPerformExpirationClean(now);
             }
+            cleanMutex.release();
```
hsinn0: `cleanMutex.release()` needs to be inside `finally` because it won't be executed when `actuallyPerformExpirationClean()` throws `DataAccessException`, which also should be covered by a unit test case.

hsinn0 (changes requested):

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenStore.java
bruce-ricard: Yes I was just thinking about putting in a `finally` block.

hsinn0 (approved):

strehle (approved):

PR #2343, opened 2023-05-23
Support json key set in tokenKey

strehle:
Why: OIDC IdP trust can be verified using tokenKeyUrl or tokenKey In case of tokenKeyUrl the jwks_uri is called to fetch all current keys in order to verify the JWT signature. In case of tokenKey there is currently only a single key possible. If the IdP has a dynamic key rotation, then trust breaks.


torsten-sap:
@adrianhoelzl-sap @tack-sap Can you review this change? It allows to paste the configure the whole content of the tokens_keys endpoint as input for attribute "tokenKey", meaning potentially also multiple keys as the structure is JSON. Thanks!

tack-sap (approved):

PR #2347, opened 2023-05-30
fix: make kill more reliable

swalchemist:
* I've been testing this version of the script for a few months and it works a lot more reliably.


Tallicia:
Why is this not included as part of the gradle terminiation?

hsinn0 (approved):

strehle (approved):

PR #2362, opened 2023-06-12
Dependency refactoring

strehle:
Requires first https://github.com/cloudfoundry/uaa/pull/2363

- overwrite all selenium library versions
- use tomcat version and allow dependabot update


Tallicia (approved):
LGTM

PR #2372, opened 2023-06-15
Delete all user group members if user is deleted

### LINKED ISSUE #2368: Group member is not removed if user is deleted

I am deploying the UAA

- locally only using gradlew
- using cf-release
- using cf-deployment
- other (please explain)

### What did you do?

Steps to reproduce:
1. Create a user
2. Assign this user to a group
3. Delete the user
Result the user is still member of the group

User should be deleted from group members

### What did you see instead?

strehle:
Alternative Fix for issue 2398

TDD , test first


strehle:
dicussed yesterday

strehle:
@torsten-sap @tack-sap any thoughts from SAP side to fix issue https://github.com/cloudfoundry/uaa/issues/2368 with this PR or the other one? this here would be without disruption

tack-sap:
> @torsten-sap @tack-sap any thoughts from SAP side to fix issue #2368 with this PR or the other one? this here would be without disruption

I would prefer this one here. In the other PR I see risks of breaking existing and important flows.
This PR just fixes the issue, which does not bear any risk.
While the current usage of the "origin" (that it can be different from the origin of the user, and i.e. have the value "uaa" for non-internal users) in the member is not consistent to what's documented, it is still required by other components - not only inside SAP, but also in CF itself.

torsten-sap (approved):

Issue #2235, opened 2023-03-10
Support private_key_jwt from https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication


strehle:
### What version of UAA are you running?

What output do you see from `curl <YOUR_UAA>/info -H'Accept: application/json'`?
> Develop / 76.x.x

### How are you deploying the UAA?

I am deploying the UAA

- locally only using gradlew
- using a bosh release I downloaded from bosh.io
- using cf-release
- using cf-deployment
- as part of a commercial Cloud Foundry distribution
- other (please explain)

### What did you do?

Authenticate OAuth clients without secrets. Public Flow does not solve all cases, e.g. client_credential or password grant .
Private_key_jwt and https://www.rfc-editor.org/rfc/rfc8705 would solve that issue.

Private_key_jwt is part of OIDC standard , e.g.
https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication

And therefore this is Expected to be Part of UAA

### What did you expect to see? What goal are you trying to achieve with the UAA?

Private_key_jwt is part of OIDC standard , e.g.
https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication

### What did you see instead?

secret based authentication only

swalchemist:
Was this fixed by #2385?

strehle:
> Was this fixed by #2385?

No but https://github.com/cloudfoundry/uaa/pull/2385 is a preparation for it

strehle:
Next stepp is, that oauth_client_details table will be extended with client_keys  or client_key_config so that an admin can upload trusted keys. The pure validation is a smaller step. 
Tokens with private_key_jwt will have this authentication method in the token.

strehle:
@Tallicia @bruce-ricard Started to work on this. I created https://github.com/cloudfoundry/uaa/pull/2433 and I would like to discuss with you in next sync about the possible options to add the key configuration into DB table oauth_client_details.

1. Add an extra column e.g. private_key_jwt_config
2. Migrate additonal_information into text (to have capacity for it) and add the keys in map additional_information.

If you agree / vote for 1. then this PR https://github.com/cloudfoundry/uaa/pull/2433 is a pre-requisit. We need then to decouple from spring security oauth2 classes

Tallicia:
I have added this to the agenda.

strehle:
Added private_key_jwt into my spring demo app

[...]

PR #2385, opened 2023-06-22
feature: Store client authentication method in JWT

strehle:
Why: UAA historical supported only secret based client authentication, so no need to have this information on client side. 
UAA supports until now client_secret_basic and client_secret_post from section 9 of standard: 
See: https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication

Now there is the public usage, later private_key_jwt should be supported. Maybe then tls_client_auth.

Because of the support of other client authentication methods, the information might be needed or at least helpful for clients.

To be compatible. We add this information only if another (new) authentication was used. New means public and later private_key_jwt. So the new claim in token will only appear if a new method was used.

This PR should serve for solving
- https://github.com/cloudfoundry/uaa/issues/2138
->  need to know if public
- https://github.com/cloudfoundry/uaa/issues/2235
-> clients should see if private_key_jwt


server/src/main/java/org/cloudfoundry/identity/uaa/authentication/ClientDetailsAuthenticationProvider.java
```diff
@@ -61,6 +61,7 @@ protected void additionalAuthenticationChecks(UserDetails userDetails, UsernameP
                     Object allowPublic = uaaClient.getAdditionalInformation().get(ClientConstants.ALLOW_PUBLIC);
                     if (allowPublic instanceof String && Boolean.TRUE.toString().equalsIgnoreCase((String)allowPublic) ||
                         allowPublic instanceof Boolean && Boolean.TRUE.equals(allowPublic)) {
+                        ((UaaAuthenticationDetails) authentication.getDetails()).setAuthenticationMethod("none");
```
adrianhoelzl-sap, line 64: Is it intended that there is no dedicated value for the regular client ID/client secret authentication?

uaa/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServicesTests.java
```diff
@@ -275,9 +279,16 @@ void happyCase() {
             UaaUser uaaUser = jdbcUaaUserDatabase.retrieveUserByName("admin", "uaa");
             refreshToken = refreshTokenCreator.createRefreshToken(uaaUser, refreshTokenRequestData, null);
             assertThat(refreshToken, is(notNullValue()));
+            OAuth2Authentication authentication = mock(OAuth2Authentication.class);
+            SecurityContextHolder.getContext().setAuthentication(authentication);
```
adrianhoelzl-sap, line 289: Consider clearing the security context after the test case

uaa/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServicesTests.java
```diff
@@ -275,9 +279,16 @@ void happyCase() {
...
+            OAuth2Request auth2Request = mock(OAuth2Request.class);
+            when(authentication.getOAuth2Request()).thenReturn(auth2Request);
+            when(auth2Request.getExtensions()).thenReturn(Map.of(ClaimConstants.CLIENT_AUTH_METHOD, "client_secret_basic"));
             OAuth2AccessToken refreshedToken = tokenServices.refreshAccessToken(this.refreshToken.getValue(), new TokenRequest(new HashMap<>(), "jku_test", Lists.newArrayList("openid", "user_attributes"), GRANT_TYPE_REFRESH_TOKEN));
 
             assertThat(refreshedToken, is(notNullValue()));
+            Map<String, Object> claims = UaaTokenUtils.getClaims(refreshedToken.getValue());
+            assertThat(claims, hasKey(ClaimConstants.CLIENT_AUTH_METHOD));
```
adrianhoelzl-sap: Consider also checking the value of the key.

server/src/main/java/org/cloudfoundry/identity/uaa/authentication/ClientDetailsAuthenticationProvider.java
adrianhoelzl-sap: In your tests, we found the value `client_secret_basic` for such cases

strehle: See into of the PR , I want set this information only if none and later private_key_jwt, because they are new.
With this all existing UAA installations wont change or see the new claim.
Only if they switch to new behaviour then they should see the new claim in the token.

I will use this information in next PR in order to allow public clients for refresh token, but therefore I need to know what was the authentication in the token before.

uaa/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServicesTests.java
strehle: done

adrianhoelzl-sap (approved):

PR #2197, opened 2023-02-14
Bump Gradle to 8.0.1

strehle:
./gradlew wrapper --gradle-version=8.0.1

In addition update plugins

PR #2386, opened 2023-06-22
Build fix

strehle:
Gradle should use its plugins from gradle
ascii-doctor is no long maintained

@Tallicia only a small fix, but maybe you take all build files from develop and also upgrade gradle.... The code can stay as it is

e.g. https://github.com/cloudfoundry/uaa/pull/2197/files


Tallicia (approved):
Looks good, and thanks for the quick fix!

PR #2405, opened 2023-07-15
uaa-ci: use RS256 key as default

strehle:
Why: HS256 is not used in any CF production landscape and we should use locally what we support in production We may, should think about a list of allowed algorithms and here omit HSxxx algorithms


adrianhoelzl-sap (approved):

Tallicia:
I am not aware of any other testing pipelines this could break, I will ask the team if they know of any and mention them here if they need to prepare to update them.

hsinn0 (approved):

strehle:
merging and check if ci runs into issues

PR #2406, opened 2023-07-17
Change default of refresh token format

strehle:
Default should be opaque, at least the default.
History about 

https://github.com/cloudfoundry/uaa-release/issues/93 

Some tests rely on JWT for refresh, therefore added for these tests again jwt.token.refresh.format=jwt


Tallicia:
Need discuss deprecation timeline and notificaiton.

hsinn0:
> Need discuss deprecation timeline and notificaiton.

Can someone help me to understand this statement? Was there some offline discussion? What deprecation are we talking about?

hsinn0 (commented):
It is my understanding this change affects the UAA API doc, `config.tokenPolicy.refreshTokenFormat` at https://docs.cloudfoundry.org/api/uaa/version/76.16.0/index.html#creating-an-identity-zone for example. Do we have plan to also update the doc?

strehle:
> It is my understanding this change affects the UAA API doc, `config.tokenPolicy.refreshTokenFormat` at https://docs.cloudfoundry.org/api/uaa/version/76.16.0/index.html#creating-an-identity-zone for example. Do we have plan to also update the doc?

Thanks , missed that. 
I changed more defaults - internally and adopted the tests, which rely on JWT in refresh token, so that there a JWT comes, but the rest of the tests should not have problems with the opaque token

So this PR simply changes the default but if some one wants JWT for refresh tokens, then this can be configured, now with an extra step

hsinn0 (approved):

PR #2422, opened 2023-07-25
Delete unused script & dockerfile

peterhaochen47:
- This script (scripts/docker_run) and this dockerfile (ci/image/*) do not appear to be used or maintained anymore.


strehle:
The Dockerfile at least was helpful in past because if pipeline issues occurred I could there steps with it, e.g. Download of java or chromedriver etc....
Can we have the current Dockerfile in an example directory at least ?

peterhaochen47:
Hi @strehle, I see. 
This docker image is not the docker image used in CIs, though it might look similar to the docker images used in the CIs because it was likely based on them when it was created. As this image is not maintained, it has diverged quite a bit from the CI docker images. So its usefulness as an approximation of what's happening in CI is diminishing. So I still advocate for the deletion of these files in this PR, as they are no longer maintained/kept up to date. If you wish to see these files for context, they are still preserved in git history.

We do try to fix CI errors ASAP. But if you wish to participate in CI fixing, that would be great! And I do recognize the problem that you don't have access to the CI repo (including the actual docker images); we should try to think about how to solve that problem (potentially by extracting credentials & any proprietary info from the CI repo, and making the CI repo public; or by moving the dockerfiles somewhere public).

strehle (approved):
Ok

PR #2423, opened 2023-07-27
fix: expired X509 certificates should be ignored for JWT usage

strehle:
An expired X509 is ignored with this fix and therefore internally not used anymore. The expired certificate won't appear in /token_key, /token_keys anymore and private_key_jwt is not done with x5t, that the X509 as trust anchor is not possible anymore.

The behavior is not really changed with this PR, but no exception should be thrown in runtime. Before this PR, an Exception and with that HTTP 5xx was thrown


strehle:
sonar green

https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2423

tack-sap (approved):

PR #2432, opened 2023-08-01
fix: update k8s to go 1.20

Tallicia:
updated and has updated with go mod tidy.

[#185619304]

@peterhaochen47

### DIFF
--- a/k8s/go.mod
+++ b/k8s/go.mod
@@ -1,6 +1,6 @@
 module github.com/cloudfoundry/uaa
 
-go 1.18
+go 1.20
 
 require (
 	github.com/onsi/ginkgo v1.16.5


hsinn0:
The change itself, which has the one-line change in `go.mod` is as expected. However, there is something wrong with the commit message.

> has updated with go mod vendor

`go mod vendor` would have created vendor folder as it did not exist yet. However, `vendor` folder is not a part of the commit and it should not be as we do not need/want that for this go module. Just doing `go mod tidy` was enough for this change. I suggest to remove "go mod vendor" from the commit message.
See https://go.dev/ref/mod#vendoring.

hsinn0 (changes requested):
See the comment regarding the commit message.

Tallicia:
@hsinn0 The requested changes have been made.

hsinn0:
I see the PR comment updated. However, the commit message was not. Isn't the commit message the one added as commit message after "Rebase and merge"?

hsinn0 (approved):

PR #2433, opened 2023-08-07
refactor: prepare for private_key_jwt in oauth_client_details

strehle:
BaseClientDetails from spring security oauth2 cannot be changed, therefore move to UaaClientDetails for client details load

Add new column : client_jwt_config

The content of client_jwt_config will be JSON . The PR https://github.com/cloudfoundry/uaa/pull/2449 will make use of this 

The client jwt authentication is based on JsonWebKey trust.  See https://datatracker.ietf.org/doc/html/rfc7517
The internal store can contain either.

* jwks:  similar to that what you see in /token_key or /token_keys of your own UAA but the real JSON content.
* jwks_uri: is the URL to a remove jwks_uri e.g. http://localhost:8080/uaa/token_key(s)

The idea is, that with such a trust you can now configure in your zone1 a client with jwks_uri pointing to system zone, e.g.  http://localhost:8080/uaa/token_keys . Now you define a OIDC identity provider object with jwtClientAuthentication: true and your set in tokenKeyUrl the other jwks_uri, e.g.  http://zone.localhost:8080/uaa/token_keys

Now you have jwt based client authentication between zones: uaa and zone. 
You can also add other clients in customer zone where client trust is set to http://localhost:8080/uaa/token_keys (system uaa zone) and then you do other calls but the client in the customer zone always trusts your system zone, even if you update your own JWT signing keys.


strehle:
@Tallicia @bruce-ricard can you please review this PR ?
In meeting we have decided to go for option 1, (own DB column) but then BaseClientDetails (from spring securit oauth2) cannot be used anymore but I have to move to own Class as abstraction for DB table oauth_client_details Is the Name UaaClientDetails Ok with you ?

server/src/main/java/org/cloudfoundry/identity/uaa/client/UaaClientDetails.java
```diff
@@ -1,23 +1,76 @@
...
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        if (o instanceof UaaClientDetails) {
+            UaaClientDetails uaaClientDetails = (UaaClientDetails) o;
+            return Objects.equals(clientJwtConfig, uaaClientDetails.clientJwtConfig);
```
adrianhoelzl-sap: `equals` and `hashCode` have different behavior here. Consider adding checks for the properties inherited from the super-class.

server/src/main/java/org/cloudfoundry/identity/uaa/zone/MultitenantJdbcClientDetailsService.java
```diff
@@ -80,7 +80,7 @@ public class MultitenantJdbcClientDetailsService extends MultitenantClientServic
 
     private static final String DEFAULT_INSERT_STATEMENT =
             "insert into oauth_client_details (" + CLIENT_FIELDS
-                    + ", client_id, identity_zone_id, created_by) values (?,?,?,?,?,?,?,?,?,?,?,?,?,?,?)";
+                    + ", client_id, identity_zone_id, created_by) values (?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?)";
 
     private static final String DEFAULT_UPDATE_STATEMENT =
```
adrianhoelzl-sap, line 85: Is there a reason why the new field cannot be changed through the regular update operation?

server/src/test/java/org/cloudfoundry/identity/uaa/client/UaaClientDetailsTest.java
```diff
@@ -63,6 +66,35 @@ void copiesAdditionalInformation() {
...
+        @Test
+        void testEquals() {
+          UaaClientDetails copy = new UaaClientDetails(testClient);
+          UaaClientDetails copy2 = new UaaClientDetails(testClient);
+          copy.setClientJwtConfig("test");
+          assertNotEquals(copy, copy2);
+          assertNotEquals(copy, new UaaClientDetails());
+          copy.setClientJwtConfig(null);
+          assertEquals(copy, copy2);
+        }
```
adrianhoelzl-sap, line 88: Please also add a test case that checks different properties from the super-class.

server/src/test/java/org/cloudfoundry/identity/uaa/client/UaaClientDetailsTest.java
```diff
@@ -63,6 +66,35 @@ void copiesAdditionalInformation() {
...
+
+        @Test
+        void testHashCode() {
+          UaaClientDetails copy = new UaaClientDetails(testClient);
+          UaaClientDetails copy2 = new UaaClientDetails(testClient.getClientId(), "",
```
adrianhoelzl-sap, line 93: Is there a reason why this constructor is used instead of the one used one line above?

server/src/test/java/org/cloudfoundry/identity/uaa/zone/MultitenantJdbcClientDetailsServiceTests.java
```diff
@@ -294,6 +295,7 @@ void loadingClientIdWithSingleDetails() {
         jdbcTemplate.update(INSERT_SQL,
                 "clientIdWithSingleDetails",
                 "mySecret",
+                "myClientJwtConfig",
```
adrianhoelzl-sap, line 298: Please cast the result in line 308 to `UaaClientDetails` and then check whether the value is set correctly

server/src/test/java/org/cloudfoundry/identity/uaa/client/UaaClientDetailsTest.java
strehle: because we have 2 constructor, but same content. Simply to show that Independent from constructor only the content is compared

server/src/test/java/org/cloudfoundry/identity/uaa/client/UaaClientDetailsTest.java
strehle: done, refactored the equals check

server/src/main/java/org/cloudfoundry/identity/uaa/zone/MultitenantJdbcClientDetailsService.java
strehle: the standard update only update the plain (primitive / or single value) fields. Others like secret or here the jwt trust are updated with extra methods, therefore extra SQL statements. Same as for secrets change

server/src/main/java/org/cloudfoundry/identity/uaa/client/UaaClientDetails.java
strehle: changed,

server/src/test/java/org/cloudfoundry/identity/uaa/zone/MultitenantJdbcClientDetailsServiceTests.java
strehle: done

hsinn0 (approved):

tack-sap (approved):

PR #2434, opened 2023-08-08
fix: Dependabot can't authenticate to the private package registry ht…

hsinn0:
…tps://repo.spring.io

- We get errors as follows at https://github.com/cloudfoundry/uaa/network/updates/705768660:
```
Dependabot can't authenticate to a private package registry
The following private package registry was used and caused the update to fail: https://repo.spring.io/release.
```
- Stop using `repo.spring.io` for dependencies. See https://spring.io/blog/2022/12/14/notice-of-permissions-changes-to-repo-spring-io-january-2023.

[#185768114]

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,9 +12,6 @@ buildscript {
         maven {
             url("https://plugins.gradle.org/m2/")
         }
-        maven {
-            url("https://repo.spring.io/plugins-release")
-        }
     }
 
     dependencies {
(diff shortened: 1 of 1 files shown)


swalchemist (approved):
Ok with me. Need review from SAP too?

hsinn0:
> Need review from SAP too?

Nah, especially because I think they are still on vacation.

PR #2436, opened 2023-08-09
Bump hsqldb version 2.7.1 to 2.7.2

strehle:
hsqldb.version


Tallicia:
Shouldn't dependabot be catching this new version?

strehle:
> Shouldn't dependabot be catching this new version?

yes I thought that would be the case but for versions here it does not work always

PR #2439, opened 2023-08-10
feature: allow setting SameSite on X-Uaa-Csrf cookie

mikeroda:


strehle:
@mikeroda can you please describe how you want use this feature. I see a test, but there is no external configuration option, so to me you can only use this if you have our own UAA as fork but call the api in your forked UAA.

strehle (approved):
ok for me

mikeroda:
@strehle that is correct. This is useful for projects that fork UAA and call the API to change the SameSite configuration on the X-Uaa-Csrf cookie. Might seem trivial but is a lot more convenient than extending the class to add the setter method.

strehle:
> @strehle that is correct. This is useful for projects that fork UAA and call the API to change the SameSite configuration on the X-Uaa-Csrf cookie. Might seem trivial but is a lot more convenient than extending the class to add the setter method.

Ok, I am fine with

PR #2440, opened 2023-08-10
obsolete now: Increase the client additional_information length limit 

strehle:
- Change the database data type from varchar to text type
- Ensure enough space for JwkSet configuration for private_key_jwt


strehle:
About private_key_jwt support. see https://github.com/cloudfoundry/uaa/issues/2235#issuecomment-1669167440, this would be the option 2.

strehle:
This PR was for option 2 but we should proceed with 1, -> new column in oauth_client_details

strehle:
replaced with https://github.com/cloudfoundry/uaa/pull/2433

PR #2447, opened 2023-08-18
feature: add client_auth_method=empty into tokens for clients with empty secret

strehle:
add client_auth_method=empty into tokens for clients with empty secret

e.g. cf client.
This allows clients to see, whats really used.
According to https://oauth.net/2.1/ the empty secret can be treated as public usage

Similar PR https://github.com/cloudfoundry/uaa/pull/2435 uses public for non-existing secrets and empty secrets. Here we treat empty secret as own method


strehle:
not need anymore, thus no option anymore

PR #2449, opened 2023-08-22
feature: add persistence support for private_key_jwt client authentication

strehle:
More details in https://github.com/cloudfoundry/uaa/pull/2433 as this PR include https://github.com/cloudfoundry/uaa/pull/2433. -> because to have smaller review packages

Allow to setup jwks_uri and jwks, similar to OIDC proxy mode with tokenKeyUrl and tokenKey. The private_key_jwt metadata is stored in additional_information (could be switched to own column)

The setup can be done from REST and yaml.


model/src/main/java/org/cloudfoundry/identity/uaa/oauth/client/ClientJwtChangeRequest.java
```diff
@@ -0,0 +1,81 @@
...
+
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import static org.cloudfoundry.identity.uaa.oauth.client.ClientJwtChangeRequest.ChangeMode.ADD;
+
+/**
+ */
```
hsinn0: Can we have actual Javadoc for the class here? Or just get rid of `/** */` if you don't think that we need any class doc?

model/src/main/java/org/cloudfoundry/identity/uaa/oauth/client/ClientJwtChangeRequest.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    public String getKeyId() { return keyId;}
+
+    public void setKeyId(String keyId) {
+        this.keyId = keyId;
+    }
+
+    public String getKey() {
+        return jsonWebKeyUri != null ? jsonWebKeyUri : jsonWebKeySet;
```
hsinn0, line 84: So if both `jsonWebKeyUri` and `jsonWebKeySet` are set, the later is ignored. Wouldn't it create some confusion?

server/src/main/java/org/cloudfoundry/identity/uaa/client/ClientAdminBootstrap.java
```diff
@@ -204,11 +205,26 @@ private void addNewClients() {
...
+                if (keyConfig != null && keyConfig.getCleanString() != null) {
+                    keyConfig.writeValue(client);
+                }
+            } else if (map.get("jwks") instanceof String) {
+                String jwks = (String) map.get("jwks");
+                ClientJwtConfiguration keyConfig = ClientJwtConfiguration.parse(null, jwks);
+                if (keyConfig != null && keyConfig.getCleanString() != null) {
+                    keyConfig.writeValue(client);
+                }
+            }
```
hsinn0, line 224: If  `map.get("jwks_uri")` or `map.get("jwks")` is not `instanceof` String, that would be unexpected situation. Here, the code just ignores the unexpected situation. Should it not be considered as exception instead?

server/src/main/java/org/cloudfoundry/identity/uaa/client/ClientAdminEndpoints.java
```diff
@@ -535,6 +536,53 @@ public ActionResult changeSecret(@PathVariable String client_id, @RequestBody Se
         return result;
     }
 
+    @RequestMapping(value = "/oauth/clients/{client_id}/clientjwt", method = RequestMethod.PUT)
+    @ResponseBody
+    public ActionResult changeClientJwt(@PathVariable String client_id, @RequestBody ClientJwtChangeRequest change) {
```
hsinn0, line 549: This appears to be a new endpoint. Do we have an API doc for the new endpoint?

server/src/main/java/org/cloudfoundry/identity/uaa/client/ClientAdminEndpointsValidator.java
```diff
@@ -245,6 +246,19 @@ public ClientDetails validate(ClientDetails prototype, boolean create, boolean c
...
+
+            if (prototype instanceof ClientDetailsCreation) {
+                ClientDetailsCreation clientDetailsCreation = (ClientDetailsCreation) prototype;
+                if (StringUtils.hasText(clientDetailsCreation.getJsonWebKeyUri()) || StringUtils.hasText(clientDetailsCreation.getJsonWebKeySet())) {
+                    ClientJwtConfiguration clientJwtConfiguration = ClientJwtConfiguration.parse(clientDetailsCreation.getJsonWebKeyUri(),
+                        clientDetailsCreation.getJsonWebKeySet());
+                    if (clientJwtConfiguration != null) {
+                        clientJwtConfiguration.writeValue(client);
+                    } else {
+                        logger.warn("Client with client jwt configuration not valid");
```
hsinn0: Should this not be treated as invalid case?

server/src/main/java/org/cloudfoundry/identity/uaa/client/ClientJwtConfiguration.java
```diff
@@ -0,0 +1,308 @@
...
+          cleanJwtString = JWK.parse(jsonMap).toPublicJWK().toString();
+        }
+        clientJwtConfiguration = new ClientJwtConfiguration(null, JsonWebKeyHelper.deserialize(cleanJwtString));
+        clientJwtConfiguration.validateJwkSet();
+      } catch (ParseException e) {
+        throw new InvalidClientDetailsException("Client jwt configuration cannot be parsed", e);
+      }
+    }
+    return clientJwtConfiguration;
+  }
```
hsinn0, line 161: It would make more sense to split this method to two separate ones, one for parsing `privateKeyUrl` and the other for parsing `privateKeyJwt`. I don't see any code shared between the two.

hsinn0 (changes requested):

model/src/main/java/org/cloudfoundry/identity/uaa/oauth/client/ClientJwtChangeRequest.java
strehle: done

server/src/main/java/org/cloudfoundry/identity/uaa/client/ClientAdminBootstrap.java
strehle: done

model/src/main/java/org/cloudfoundry/identity/uaa/oauth/client/ClientJwtChangeRequest.java
strehle: there are 3 input possiblities, therefore one method for convenience

server/src/main/java/org/cloudfoundry/identity/uaa/client/ClientAdminEndpoints.java
strehle: done

server/src/main/java/org/cloudfoundry/identity/uaa/client/ClientAdminEndpointsValidator.java
strehle: done

server/src/main/java/org/cloudfoundry/identity/uaa/client/ClientJwtConfiguration.java
strehle: allowed is one, either URI or JWK itself, therefore I wanted to have it in one class.

server/src/main/java/org/cloudfoundry/identity/uaa/client/ClientAdminBootstrap.java
hsinn0: What about when `!(map.get("jwks_uri") instanceof String || map.get("jwks") instanceof String)`?

server/src/main/java/org/cloudfoundry/identity/uaa/client/ClientAdminEndpointsValidator.java
hsinn0: Now we have changed the code to throw InvalidClientDetailsException if clientJwtConfiguration == null, we don't really need logger.warn() there, do we?

server/src/main/java/org/cloudfoundry/identity/uaa/client/ClientJwtConfiguration.java
hsinn0: The question was not about the class code but about having one method doing two different things. Not a big enough problem to block this PR, though. So I am fine with not changing the method.

server/src/main/java/org/cloudfoundry/identity/uaa/client/ClientAdminBootstrap.java
strehle: if now jwks/uri is set then now private_key_jwt is possible.
Similar to if you have no secret set, then client_secret cannot work

server/src/main/java/org/cloudfoundry/identity/uaa/client/ClientAdminEndpointsValidator.java
strehle: done

server/src/main/java/org/cloudfoundry/identity/uaa/client/ClientJwtConfiguration.java
strehle: i did some refactoring, if you have improvements you can do pushes

hsinn0 (approved):

PR #2453, opened 2023-08-24
Fix: UAA login page breaks when the product logo image is over 100000 characters

Tallicia:


Tallicia:
We need to fix the existing tests so this passes and then add an additional test.

server/src/main/java/org/cloudfoundry/identity/uaa/zone/MergedZoneBrandingInformation.java
Tallicia, line 1: This should be removed, not commented out.

Tallicia:
@strehle is out, @torsten-sap can you give your intput on this?

Side note, I'm not sure why the entire PR removed all the commits to refactor the function name.

server/src/main/java/org/cloudfoundry/identity/uaa/zone/MergedZoneBrandingInformation.java
```diff
@@ -33,6 +33,14 @@ public String getProductLogo() {
         return tryGet(IdentityZoneHolder.get(), BrandingInformationSource::getProductLogo).orElse(null);
     }
 
+    public String getInlinedBase64PngLogo() {
```
bruce-ricard, line 36: I happen to have context on this, so I think that I understand what you mean by "inlined", but I think that it wouldn't make much sense to someone out of context.

Naming is hard.

I was trying to start the 7 stages of naming with something like `getFullyQualifiedWithDataTypeForHtmlUseBase64PngLogo()`.

server/src/main/java/org/cloudfoundry/identity/uaa/zone/MergedZoneBrandingInformation.java
```diff
@@ -33,6 +33,14 @@ public String getProductLogo() {
...
+        String logo = getProductLogo();
```
bruce-ricard, line 37: Especially since this one appears to already be in base64, despite its name not reflecting that.

uaa/src/test/java/org/cloudfoundry/identity/uaa/login/LoginMockMvcTests.java
```diff
@@ -694,6 +694,15 @@ void testCustomFavIcon() throws Exception {
                 .andExpect(content().string(allOf(containsString("<link href='data:image/png;base64,/sM4lL==' rel='shortcut icon' />"), not(containsString("square-logo.png")))));
     }
 
+    @Test
+    void testProductLogoOver100kChars() throws Exception {
+        String bigLogo = new String(new char[150000]).replace('\0', 'x');
+        setZoneFavIconAndProductLogo(webApplicationContext, identityZoneConfiguration, null, bigLogo);
+
+        mockMvc.perform(get("/login"))
+                .andExpect(content().string(allOf(containsString("<style>.header-image {background-image: url(data:image/png;base64," + bigLogo + ");}</style>"), not(containsString("product-logo.png")))));
```
bruce-ricard, line 703: I guess this expect is not strictly required. All we need to validate is that the page loads correctly, don't we?
There already is a test that checks that the logo is present in the html just above.

Did you consider that?

bruce-ricard (commented):
Looks really good. Just made a couple comments related to the new function's name and the test. Please let me know what you think.

uaa/src/test/java/org/cloudfoundry/identity/uaa/login/LoginMockMvcTests.java
Tallicia: We also have to make sure this includes the biglogo, not the default fallback logo which happens if a concat returns a null.

server/src/main/java/org/cloudfoundry/identity/uaa/zone/MergedZoneBrandingInformation.java
hsinn0: We think that the existing name is sufficiently clear.

PR #2455, opened 2023-08-25
fix: do not default a missing secret to an empty one

### LINKED ISSUE #2454: Missing secret attribute in UAA YAML must not be used for any default passwords

Dev UAA , 76.19. , latest one

I am deploying the UAA

- using cf-release

### What did you do?

Defined a client in a yaml file to be provisioned with UAA. Usage was authorization code without a secret. So the client should not have a secret at all.
However the "" (empty) secret is created and can be used, this was unexpected. 

Omitting a secret should not lead to an empty secret. If I set secret: "" then OK, it was specified in yaml. But the secret was not defined.

### What did you see instead?

Code exchange was working with "" secret and the resulted token did no had the claim client_auth_method so from UAA it was a normal client authentication.

strehle:
If client has no secret set, then configuration should have no secret

Before: a missing secret was set as empty and this empty secret can be used even with client credential. After: a missing secret is null in DB layer so that you cannot do secret based login. There are other authentications so that the client is still usable.


server/src/main/java/org/cloudfoundry/identity/uaa/client/UaaClient.java
```diff
@@ -1,24 +1,53 @@
...
+    } else {
+      return false;
+    }
+  }
+
+  private Map<String, Object> getAdditionalInformation() {
     return this.additionalInformation;
   }
 
+  @Override
```
adrianhoelzl-sap, line 54: Can you please add a comment that this method - in contrast to the method in the super-class - never returns an empty string?

server/src/main/java/org/cloudfoundry/identity/uaa/client/UaaClient.java
```diff
@@ -1,24 +1,53 @@
...
+import java.util.Collections;
 import java.util.Map;
+import java.util.Optional;
 
 public class UaaClient extends User {
 
   private static final long serialVersionUID = SpringSecurityCoreVersion.SERIAL_VERSION_UID;
   private transient Map<String, Object> additionalInformation;
 
+  private String secret;
```
adrianhoelzl-sap: Consider making this `final`

uaa/src/test/java/org/cloudfoundry/identity/uaa/security/UaaUsingDelegatingPasswordEncoderMockMvcTest.java
```diff
@@ -54,10 +54,9 @@ void getClientCredentialsTokenWithValidPassword(String clientId) throws Exceptio
 
     @ParameterizedTest
     @ValueSource(strings = {
-            "client_id_with_no_password",
             "client_id_with_empty_password"
     })
-    void tryToGetTokenWithNoPasswordSucceeds(String clientId) throws Exception {
+    void tryToGetTokenWithEmtpyPasswordSucceeds(String clientId) throws Exception {
```
adrianhoelzl-sap, line 59:
```suggestion
    void tryToGetTokenWithEmptyPasswordSucceeds(String clientId) throws Exception {
```

server/src/main/java/org/cloudfoundry/identity/uaa/client/UaaClient.java
strehle: done

server/src/main/java/org/cloudfoundry/identity/uaa/client/UaaClient.java
strehle: done

uaa/src/test/java/org/cloudfoundry/identity/uaa/security/UaaUsingDelegatingPasswordEncoderMockMvcTest.java
strehle: changed name

strehle:
@Tallicia can you please ask a colleague to check this PR

hsinn0:
> @Tallicia can you please ask a colleague to check this PR

I am looking into this.

server/src/main/java/org/cloudfoundry/identity/uaa/authentication/ClientDetailsAuthenticationProvider.java
```diff
@@ -56,18 +55,18 @@ protected void additionalAuthenticationChecks(UserDetails userDetails, UsernameP
...
+                if (authentication.getCredentials() == null) {
+                    if (isPublicGrantTypeUsageAllowed(authentication.getDetails()) && uaaClient.isAllowPublic()) {
+                        // in case of grant_type=authorization_code and code_verifier passed (PKCE) we check if client has option allowpublic with true and continue even if no secret is in request
                         ((UaaAuthenticationDetails) authentication.getDetails()).setAuthenticationMethod(CLIENT_AUTH_NONE);
                         break;
                     }
                 }
-                super.additionalAuthenticationChecks(user, authentication);
+                if (uaaClient.getPassword() == null) {
+                    throw new BadCredentialsException("Missing credentials");
```
hsinn0: It would be more efficient doing `error = new BadCredentialsException("Missing credentials"); break;` here instead of throwing the new `BadCredentialsException` and catching and assigning it to `error` below.

strehle: done

hsinn0 (approved):

PR #2484, opened 2023-09-12
test: Authorization Grant Flow without Redirect URI

strehle:
Add test for this hidden feature thus not to run into a regression

### DIFF
--- a/uaa/src/main/webapp/WEB-INF/spring/oauth-clients.xml
+++ b/uaa/src/main/webapp/WEB-INF/spring/oauth-clients.xml
@@ -82,7 +82,7 @@
                             <entry key="secret" value="loginsecret"/>
                             <entry key="scope" value="openid,oauth.approvals"/>
                             <entry key="authorized-grant-types" value="client_credentials,authorization_code"/>
-                            <entry key="redirect-uri" value="http://localhost/**,http://localhost:7000/**"/>
+                            <entry key="redirect-uri" value="http://localhost/**"/>
                             <entry key="authorities"
                                    value="oauth.login,scim.write,clients.read,notifications.write,critical_notifications.write,emails.write,scim.userids,password.write,idps.write"/>
                             <entry key="autoapprove" value="true"/>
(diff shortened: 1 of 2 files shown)


uaa/src/main/webapp/WEB-INF/spring/oauth-clients.xml
```diff
@@ -82,7 +82,7 @@
                             <entry key="secret" value="loginsecret"/>
                             <entry key="scope" value="openid,oauth.approvals"/>
                             <entry key="authorized-grant-types" value="client_credentials,authorization_code"/>
-                            <entry key="redirect-uri" value="http://localhost/**,http://localhost:7000/**"/>
```
peterhaochen47, line 85: What is the purpose of this change? Is this required for the added test?

strehle: it mainly is the revert of my change https://github.com/cloudfoundry/uaa/pull/2403/files#diff-d02078cfeded6b34f4eb0a68b222758f944db9bbe2fa1227249a9c782718cf09R85

the login client is used in our internal forked scenario and it was broken. Because UAA has this hidden feature of redirect uris. It means if you have only one redirect uri defined you can omit the parameter redirect_uri in authorize request.
If you have an array in the client you cannot use this feature anymore therefore removed the 2nd entry

strehle: and create an explicit test for this hidden feature

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/AuthorizationCodeGrantIntegrationTests.java
```diff
@@ -192,6 +194,34 @@ public void testZoneInactive() {
         assertEquals(HttpStatus.NOT_FOUND, result.getStatusCode());
     }
 
+    @Test
+    public void testAuthorizationRequestWithoutRedirectUri() {
+
+      Map<String, String> body = IntegrationTestUtils.getAuthorizationCodeTokenMap(serverRunning,
+          testAccounts,
+          "login",
+          "loginsecret",
```
strehle, line 203: this client has only one redirect uri therefore you can omit it.

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/AuthorizationCodeGrantIntegrationTests.java
```diff
@@ -192,6 +194,34 @@ public void testZoneInactive() {
...
+          null,
+          null,
+          null,
+          false);
+
+      assertNotNull("Token not received", body.get("access_token"));
+
+      try {
+        IntegrationTestUtils.getAuthorizationCodeTokenMap(serverRunning, testAccounts, "app", "appclientsecret",
+            testAccounts.getUserName(), testAccounts.getPassword(),
```
strehle, line 216: the app client has 2 entries for redirect-uri thus here it does not work.

These tests simply should help to know what we have already, nothing new

peterhaochen47 (approved):

PR #2486, opened 2023-09-13
fix: error go: download go1.21 for linux/amd64: toolchain not available

strehle:
https://github.com/golang/go/issues/62278

### DIFF
--- a/k8s/go.mod
+++ b/k8s/go.mod
@@ -1,6 +1,8 @@
 module github.com/cloudfoundry/uaa
 
-go 1.21.0
+go 1.21
+
+toolchain go1.21.0
 
 require (
 	github.com/onsi/ginkgo v1.16.5
--- a/k8s/go.mod
+++ b/k8s/go.mod
@@ -1,6 +1,6 @@
 module github.com/cloudfoundry/uaa
 
-go 1.21
+go 1.21.0
 
 require (
 	github.com/onsi/ginkgo v1.16.5


strehle:
see error https://github.com/cloudfoundry/uaa/network/updates

swalchemist:
@strehle We have a CredHub repo that is configured similarly without the toolchain directive and we haven't seen any issues yet. Did you see the problem when you ran "go mod tidy" again?

strehle:
Hi @swalchemist , I have seen the error after an update we see internally but not in https://github.com/cloudfoundry/uaa
Then I checked dependabot and found error/warning, see https://github.com/cloudfoundry/uaa/network/updates/721814377

Open https://github.com/cloudfoundry/uaa/network/updates and then you should see what I mean, e.g.

[k8s/go.mod](https://github.com/cloudfoundry/uaa/blob/develop/k8s/go.mod) 
 [Last checked 8 minutes ago](https://github.com/cloudfoundry/uaa/network/updates/721814377)

Internally we have in addition renovate configured and this has created an update for k8s.io
v0.28.1 -> v0.28.2. Dependabot normally does the same updates as our internal repo, therefore my request to have a look on it. 
The go updates are not that important since uaa has deprecated k8s support.

I see that GHAS (github action) runs with go 1.20.xxx workers and then you are behind what we define btw. your https://github.com/cloudfoundry/credhub-cli has no dependabot at all.

PR #2490, opened 2023-09-16
Page object refactoring for two additional tests

swalchemist:
This brings two more tests into the page object pattern. As is typical when adding more tests using the page objects, some functionality is added to the page objects, but it wasn't much.

For testSimpleSamlPhpLoginDisplaysLastLogin, this replaces what was already a basic procedural method design for login() and logout() with an object-oriented one. Also, an interaction with webdriver in an assertion is extracted to the appropriate object for possible reuse.

For testSingleLogoutWithNoLogoutUrlOnIDP, it violated the DRY principle fairly thoroughly and the page objects fix that for the webdriver-related interactions. Also, the introduction of the clickSamlLink_goToHomePage() method makes the purpose of the test easier to understand.

Several magic strings for test credentials are replaced with constants, though further refactoring is probably warranted to use a similar mechanism as with test accounts used in other tests in this class and elsewhere.

### DIFF
--- a/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/SamlLoginIT.java
+++ b/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/SamlLoginIT.java
@@ -38,7 +38,6 @@ import org.cloudfoundry.identity.uaa.util.JsonUtils;
 import org.cloudfoundry.identity.uaa.util.RetryRule;
 import org.cloudfoundry.identity.uaa.zone.IdentityZone;
 import org.cloudfoundry.identity.uaa.zone.IdentityZoneConfiguration;
-import org.cloudfoundry.identity.uaa.zone.IdentityZoneHolder;
 import org.flywaydb.core.internal.util.StringUtils;
 import org.hamcrest.Matchers;
 import org.junit.After;
@@ -445,21 +444,11 @@ public class SamlLoginIT {
 
         provider = IntegrationTestUtils.createOrUpdateProvider(zoneAdminToken, baseUrl, provider);
 
-        webDriver.get(baseUrl + "/login");
-        Assert.assertEquals("Cloud Foundry", webDriver.getTitle());
-        webDriver.findElement(By.xpath("//a[text()='" + provider.getConfig().getLinkText() + "']")).click();
-        webDriver.findElement(By.xpath(SIMPLESAMLPHP_LOGIN_PROMPT_XPATH_EXPR));
-        webDriver.findElement(By.name("username")).clear();
-        webDriver.findElement(By.name("username")).sendKeys(testAccounts.getUserName());
-        webDriver.findElement(By.name("password")).sendKeys(testAccounts.getPassword());
-        webDriver.findElement(By.xpath("//input[@value='Login']")).click();
-        assertThat(webDriver.findElement(By.cssSelector("h1")).getText(), Matchers.containsString("Where to"));
-
-        webDriver.findElement(By.cssSelector(".dropdown-trigger")).click();
-        webDriver.findElement(By.linkText("Sign Out")).click();
-        webDriver.findElement(By.xpath("//a[text()='" + provider.getConfig().getLinkText() + "']")).click();
-
-        assertThat(webDriver.findElement(By.cssSelector("h1")).getText(), Matchers.containsString("Where to"));
+        LoginPage.go(webDriver, baseUrl)
+                .startSamlLogin()
+                .login(testAccounts.getUserName(), testAccounts.getPassword())
+                .logout()
+                .automaticSamlLogin();
     }
 
     @Test
--- a/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/SamlLoginIT.java
+++ b/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/SamlLoginIT.java
@@ -117,6 +117,13 @@ import static org.springframework.security.oauth2.common.util.OAuth2Utils.GRANT_
 @ContextConfiguration(classes = DefaultIntegrationTestConfig.class)
 public class SamlLoginIT {
 
+    public static final String MARISSA4_USERNAME = "marissa4";
+    private static final String MARISSA4_PASSWORD = "saml2";
+    public static final String MARISSA4_EMAIL = "marissa4@test.org";
+    public static final String MARISSA2_USERNAME = "marissa2";
+    private static final String MARISSA2_PASSWORD = "saml2";
+    public static final String MARISSA3_USERNAME = "marissa3";
+    private static final String MARISSA3_PASSWORD = "saml2";
     private static final String SAML_ORIGIN = "simplesamlphp";
     @Autowired @Rule
     public IntegrationTestRule integrationTestRule;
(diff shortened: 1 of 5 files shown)


uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/SamlLoginIT.java
```diff
@@ -445,21 +444,11 @@ public void testSingleLogoutWithNoLogoutUrlOnIDP() throws Exception {
...
-        webDriver.findElement(By.cssSelector(".dropdown-trigger")).click();
-        webDriver.findElement(By.linkText("Sign Out")).click();
-        webDriver.findElement(By.xpath("//a[text()='" + provider.getConfig().getLinkText() + "']")).click();
-
-        assertThat(webDriver.findElement(By.cssSelector("h1")).getText(), Matchers.containsString("Where to"));
+        LoginPage.go(webDriver, baseUrl)
+                .startSamlLogin()
+                .login(testAccounts.getUserName(), testAccounts.getPassword())
+                .logout()
+                .automaticSamlLogin();
```
peterhaochen47: Does `.automaticSamlLogin();` whose definition is just
```
public HomePage automaticSamlLogin() {
        clickFirstSamlLoginLink();
        return new HomePage(driver);
    }
```
cover the assertion `assertThat(webDriver.findElement(By.cssSelector("h1")).getText(), Matchers.containsString("Where to"));`?

peterhaochen47: I see, it does [here](https://github.com/cloudfoundry/uaa/blob/9108bcca50b86e3591e0b8590b58c0e20be66c74/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/pageObjects/HomePage.java#L17-L18).

peterhaochen47: My feeling is that `automaticSamlLogin();` abstracts out too much details & a new reader wouldn't understand the basic test logic without clicking into `automaticSamlLogin()`. And after that, there's a slight learning curve in understanding that `return new XXXPage(driver);` contains an assertion that you on XXXPage (not sure how to improve this though).

I think `startSamlLogin` might have a similar problem of being too abstract too, but to a lesser degree.
It's also slightly hard to figure out what `.login` and `.logout` do (are you performing login and logout on UAA page or the SAML page?) without checking the return type (aka which Page object) of the previous function call.

Maybe refactoring `login` to `loginOnSamlLoginPage`? And refactoring `logout` to `logoutOnUAAHomePage`? Verbosity doesn't harm here right?

I'm thinking about what would be readable, like an English sentence. Without considering implementation, maybe something like this:

        LoginPage.go(webDriver, baseUrl)
                .clickFirstSamlLoginLink() 
                .assertOnSamlLoginPage()  
                .loginOnSamlLoginPage(testAccounts.getUserName(), testAccounts.getPassword())
                .assertOnUAAHomePage()
                .logoutOnUAAHomePage()
                .clickFirstSamlLoginLink() 
                .assertOnUAAHomePage()  
or
```
        LoginPage.go(webDriver, baseUrl)
                .clickFirstSamlLoginLinkAndLandOnSamlLoginPage()
                .loginOnSamlLoginPageAndGetsRedirectedBackToUAAHomePage(testAccounts.getUserName(), testAccounts.getPassword())
                .logoutOnUAAHomePage()     
                .clickFirstSamlLoginLinkWithoutHavingToReenterCredentialsAndGetsRedirectedBackToUAAHomePage().
```

swalchemist: > Does `.automaticSamlLogin();` ...
> cover the assertion `assertThat(webDriver.findElement(By.cssSelector("h1")).getText(), Matchers.containsString("Where to"));`?

Yes, as you found, every time we go to a page, when we construct the object for it, it starts by validating that we're on that page. The page objects, however, could be changed to do that in a different way. So if a test really wanted to make sure that text was on the page, a test (ideally only 1 test), should have code to explicitly test that. We would add a method to the page Page object to facilitate this. In fact, I think I'll do some refactoring to show how.

Note that the new check looks at all the text on the page and doesn't care if it's in an h1 block. That's intended to make the test easier to maintain by removing unnecessary details.

peterhaochen47: Great! FYI, I further edited my pseudocode above to demonstrate my points.

swalchemist: > My feeling is that `automaticSamlLogin();` abstracts out too much details & a new reader wouldn't understand the basic test logic without clicking into `automaticSamlLogin()`.

clickFirstSamlLoginLink() is a private method, and it doesn't support the fluent interface. The design I'm coalescing on is having the expected landing page indicated by the choice of navigation method. Let's see if we can improve the automaticSamlLogin() name. How about goToSamlLoginPageWhenAlreadyLoggedIn()?

Would you want the landing page included in this and every other method that navigates to a different page? Something like goToSamlLoginPageWhichRedirectsToHomePage()? I don't think I like that as much.

peterhaochen47: My guess is that a new reader wouldn't understand 
```
        LoginPage.go(webDriver, baseUrl)
                .startSamlLogin()
                .login(testAccounts.getUserName(), testAccounts.getPassword())
                .logout()
                .goToSamlLoginPageWhenAlreadyLoggedIn();
```
either.

peterhaochen47: My guess is that a new reader wouldn't understand 
```
        LoginPage.go(webDriver, baseUrl)
                .startSamlLogin()
                .login(testAccounts.getUserName(), testAccounts.getPassword())
                .logout()
                .goToSamlLoginPageWhenAlreadyLoggedIn();
```
either without clicking into all these functions. Maybe we can ask a fresh pair of eyes to read this. @bruce-ricard?

swalchemist: I think it's ok if someone needs to look at the implementation of a method to understand why the navigation is going the way it is, if the page objects are well-named and possibly commented. These navigation methods can serve as documentation for how the site is expected to work. But let's see if we can minimize how often that would need to happen.

This example you wrote is interesting. Some of the navigation methods clarify where they're coming from (like logoutOnUAAHomePage()), which helps because there are many pages you can log out from, and at least two different logout pages. Following that design, however, would mean that instead of relying on inheriting the logout method from the base class like we do now, we would need to have a method in each class we need to log out from in order to give it that unique name.

Changing to using methods like assertOnSamlLoginPage() would result in more verbose tests, but it would perhaps make the flow easier to follow. I think what we'd need to do is that these validator also be the methods that return the new page objects for the new current page.

swalchemist: I just pushed a refactor. I think it's clear now which page we're coming from, what action is taken, and where we're going. @peterhaochen47 what do you think?

```
        LoginPage.go(webDriver, baseUrl)
                .clickSamlLink_goToSamlLoginPage()
                .login_goToHomePage(testAccounts.getUserName(), testAccounts.getPassword())
                .logout_goToLoginPage()
                .clickSamlLink_goToSamlLoginPage();
```

peterhaochen47: Looks good!

peterhaochen47 (approved):

PR #2498, opened 2023-09-22
refactor: use page objects for favicon test

swalchemist:
* The purpose of this test was difficult to understand until I temporarily removed the code that fixed the bug that this test is testing for.
* It should now be easier to understand why this test exists.


strehle (approved):
ok for me

PR #2501, opened 2023-09-26
Passcode test refactor

swalchemist:
This applies the page object strategy to the testSimpleSamlPhpPasscodeRedirect test. This required using a factory method to start the interaction on the passcode page that immediately redirects to the login page (requestPasscode_goToLoginPage) because the pattern I'm using here is that a constructor will expect it to land on the passcode page.

Also included a bit of code improvement in the page objects.


strehle (approved):
ok for me

Issue #1448, opened 2020-11-26
Feature request: Rotation of refresh tokens for public clients


sebastianGit:
The current OAuth drafts with security best practices for Single Page Applications / Browser-based Apps ask for refresh token rotation, in case refresh tokens are issued to public clients.
See:

- https://tools.ietf.org/html/draft-ietf-oauth-browser-based-apps-07#section-4 
- https://tools.ietf.org/html/draft-ietf-oauth-v2-1-00#section-6.1

As of now (v74.28.0), refresh token rotation is not supported by CF UAA.
Are there already plans to support this in CF UAA (e.g., as a feature that can be enabled/configured per client)?

strehle:
UAA currently does not have real public clients. We have confidential clients with known secrets, e.g. famous cf client

So I this ticket as 2 feature requests
a) support public clients ( maybe in combination with PKCE ;) )
b) support rotation in refresh flow for public clients

strehle:
@sebastianGit UAA will get public feature with next release so we will start with this request and therefore a question.

a) do you expect that if a client has the new allowpublic feature that its refresh tokens replaced always ?
b) do you expect that if a client has the new allowpublic feature that the refresh of its token also is possible without client_secret

PR #2504, opened 2023-09-26
feature: add client_auth_method=none into tokens for clients with empty secret

strehle:
Internally use empty, but write none into tokens. Externally use always none, but internally empty to distinguish in the flow.
The refresh flow with an empty secret is something we allow for now.

So a combination from https://github.com/cloudfoundry/uaa/pull/2435 (external view) and
https://github.com/cloudfoundry/uaa/pull/2447 (internal view)


server/src/test/java/org/cloudfoundry/identity/uaa/oauth/RefreshRotationTest.java
```diff
@@ -164,6 +198,35 @@ void testRefreshPublicClientWithoutRotation() {
...
+    OAuth2Request oAuth2Request = authorizationRequest.createOAuth2Request();
+    OAuth2Authentication authentication = new OAuth2Authentication(oAuth2Request, tokenSupport.defaultUserAuthentication);
+    new IdentityZoneManagerImpl().getCurrentIdentityZone().getConfig().getTokenPolicy().setRefreshTokenRotate(true);
+    CompositeToken accessToken = (CompositeToken) tokenServices.createAccessToken(authentication);
+
+    assertThat(UaaTokenUtils.getClaims(accessToken.getValue()), hasEntry(CLIENT_AUTH_METHOD, CLIENT_AUTH_NONE));
+    String refreshTokenValue = accessToken.getRefreshToken().getValue();
+    assertThat(refreshTokenValue, is(notNullValue()));
+
+    new IdentityZoneManagerImpl().getCurrentIdentityZone().getConfig().getTokenPolicy().setRefreshTokenRotate(false);
```
peterhaochen47, line 222: How does `tokenPolicy.refreshTokenRotate` interact with the logic added in this PR? Based on this test, it seems like when `tokenPolicy.refreshTokenRotate=false`, you forbid the public client to refresh tokens, and when `tokenPolicy.refreshTokenRotate=true`, you allow?

Why?

When `tokenPolicy.refreshTokenRotate=true`, the refresh flow will still return a valid (though different) refresh token. Why more permissive under this case?

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/RefreshRotationTest.java
```diff
@@ -164,6 +198,35 @@ void testRefreshPublicClientWithoutRotation() {
...
+    tokenSupport.clientDetailsService.setClientDetailsStore(IdentityZoneHolder.get().getId(), Collections.singletonMap(CLIENT_ID, clientDetails));
+    AuthorizationRequest authorizationRequest = new AuthorizationRequest(CLIENT_ID, tokenSupport.requestedAuthScopes);
+    authorizationRequest.setResourceIds(new HashSet<>(tokenSupport.resourceIds));
+    Map<String, String> azParameters = new HashMap<>(authorizationRequest.getRequestParameters());
+    azParameters.put(GRANT_TYPE, GRANT_TYPE_AUTHORIZATION_CODE);
+    authorizationRequest.setRequestParameters(azParameters);
+    authorizationRequest.setExtensions(Map.of(CLIENT_AUTH_METHOD, CLIENT_AUTH_EMPTY));
+    OAuth2Request oAuth2Request = authorizationRequest.createOAuth2Request();
+    OAuth2Authentication authentication = new OAuth2Authentication(oAuth2Request, tokenSupport.defaultUserAuthentication);
+    new IdentityZoneManagerImpl().getCurrentIdentityZone().getConfig().getTokenPolicy().setRefreshTokenRotate(true);
```
peterhaochen47, line 215: The default value (aka if without this line) of `tokenPolicy.refreshTokenRotate` is `false` based on my debugging, so this line does change something. However, commenting out this line, the test still pass.

peterhaochen47:
@strehle Based on the title & description of this PR, it seems that this PR's sole purpose is to inject an extra claim `client_auth_method=none` into the token if the client that originally requested this token is a public client (either no secret or empty string secret). 

But when I look at all the tests added (I use the test changes to understand the exact implications of the code changes), it looks like the PR does more than that (for example, [a test](https://github.com/cloudfoundry/uaa/blob/0a7433f8644d13dd6a45ce0b58694753136aaf43/server/src/test/java/org/cloudfoundry/identity/uaa/oauth/RefreshRotationTest.java#L227) is added to forbid refresh request under some conditions)? Is that true? Or are these test "backfilled tests" testing existing logics of `develop` branch? 

To help me understand this PR better, could you summarize, point by point, all the changes included in this PR. This will help me validate my understandings &  help us find out what doc edits are required.

strehle:
> @strehle Based on the title & description of this PR, it seems that this PR's sole purpose is to inject an extra claim `client_auth_method=none` into the token if the client that originally requested this token is a public client (either no secret or empty string secret).
> 
> But when I look at all the tests added (I use the test changes to understand the exact implications of the code changes), it looks like the PR does more than that (for example, [a test](https://github.com/cloudfoundry/uaa/blob/0a7433f8644d13dd6a45ce0b58694753136aaf43/server/src/test/java/org/cloudfoundry/identity/uaa/oauth/RefreshRotationTest.java#L227) is added to forbid refresh request under some conditions)? Is that true? Or are these test "backfilled tests" testing existing logics of `develop` branch?
> 
> To help me understand this PR better, could you summarize, point by point, all the changes included in this PR. This will help me validate my understandings & help us find out what doc edits are required.

The inject of ` client_auth_method=none` into the token was introduced with https://github.com/cloudfoundry/uaa/issues/2138 . The feature of allowing public scenarios like authorization code flow and refresh flow without any secret was added as new feature into UAA but with constraint of doing this only if we have a rotation of the refresh token. The rotation of the refresh token is a very important feature because it limits the risk of a stolen refresh token. This feature was done because of https://github.com/cloudfoundry/uaa/issues/1448 . The origin of it is https://oauth.net/2.1/ where public usages are explicitly mentioned.

And this blocks it right now, means the mix of the public refresh exchange with this approach to limit the empty secret. The empty secret and its refresh is an existing scenario which we cannot break. 
For new features we can have limitations, like you get the public refresh only with a refresh rotation....

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/RefreshRotationTest.java
strehle: This refreshTokenRotate is a prerequisite for the public usage, see my explanations below about https://oauth.net/2.1/.
Yes, I forbid the public usage without rotation because of standard . 
The refresh rotation is an important feature but I can enforce this only for new things.

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/RefreshRotationTest.java
strehle: yes because it was a new feature from https://github.com/cloudfoundry/uaa/issues/1448 . 
Clients need to follow the exchange on their side as well. If a client stores a refresh token only but does not care of a new value when using calling grant_type=refresh_token therefore a new feature and as always new features are off by default.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java
```diff
@@ -489,7 +491,7 @@ private CompositeToken createCompositeToken(String tokenId,
 
     private static Map<String, Object> addRootClaimEntry(Map<String, Object> additionalRootClaims, String entry, String value) {
         Map<String, Object> claims = additionalRootClaims != null ? additionalRootClaims : new HashMap<>();
-        claims.put(entry, value);
+        claims.put(entry, CLIENT_AUTH_EMPTY.equals(value) ? CLIENT_AUTH_NONE : value);
```
strehle, line 495: @peterhaochen47 here is the trick.

With this I set `client_auth_method=none` to the token, but internally we now that it was an empty secret and the deny of refresh - see line 327 - is only with none

strehle:
@peterhaochen47 please see this PR as replacement for https://github.com/cloudfoundry/uaa/pull/2435

peterhaochen47:
I'm trying to understand the original problem (that this PR is intended to help address) better: https://github.com/cloudfoundry/uaa/pull/2193#issuecomment-1741594174

strehle:
> I'm trying to understand the original problem (that this PR is intended to help address) better: [#2193 (comment)](https://github.com/cloudfoundry/uaa/pull/2193#issuecomment-1741594174)

This PR is needed to solve failing test b) , see https://github.com/cloudfoundry/uaa/pull/2193#issuecomment-1745437039

because this change add client_auth_method=none to cf client autentication and then later we can use none to identiy that it was public and public-2-public should be ok

strehle:
@peterhaochen47 After rebase this PR is smaller and therefore maybe you can now review this much faster ?

peterhaochen47:
Based on [your comment](https://github.com/cloudfoundry/uaa/pull/2504#issuecomment-1745441386), I thought that this PR was mainly for resolving the failures in #2193 which I have [requested](https://github.com/cloudfoundry/uaa/pull/2193#issuecomment-1749847451) to be closed. Do you now suggest that this PR be reviewed independently without considering #2193?

strehle:
> Do you now suggest that this PR be reviewed independently without considering #2193?

Yes please do so. I had the plan to re-use this, but we may wont fix user token.

However this PR is simply about that none should appear in access_token for a not existing secret and for an empty one, so that we can treat in future both as public usage.

peterhaochen47 (approved):

strehle:
FYI: this change is not related to https://github.com/cloudfoundry/uaa/pull/2571 even if topic is empty secret.

PR #2506, opened 2023-09-27
Upgrade to be compatible with simplesamlphp v2

bruce-ricard:
Luckily, we only needed to update tests.

Context: 
1. The newer versions of `cf-deployment` only support `cflinuxfs4` stemcells.
2. The SimpleSamlPHP app used in UAA CI (version 1) is `cf push`ed to CF, and is only compatible with `cflinuxfs3` stemcells.
3. The version 2 of that SimpleSamlPHP app is compatible with `cflinuxfs4`, so we decided that we need to start using it instead of the old v1 one.
4. When we run the integration tests against the new SimpleSamlPHP v2 server, some of the tests started failing, due to breaking changes in that new major version 2 of SimpleSamlPHP.
5. Hence this PR, with fixes to the tests.


uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/util/IntegrationTestUtils.java
```diff
@@ -86,10 +86,10 @@
 
 public class IntegrationTestUtils {
 
-    public static final String SIMPLESAMLPHP_UAA_ACCEPTANCE = "http://simplesamlphp.uaa-acceptance.cf-app.com";
+    public static final String SIMPLESAMLPHP_UAA_ACCEPTANCE = "http://simplesamlphp2.uaa-acceptance.cf-app.com";
```
bruce-ricard, line 89: Not to break the tests when we merge and the tests run on CI, we temporarily changed the URL to point towards the already deployed v2 saml server. We'll revert this after we have deployed the new v2 server to the old URL.

strehle (approved):
cannot really judge but ok for me

PR #2511, opened 2023-09-30
Make it easier to find test failures in CI output

swalchemist:
Searching for "FAILURE" in the CI output in a browser (case insensitive) to find failed tests always finds things that aren't test failures. This PR applies the principle of never logging the word "failure" except when JUnit is reporting a test failure. 

"Fail" is still ok to include in method names and other output.

Two additional refactors in the affected code are also included.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -91,12 +91,13 @@ subprojects {
         jvmArgs += ["-Xmx1024m", "-XX:+StartAttachListener", "-XX:+HeapDumpOnOutOfMemoryError", "-XX:HeapDumpPath=/var/log/uaa-tests.hprof"]
 
         retry {
-            def retryFailOnPassedAfterRetry = Boolean.parseBoolean(
+            def failOnPassedAfterRetrySystemProperty = Boolean.parseBoolean(
                     System.getProperty("failOnPassedAfterRetry", "false"));
-            if (!retryFailOnPassedAfterRetry)
+            if (!failOnPassedAfterRetrySystemProperty)
                 logger.warn("retry: Flaky tests will not make the test run fail because failOnPassedAfterRetry is false.")
 
-            failOnPassedAfterRetry = retryFailOnPassedAfterRetry
+            // Configure the retry extension
+            failOnPassedAfterRetry = failOnPassedAfterRetrySystemProperty
             maxFailures = Integer.parseInt(System.getProperty("maxFailures", "3"))
             maxRetries = Integer.parseInt(System.getProperty("maxRetries", "1"))
         }
--- a/build.gradle
+++ b/build.gradle
@@ -94,7 +94,7 @@ subprojects {
             def retryFailOnPassedAfterRetry = Boolean.parseBoolean(
                     System.getProperty("failOnPassedAfterRetry", "false"));
             if (!retryFailOnPassedAfterRetry)
-                logger.warn("retry: Failed tests are to be retried and considered as passed if the retry passes.")
+                logger.warn("retry: Flaky tests will not make the test run fail because failOnPassedAfterRetry is false.")
 
             failOnPassedAfterRetry = retryFailOnPassedAfterRetry
             maxFailures = Integer.parseInt(System.getProperty("maxFailures", "3"))
--- a/metrics-data/src/test/java/org/cloudfoundry/identity/uaa/util/JsonUtilsTest.java
+++ b/metrics-data/src/test/java/org/cloudfoundry/identity/uaa/util/JsonUtilsTest.java
@@ -10,8 +10,8 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.junit.jupiter.api.Assertions.fail;
 
 class JsonUtilsTest {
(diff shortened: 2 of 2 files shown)


hsinn0:
I do not see much usefulness of this PR. We can just search for "FAILED" case-sensitive, can't we?

hsinn0 (commented):
I do not see much usefulness of this PR. We can just search for "FAILED" case-sensitive, can't we?

swalchemist:
> I do not see much usefulness of this PR. We can just search for "FAILED" case-sensitive, can't we?

Web browsers make it difficult to do a case-sensitive search.

Also, there are a few refactors here that I think are worthwhile.

uaa/src/test/java/org/cloudfoundry/identity/uaa/authentication/manager/DynamicZoneAwareAuthenticationManagerTest.java
```diff
@@ -154,7 +154,7 @@ void testNonUAAZoneUaaActiveUaaAuthenticationSuccess() {
     }
 
     @Test
-    void testNonUAAZoneUaaActiveUaaAuthenticationFailure() {
+    void testNonUAAZoneUaaActiveUaaAuthenticationFail() {
```
bruce-ricard: > refactor: rename methods with "Failure" in the name
> * Reserve "FAILURE" for JUnit test fail logs so we won't find these methods in the logs when looking for test failures.

I don't understand this. AFAIK the Junit keyword we see in the test output is `FAILED` not `FAILURE`.

build.gradle
```diff
@@ -91,12 +91,12 @@ subprojects {
         jvmArgs += ["-Xmx1024m", "-XX:+StartAttachListener", "-XX:+HeapDumpOnOutOfMemoryError", "-XX:HeapDumpPath=/var/log/uaa-tests.hprof"]
 
         retry {
-            def retryFailOnPassedAfterRetry = Boolean.parseBoolean(
+            def failOnPassedAfterRetryProperty = Boolean.parseBoolean(
```
bruce-ricard: I really appreciate the effort of trying to make this variable easier to understand.

I don't understand the name of this new variable though.

If you think this one is better than the old one, it's better this way. I think that we should be able to do even better though.

build.gradle
```diff
@@ -94,7 +94,7 @@ subprojects {
             def retryFailOnPassedAfterRetry = Boolean.parseBoolean(
                     System.getProperty("failOnPassedAfterRetry", "false"));
```
bruce-ricard, line 95: OK I guess the problem is coming from here. `failOnPassedAfterRetry` means nothing to me. I guess that's not your fault.

metrics-data/src/test/java/org/cloudfoundry/identity/uaa/util/JsonUtilsTest.java
```diff
@@ -69,13 +69,10 @@ void testSerializeExcludingProperties() {
...
   void testSerializeExcludingPropertiesInnerCallFails() {
-    Map<String, String> groupProperties = JsonUtils.readValue(jsonTestObjectString, new TypeReference<Map<String, String>>() {});
-    try {
-      JsonUtils.serializeExcludingProperties(groupProperties, "limit.unkonwn");
-      fail("not expected");
-    } catch (Exception e) {
-      assertTrue(e instanceof JsonUtils.JsonUtilException);
-    }
+    Map<String, String> groupProperties = JsonUtils.readValue(jsonTestObjectString, new TypeReference<>() {});
+    assertThrows(JsonUtils.JsonUtilException.class, () -> {
```
bruce-ricard, line 73: Nice! Thanks.

metrics-data/src/test/java/org/cloudfoundry/identity/uaa/util/JsonUtilsTest.java
```diff
@@ -68,7 +68,7 @@ void testSerializeExcludingProperties() {
   }
 
   @Test
-  void testSerializeExcludingPropertiesInnerCallFailed() {
+  void testSerializeExcludingPropertiesInnerCallFails() {
```
bruce-ricard, line 71: This commit makes a LOT of sense to me. Great!

bruce-ricard (changes requested):
I like the idea for the change. Thank you for doing this.

I myself [committed a similar change](https://github.com/cloudfoundry/uaa/commit/9615a2da38db9f1f9a12dc3e4fb6b4719b9e2cb9) several years ago.

Mostly requesting changing because of the `FAILURE` commit that I don't understand.

uaa/src/test/java/org/cloudfoundry/identity/uaa/authentication/manager/DynamicZoneAwareAuthenticationManagerTest.java
swalchemist: Dang it. I'll revert this one.

build.gradle
swalchemist: I think the old one was very misleading - implying that it would affect whether a retry would happen, which is not what it does. It only affects whether the test will be reported as a pass or fail based on whether it failed during any of the retries.

I'm sure we can improve on the new name. I did a bit of experimenting because it's not clear why this variable is needed as an intermediary at all. Assigning directly to the failOnPassedAfterRetry property and using it in a boolean did not work, so I kept the variable and just tried to give it a better name.

build.gradle
swalchemist: Right, this is defined by a plugin. If we set to this true, and a test fails, then it seems that the plugin will retry a certain number of times to try to get a pass, but then it will still report it ultimately as a fail. I suppose we'd be able to look at the output to see if it was flaky behavior or a consistent fail.

build.gradle
swalchemist: Renamed to failOnPassedAfterRetrySystemProperty and added a comment.

PR #2513, opened 2023-10-01
build: change sonar runner to java 17

strehle:
sonar 
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2513

https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent


peterhaochen47:
Hi @strehle, is this just a preliminary test to see if UAA can be upgraded to using Java 17?

strehle:
uaa can be upgraded, because we have a github action already, see
https://github.com/cloudfoundry/uaa/blob/develop/.github/workflows/gradle.yml#L19
or run
https://github.com/cloudfoundry/uaa/actions/workflows/gradle.yml

So this is simply to switch sonar to 17 to follow the warnings which are there in current runs, means sonar wants to have now 17. 

Most github actions here depend on docker images, so the switch is done if you change the image

bruce-ricard (approved):
Makes sense to me. 👍

PR #2520, opened 2023-10-04
documentation: update system admin guide

### LINKED ISSUE #859: [docs] update sysadmin-guide for new JWT signing keys system
https://github.com/cloudfoundry/uaa/blob/master/docs/Sysadmin-Guide.rst#id8 documents the deprecated configuration for JWT signing keys; could this section be upgraded to the new syntax and also include a discussion on when/how to rotate keys?

strehle:


strehle:
@Tallicia this is not really new, but only a description for the now, means how it works now. If vmware wants to review it please go ahead

docs/Sysadmin-Guide.rst
```diff
@@ -103,14 +103,15 @@ written as an environment variable. For a Cloud Foundry application this could l
     ---
       applications:
       - name: standalone-uaa-cf-war
-        memory: 512M
+        memory: 4096M
```
hsinn0, line 106: Any specific reason increasing this example memory size?

strehle: our experience, ok we can have 2048 ? because I am sure that we 512 the current UAA with all the spring stuff inside will often crash with out of memory ....

What would you propose ?

hsinn0: > What would you propose ?

Nothing really. I was just curious about how you picked the number. I do agree that 512M was too low. The new value is fine with me.

hsinn0 (approved):

PR #2543, opened 2023-10-14
refactor two SAML tests to use page objects

swalchemist:
Convert 2 tests to use the page object strategy:
  incorrectResponseFromSamlIDP_showErrorFromSaml
  testSingleLogoutWithLogoutRedirect


hsinn0:
@swalchemist, have you looked into the check failures yet?

swalchemist:
> @swalchemist, have you looked into the check failures yet?

Checking.

swalchemist:
I removed the problematic test. All looks good and ready for review.

strehle (approved):

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/pageObjects/HomePage.java
```diff
@@ -16,6 +16,11 @@ public HomePage(WebDriver driver) {
         validatePageSource(driver, containsString("Where to?"));
     }
 
+    static public LoginPage goHome_redirectToLoginPage(WebDriver driver, String baseUrl) {
```
bruce-ricard: The function name is a bit hard to understand.

```
goHome
```

I'm already home. I WFH.

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/pageObjects/HomePage.java
```diff
@@ -16,6 +16,11 @@ public HomePage(WebDriver driver) {
...
+        driver.get(baseUrl);
+        return new LoginPage(driver);
```
bruce-ricard: This is also surprisingly hard to understand, despite its apparent simplicity. 

It appears that we are doing an HTTP get, but not doing anything to the result. And then returning the new login page. Which I'm guessing is _not_ redirecting us, but will redirect if the output of the function is called properly?

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/pageObjects/LoginPage.java
```diff
@@ -3,15 +3,15 @@
...
+import static org.hamcrest.Matchers.matchesPattern;
 
 public class LoginPage extends Page {
 
     static final protected String urlPath = "/login";
 
     public LoginPage(WebDriver driver) {
         super(driver);
-        validateUrl(driver, endsWith(urlPath));
+        validateUrl(driver, matchesPattern(".*" + urlPath + "(\\?.*)?$"));
```
bruce-ricard, line 14: I think that I would prefer

```
endsWith(urlPath + "(\\?.*)?")
```

is there a reason you moved away from `endsWith`?

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/pageObjects/LogoutDoEndpoint.java
```diff
@@ -0,0 +1,26 @@
...
+
+    static public LoginPage logout_goToLoginPage(WebDriver driver, String baseUrl, String redirectUrl, String clientId) {
+        driver.get(
+                baseUrl
+                + urlPath
+                + "?redirect=" + URLEncoder.encode(redirectUrl, StandardCharsets.UTF_8)
+                + "&client_id=" + clientId
+        );
+        return new LoginPage(driver);
+    }
```
bruce-ricard: I guess I just don't understand the pattern here.

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/pageObjects/Page.java
```diff
@@ -11,16 +13,29 @@ public class Page {
 
     public Page(WebDriver driver) {
         this.driver = driver;
+        driver.manage().timeouts().implicitlyWait(Duration.ofSeconds(5));
```
bruce-ricard, line 16: 🤔 

This looks like a 5-second wait on every page creation.

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/pageObjects/SamlErrorPage.java
```diff
@@ -0,0 +1,16 @@
+package org.cloudfoundry.identity.uaa.integration.pageObjects;
+
+import org.openqa.selenium.WebDriver;
+
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.endsWith;
+
+public class SamlErrorPage extends Page {
+    static final protected String urlPath = "/saml_error";
```
bruce-ricard: Is there a reason you make these constants `protected` instead of `private`, even though it appears nothing inherits the class?

I would advocate for the principle of least privilege.

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/pageObjects/LoginPage.java
swalchemist: I needed to change the matcher when I found that the URL sometimes has GET parameters in it. `endsWith()` does not accept a regex, just a literal string. I chose to use a regex instead of `contains()` so the matcher can be a bit more precise with matching the start of the parameters.

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/pageObjects/Page.java
swalchemist: This changes the default timeout of 0 (fail immediately when looking for an element if the page isn't completely loaded already) to a 5 second timeout. It does not introduce an arbitrary wait in the case of the element being found right away. 

This is an attempt to make the tests more robust and it should not have a significant impact on the execution time. It's most likely to affect the check in the page object constructors that validates that we're already on the expected page.

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/pageObjects/LogoutDoEndpoint.java
swalchemist: What part of it is confusing? The method name? The code in the body?

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/pageObjects/HomePage.java
swalchemist: The redirect is done by UAA, because we're not logged in. Would changing the method name from `goHome_redirectToLoginPage` to `goHome_redirectsToLoginPage` help any? The redirect action is a result of the get, not an explicit action taken by the test.

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/pageObjects/HomePage.java
bruce-ricard: I actually have a real point here. What is "home" here?

Isn't the home page the same as the login page? Or maybe only if you are logged out?

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/pageObjects/HomePage.java
bruce-ricard: I like the idea of "redirects". 

How do you know, in the context of this method, that we are not logged in? Could the method then be named something like `loggedOutHome` or maybe just `goToLoginPage`?

Why are you trying to document the 2 different levels of abstraction in the method name? So both `goHome` and `redirect...`?

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/pageObjects/LoginPage.java
bruce-ricard: > I found that the URL sometimes has GET parameters in it

I would like it if the code could convey that information

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/pageObjects/LogoutDoEndpoint.java
bruce-ricard: This one appears to make more sense, now that you explained.

So this method is to simulate clicking the "logout" button, which redirects you to the login page.

But the code doesn't seem to by logging anybody out.

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/pageObjects/Page.java
bruce-ricard: Got it. I wish this was clearly explained in the code, for everybody's knowledge. 

Personally, I would have made this change in its own commit, so that I could have added what you just wrote in the description.

Another idea is to abstract this code into a function with a name of the type "ensure that we wait long enough, to the limit of 5 seconds, for html elements to be populated, instead of immediately failing".

Though waiting for these things is generally considered an anti-pattern. We should really wait for the page to be fully loaded before starting testing on it.

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/pageObjects/SamlErrorPage.java
swalchemist: I think I had inheritance in mind, but as the limited use of inheritance so far doesn't need it, I have changed all of these constants to private.

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/pageObjects/HomePage.java
swalchemist: I added the result to the end of the method names as a result of an earlier code review. The return type indicates that we expect to end up on the login page, but some readers liked it better to also have that in the name, because the return type tends to be hidden in the chained calls that we typically used for these objects.

I have renamed the methods to try to indicate that the second part of the name is a result, not a direct action taken.

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/pageObjects/LogoutDoEndpoint.java
swalchemist: It's hitting logout.do. I extracted a method in hope of making that more clear.

It's not directly clicking the logout button, but taking the same sort of shortcut that the tests were using before the refactor, by hitting the endpoint directly.

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/pageObjects/HomePage.java
swalchemist: > How do you know, in the context of this method, that we are not logged in?

The test case know whether it has logged in or not, so I put this responsibility on the test case rather than the page objects.

bruce-ricard: But then you can't know that it "redirects to login page", can you?

swalchemist: That's an expectation that the test will verify.

PR #2546, opened 2023-10-16
feature: add change size to pull request

bruce-ricard:
* to be able to quickly see if a PR is going to take a long time to review or not

* The file was copied from this page https://github.com/marketplace/actions/pull-request-size-labeler


strehle (approved):
if our own review is faster then

PR #2549, opened 2023-10-18
fix: unnecessary method call

klaus-sap:
fix unnecessary method call in JdbcScimGroupMembershipManager


server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupMembershipManager.java
```diff
@@ -118,10 +118,9 @@ private Set<ScimGroup> getDefaultUserGroups(String zoneId) {
             return emptySet();
         }
         IdentityZone currentZone = IdentityZoneHolder.get();
-        List<String> zoneDefaultGroups = currentZone.getConfig().getUserConfig().getDefaultGroups();
-        if (!zoneId.equals(currentZone.getId())) {
-            zoneDefaultGroups = zoneProvisioning.retrieve(zoneId).getConfig().getUserConfig().getDefaultGroups();
-        }
+        List<String> zoneDefaultGroups = (zoneId.equals(currentZone.getId())) ?
```
strehle, line 121: I dont see the reason why one test fails, but you can generate docs with ./gradlew generateDocs . The voter https://github.com/cloudfoundry/uaa/blob/develop/.github/workflows/generate-api-docs.yml fails constantly

strehle (commented):
please check the failure

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupMembershipManager.java
klaus-sap: > Task :cloudfoundry-identity-uaa:gemInstallBundle
Ignoring eventmachine-1.2.7 because its extensions are not built. Try: gem pristine eventmachine --version 1.2.7
Ignoring http_parser.rb-0.6.0 because its extensions are not built. Try: gem pristine http_parser.rb --version 0.6.0
ERROR:  While executing gem ... (Gem::FilePermissionError)
    You don't have write permissions for the /Library/Ruby/Gems/2.6.0 directory.

klaus-sap: What might be the reason for this error?

strehle: this is a local issue, in case of problems install ruby with rvm https://rvm.io/ then you dont need root

strehle: use intellij and run docs tests manually, but only LogoutInfoEndpointDocs, maybe you have to change an assert there

LogoutInfoEndpointDocs > logout() FAILED
    java.lang.AssertionError at LogoutInfoEndpointDocs.java:44

restarted the test
https://github.com/cloudfoundry/uaa/actions/runs/6558516966/usage?pr=2549

klaus-sap: @strehle now the voter is green

strehle (approved):
sonar ok, then I am ok, 

https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2549

PR #2552, opened 2023-10-19
fix: define Java version in build.gradle

klaus-sap:
https://docs.gradle.org/current/javadoc/org/gradle/api/JavaVersion.html


strehle (approved):

strehle:
@Tallicia concerns from your side,

PR #2562, opened 2023-10-19
Bump: Java version to 17

peterhaochen47:
- build UAA with Java 17 (bumped from Java 11)
- fails with error if built with Java version lower than 17

[#186054019]

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -81,8 +81,8 @@ subprojects {
 
     [compileJava, compileTestJava]*.options*.compilerArgs = ["-Xlint:none", "-nowarn"]
 
-    sourceCompatibility = 1.11
-    targetCompatibility = 1.11
+    sourceCompatibility = 1.17
+    targetCompatibility = 1.17
 
     test {
         maxParallelForks = 1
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -81,8 +81,8 @@ subprojects {
 
     [compileJava, compileTestJava]*.options*.compilerArgs = ["-Xlint:none", "-nowarn"]
```
strehle, line 83: Can we have the constants from https://github.com/cloudfoundry/uaa/pull/2552/files ?
@peterhaochen47 we can merge https://github.com/cloudfoundry/uaa/pull/2552 or you can adopt, however @klaus-sap mention inconsistencies... I think JavaVersion.VERSION_11 is 11 simply and not 1.11, however if we use always constants it is solved

strehle (commented):
see inline

build.gradle
klaus-sap: JavaVersion.VERSION_11 is '1.11' (with quotation marks). Setting sourceCompatibility = 1.11 (w/o quotation marks) does not work on Mac, please merge https://github.com/cloudfoundry/uaa/pull/2552

build.gradle
```diff
@@ -81,8 +81,8 @@ subprojects {
...
 
-    sourceCompatibility = 1.11
-    targetCompatibility = 1.11
+    sourceCompatibility = 1.17
```
klaus-sap, line 84: JavaVersion.VERSION_17

klaus-sap: JavaVersion.VERSION_17

build.gradle
peterhaochen47: Hi, just made a commit about this.

strehle (approved):

klaus-sap (approved):

strehle (approved):

Tallicia (approved):
This looks good

Issue #2570, opened 2023-10-24
Create client with empty secret is not possible


strehle:
### What version of UAA are you running?

76.22.0

### How are you deploying the UAA?

I am deploying the UAA

- locally only using gradlew
- using a bosh release I downloaded from bosh.io
- using cf-release
- using cf-deployment
- as part of a commercial Cloud Foundry distribution

### What did you do?

uaac client add newclient --name xuaa --scope openid --authorized_grant_types password --autoapprove true --secret ''

error response:
{
  "error": "invalid_client",
  "error_description": "Client Secret must be at least 1 characters in length."
}

### What did you expect to see? What goal are you trying to achieve with the UAA?

Create a client with an empty secret via REST

### What did you see instead?

 "Client Secret must be at least 1 characters in length."

The client secret policy by default allow a min length of 0 but in 
https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/util/PasswordValidatorUtil.java#L55
This is prevented because this code "thinks" it runs for user passwords

peterhaochen47:
A similar discrepancy/confusion exists with the user password validation, in the context of uaa-release, where [uaa.password.policy.minLength is defaulted to 0](https://github.com/cloudfoundry/uaa-release/blob/b27bedbe7e5ff1768be28c29ffaab99731e1bc77/jobs/uaa/spec#L620), but in reality, UAA [does not allow zero-length user password](https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/util/PasswordValidatorUtil.java#L55). 

But the addition of this "not allow zero-length password" logic seems deliberate based on the [original tracker story](https://www.pivotaltracker.com/n/projects/997278/stories/136465473) associated with this restriction. Hence, we should investigate more before we attempt to unify the docs with the actual code. Perhaps the correct approach to eventually resolve this discrepancy is to, in a major release, maintain the API code logic (user/client cred needs to be at least 1 char), correct the docs/uaa-release config, and disallow empty string cred entirely.

strehle:
[...]

PR #2571, opened 2023-10-24
fix client admin rest call with empty secret

strehle:
Empty client secret is allowed if client secret policy allows it.
Error says 1 char is needed, this is wrong

Why 2 validators, distinguish between user and client policy validation.

For users 1 minimum char for a password might be ok, for a client definitely not.
The empty secret should be treated as public, but we cannot disallow it.

Therefore, this here is a fix for a regression introduced with 76.22.0.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/scim/validate/UaaPasswordPolicyValidator.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/scim/validate/UaaPasswordPolicyValidator.java
@@ -16,7 +16,7 @@ import java.util.LinkedList;
 import java.util.List;
 
 import static org.cloudfoundry.identity.uaa.util.PasswordValidatorUtil.messageResolver;
-import static org.cloudfoundry.identity.uaa.util.PasswordValidatorUtil.validator;
+import static org.cloudfoundry.identity.uaa.util.PasswordValidatorUtil.userValidator;
 
 /**
  * ****************************************************************************
@@ -62,7 +62,7 @@ public class UaaPasswordPolicyValidator implements PasswordValidator {
             policy = idpDefinition.getPasswordPolicy();
         }
 
-        org.passay.PasswordValidator validator = validator(policy, messageResolver);
+        org.passay.PasswordValidator validator = userValidator(policy, messageResolver);
         RuleResult result = validator.validate(new PasswordData((password != null) ? password : ""));
         if (!result.isValid()) {
             List<String> errorMessages = new LinkedList<>(validator.getMessages(result));
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/util/PasswordValidatorUtil.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/util/PasswordValidatorUtil.java
@@ -46,13 +46,23 @@ public final class PasswordValidatorUtil {
         }
     }
 
+    public static PasswordValidator secretValidator(GenericPasswordPolicy<?> policy,
+        MessageResolver messageResolver) {
+        return internalValidator(policy, 0, messageResolver);
+    }
 
-    public static PasswordValidator validator(GenericPasswordPolicy<?> policy,
+    public static PasswordValidator userValidator(GenericPasswordPolicy<?> policy,
+        MessageResolver messageResolver) {
+        return internalValidator(policy, 1, messageResolver);
+    }
+
+    private static PasswordValidator internalValidator(GenericPasswordPolicy<?> policy,
+                                              int minimumLength,
                                               MessageResolver messageResolver) {
         List<Rule> rules = new ArrayList<>();
 
-        //length is always a rule. We do not allow blank password
-        int minLength = Math.max(1, policy.getMinLength());
+        //length is always a rule. We do not allow blank password, but a blank secret
+        int minLength = Math.max(minimumLength, policy.getMinLength());
         int maxLength = policy.getMaxLength()>0 ? policy.getMaxLength() : Integer.MAX_VALUE;
         rules.add(new LengthRule(minLength, maxLength));
 
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/zone/ZoneAwareClientSecretPolicyValidator.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/zone/ZoneAwareClientSecretPolicyValidator.java
@@ -23,7 +23,7 @@ import java.util.LinkedList;
 import java.util.List;
 
 import static org.cloudfoundry.identity.uaa.util.PasswordValidatorUtil.messageResolver;
-import static org.cloudfoundry.identity.uaa.util.PasswordValidatorUtil.validator;
+import static org.cloudfoundry.identity.uaa.util.PasswordValidatorUtil.secretValidator;
 
 
 
@@ -82,8 +82,7 @@ public class ZoneAwareClientSecretPolicyValidator implements ClientSecretValidat
             clientSecretPolicy = zone.getConfig().getClientSecretPolicy();
         }
 
-        PasswordValidator clientSecretValidator = validator(clientSecretPolicy,
-                                                        messageResolver);
+        PasswordValidator clientSecretValidator = secretValidator(clientSecretPolicy, messageResolver);
         RuleResult result = clientSecretValidator.validate(new PasswordData(clientSecret));
         if (!result.isValid()) {
             List<String> errorMessages = new LinkedList<>(clientSecretValidator.getMessages(result));
--- a/server/src/test/java/org/cloudfoundry/identity/uaa/zone/ZoneAwareClientSecretPolicyValidatorTests.java
+++ b/server/src/test/java/org/cloudfoundry/identity/uaa/zone/ZoneAwareClientSecretPolicyValidatorTests.java
@@ -37,8 +37,14 @@ class ZoneAwareClientSecretPolicyValidatorTests {
     }
 
     @Test
-    void testEmptyClientSecret() {
+    void testEmptyClientSecretAllowed() {
         zone.getConfig().setClientSecretPolicy(defaultPolicy);
+        validator.validate(TEST_SECRET_1);
+    }
+
+    @Test
+    void testEmptyClientSecretForbidden() {
+        zone.getConfig().setClientSecretPolicy(strictPolicy);
         assertThrows(InvalidClientSecretException.class, () -> validator.validate(TEST_SECRET_1));
     }
 
--- a/server/src/test/java/org/cloudfoundry/identity/uaa/zone/ZoneAwareClientSecretPolicyValidatorTests.java
+++ b/server/src/test/java/org/cloudfoundry/identity/uaa/zone/ZoneAwareClientSecretPolicyValidatorTests.java
@@ -17,7 +17,7 @@ class ZoneAwareClientSecretPolicyValidatorTests {
     private ClientSecretPolicy defaultPolicy = new ClientSecretPolicy(0,255,0,0,0,0,6);
     private ClientSecretPolicy strictPolicy = new ClientSecretPolicy(6,10,1,1,1,1,6);
 
-    private static final String TEST_SECRET_1 = "";
+    private static final String TEST_EMPTY_SECRET = "";
     private static final String TEST_SECRET_2 = "testsecret";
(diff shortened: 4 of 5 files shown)


adrianhoelzl-sap (approved):

server/src/main/java/org/cloudfoundry/identity/uaa/scim/validate/UaaPasswordPolicyValidator.java
```diff
@@ -62,7 +62,7 @@ public void validate(String password) throws InvalidPasswordException {
             policy = idpDefinition.getPasswordPolicy();
         }
 
-        org.passay.PasswordValidator validator = validator(policy, messageResolver);
```
strehle, line 65: no logic change here, only renamed it to make clear, we have users here

server/src/main/java/org/cloudfoundry/identity/uaa/util/PasswordValidatorUtil.java
```diff
@@ -46,13 +46,23 @@ public static PropertiesMessageResolver messageResolver(String messagesResourceP
         }
     }
 
+    public static PasswordValidator secretValidator(GenericPasswordPolicy<?> policy,
```
strehle, line 49: created 2 methods to have it separated that for users a hard coded minimum password needs 1 char 
For clients this is not true we had and can have empty secrets.

The min length is validated with the client.secret.policy .... settings, but here the default for minLength is 0 , so should be alloed

strehle (commented):
inline

server/src/main/java/org/cloudfoundry/identity/uaa/util/PasswordValidatorUtil.java
```diff
@@ -46,13 +46,23 @@ public static PropertiesMessageResolver messageResolver(String messagesResourceP
...
+    }
 
-    public static PasswordValidator validator(GenericPasswordPolicy<?> policy,
+    private static PasswordValidator internalValidator(GenericPasswordPolicy<?> policy,
+                                              int minimumLength,
                                               MessageResolver messageResolver) {
         List<Rule> rules = new ArrayList<>();
 
-        //length is always a rule. We do not allow blank password
-        int minLength = Math.max(1, policy.getMinLength());
```
bruce-ricard, line 55: I don't understand this line. This seems to be the source of the problem. Why are we overriding the policy with this 1? This is what needs to be changed. If we want to enforce a minimum 1 policy, it should be in the policy.

server/src/main/java/org/cloudfoundry/identity/uaa/util/PasswordValidatorUtil.java
```diff
@@ -46,13 +46,23 @@ public static PropertiesMessageResolver messageResolver(String messagesResourceP
...
+        //length is always a rule. We do not allow blank password, but a blank secret
+        int minLength = Math.max(minimumLength, policy.getMinLength());
```
bruce-ricard, line 65: It really looks like this code should be in the `PasswordValidator` class.

server/src/main/java/org/cloudfoundry/identity/uaa/zone/ZoneAwareClientSecretPolicyValidator.java
```diff
@@ -82,8 +82,7 @@ public void validate(String clientSecret) throws InvalidClientSecretException {
             clientSecretPolicy = zone.getConfig().getClientSecretPolicy();
         }
 
-        PasswordValidator clientSecretValidator = validator(clientSecretPolicy,
```
bruce-ricard, line 85: Is this the source of the regression that was introduced recently? That we started validating client secrets?

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/ClientAdminEndpointsIntegrationTests.java
```diff
@@ -170,6 +171,21 @@ public void createClientWithSecondarySecret() {
         assertEquals(HttpStatus.CREATED, result.getStatusCode());
     }
 
+    @Test
+    public void createClientWithEmptySecret() {
```
bruce-ricard, line 175: Is this test the one that causes the fix in the second commit of this PR?

If that's the case, could both changes be in the same commit? It's a lot easier to follow when a code change and the tests that go with it are together in the same commit.

server/src/test/java/org/cloudfoundry/identity/uaa/zone/ZoneAwareClientSecretPolicyValidatorTests.java
```diff
@@ -39,6 +39,8 @@ void setUp() {
     @Test
     void testEmptyClientSecret() {
         zone.getConfig().setClientSecretPolicy(defaultPolicy);
+        validator.validate(TEST_SECRET_1);
```
bruce-ricard: Is `TEST_SECRET_1` an empty secret? I don't understand this test if it isn't.

server/src/test/java/org/cloudfoundry/identity/uaa/zone/ZoneAwareClientSecretPolicyValidatorTests.java
```diff
@@ -39,6 +39,8 @@ void setUp() {
     @Test
```
bruce-ricard, line 39: I really don't understand the commit message of this commit:

> allow empty secret int test with default policy because minlength is …
…0, only with strict it should fail

Also, if you could try to have a short title of <50 characters, and then you can add whatever description you want in the commit body.

bruce-ricard (changes requested):

server/src/main/java/org/cloudfoundry/identity/uaa/scim/validate/UaaPasswordPolicyValidator.java
bruce-ricard: When you feel like you have to clarify things in github comments, IMO it means that you should have clarified them in the code. Or through git commits.

For example, what I would have done in this specific case, is to made a separate commit for the renaming of the function that I would title: "refactor: rename function". This way it is obvious to everyone that it is a refactor. Not only to people reading the PR.

server/src/main/java/org/cloudfoundry/identity/uaa/util/PasswordValidatorUtil.java
strehle: Please check the history and you can also use an older UAA, but even if you run default client policy, means 0 as minimum secret is allowed you cannot have 

uaac client add newclient --name xuaa --scope openid --authorized_grant_types password --autoapprove true --secret ''


We agree that the empty secret still should be allowed but with hard coded 1 this does not work.

For user passwords maybe and therefore I have divided the check into user and client validator.

server/src/main/java/org/cloudfoundry/identity/uaa/util/PasswordValidatorUtil.java
strehle: the code is in the class. , 

I dont was change an existing beviour. Existing was. For a  user password at least 1 character is needed.

server/src/main/java/org/cloudfoundry/identity/uaa/zone/ZoneAwareClientSecretPolicyValidator.java
strehle: no, the issue is seperate.

As mentioned before use an older UAA start it and run uaac to create an client with empty secret......

server/src/test/java/org/cloudfoundry/identity/uaa/zone/ZoneAwareClientSecretPolicyValidatorTests.java
strehle: yes it is empty and the test was wrong before... and this is independent from all other changes related to empty secret.

Here we simply have the issue, that via REST the client creation with empty secret was never possible.... or lets says... since this valdation here... because in YAML you have client.secret.policy ...minLenght: 0, but in code it then wants 1 as mimimum.

server/src/test/java/org/cloudfoundry/identity/uaa/zone/ZoneAwareClientSecretPolicyValidatorTests.java
strehle: I committed first the test only
https://github.com/cloudfoundry/uaa/pull/2571/commits/66841d6e4ce6187ec6a06a547b7e9f4192526b92
Then the fix
https://github.com/cloudfoundry/uaa/pull/2571/commits/66841d6e4ce6187ec6a06a547b7e9f4192526b92

And then found out that the even the test was wrong
https://github.com/cloudfoundry/uaa/pull/2571/commits/66841d6e4ce6187ec6a06a547b7e9f4192526b92

server/src/main/java/org/cloudfoundry/identity/uaa/util/PasswordValidatorUtil.java
bruce-ricard: > For a user password at least 1 character is needed

This is ridiculous. A 1-character password and an empty password have literally the same level of security: none.
This line is completely useless.

server/src/main/java/org/cloudfoundry/identity/uaa/zone/ZoneAwareClientSecretPolicyValidator.java
bruce-ricard: I don't want to reproduce the issue. I'm trying to find out what new code is the source of the problem. It seems to me like it should be documented in the fix, so that people understand what is going on.

server/src/test/java/org/cloudfoundry/identity/uaa/zone/ZoneAwareClientSecretPolicyValidatorTests.java
bruce-ricard: I understand that. What I'm saying is that I'm looking at this test called `testEmptyClientSecret` that you are modifying, and I cannot figure out where the empty client secret actually is.

Could you either rename the constant to be called `EMPTY_SECRET` or something so that we understand what the test is doing?

bruce-ricard: It's also not obvious to understand that this line is actually testing something. It really looks like some sort of test setup. People could miss that we are actually expecting this call not to throw any exception. 

The typical way of doing what this test is doing is to create 2 different tests. One that tests with the default policy, and one that tests with the strict policy.

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/ClientAdminEndpointsIntegrationTests.java
strehle: I always do a squash commit because this is my way the last years... I do many commits. but if a featue/fix is ready, then I squash it into one commit and merge to repo only the consolidated version, .... I can squash everything now if you dont like the small commits

server/src/test/java/org/cloudfoundry/identity/uaa/zone/ZoneAwareClientSecretPolicyValidatorTests.java
strehle: squashed now, close this here

server/src/test/java/org/cloudfoundry/identity/uaa/zone/ZoneAwareClientSecretPolicyValidatorTests.java
strehle: split tests and renamed variable

strehle: rename TEST_EMPTY_SECRET

server/src/main/java/org/cloudfoundry/identity/uaa/zone/ZoneAwareClientSecretPolicyValidator.java
strehle: well, after the fix, the documentation matches again, because it says
minLength defaults to 0, e.g. https://docs.cloudfoundry.org/api/uaa/version/76.23.0/index.html#identity-zones
But this was not the case

strehle:
close this, related issue was https://github.com/cloudfoundry/uaa/issues/2570
do not use it, because we would have to touch the validator which is currently in-consistent from point of was documentations states and what it does in real

PR #2572, opened 2023-10-24
fix client admin rest call with empty secret - alternative PR

strehle:
This PR simply uses PasswordValidatorUtil but set the password minLength to 1 which was the default because it was in code The minLength for clients should be 0 and this was the case before

Same Tests as before, means, with Default Policy the empty secret is allowed, with the strictPolicy not, see testEmptyClientSecret

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/util/PasswordValidatorUtil.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/util/PasswordValidatorUtil.java
@@ -51,8 +51,8 @@ public final class PasswordValidatorUtil {
                                               MessageResolver messageResolver) {
         List<Rule> rules = new ArrayList<>();
 
-        //length is always a rule. We do not allow blank password
-        int minLength = Math.max(1, policy.getMinLength());
+        //length is always a rule
+        int minLength = Math.max(0, policy.getMinLength());
         int maxLength = policy.getMaxLength()>0 ? policy.getMaxLength() : Integer.MAX_VALUE;
         rules.add(new LengthRule(minLength, maxLength));
 
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/util/PasswordValidatorUtil.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/util/PasswordValidatorUtil.java
@@ -51,8 +51,7 @@ public final class PasswordValidatorUtil {
                                               MessageResolver messageResolver) {
         List<Rule> rules = new ArrayList<>();
 
-        //length is always a rule
-        int minLength = Math.max(0, policy.getMinLength());
+        int minLength = policy.getMinLength()>0 ? policy.getMinLength() : 0;
         int maxLength = policy.getMaxLength()>0 ? policy.getMaxLength() : Integer.MAX_VALUE;
         rules.add(new LengthRule(minLength, maxLength));
 
--- a/server/src/test/java/org/cloudfoundry/identity/uaa/zone/ZoneAwareClientSecretPolicyValidatorTests.java
+++ b/server/src/test/java/org/cloudfoundry/identity/uaa/zone/ZoneAwareClientSecretPolicyValidatorTests.java
@@ -37,13 +37,13 @@ class ZoneAwareClientSecretPolicyValidatorTests {
     }
 
     @Test
-    void testEmptyClientSecretAllowed() {
+    void defaultPolicyAcceptsEmptySecret() {
         zone.getConfig().setClientSecretPolicy(defaultPolicy);
         validator.validate(TEST_EMPTY_SECRET);
     }
 
     @Test
-    void testEmptyClientSecretForbidden() {
+    void strictPolicyRejectsEmptySecret() {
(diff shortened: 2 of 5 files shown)


server/src/main/java/org/cloudfoundry/identity/uaa/util/PasswordValidatorUtil.java
```diff
@@ -51,8 +51,8 @@ public static PasswordValidator validator(GenericPasswordPolicy<?> policy,
                                               MessageResolver messageResolver) {
         List<Rule> rules = new ArrayList<>();
 
-        //length is always a rule. We do not allow blank password
-        int minLength = Math.max(1, policy.getMinLength());
+        //length is always a rule
+        int minLength = Math.max(0, policy.getMinLength());
```
bruce-ricard: Is there a reason you didn't just delete the line?

`minLength` is now exactly `policy.getMinLength()`.

server/src/test/java/org/cloudfoundry/identity/uaa/zone/ZoneAwareClientSecretPolicyValidatorTests.java
```diff
@@ -37,9 +37,15 @@ void setUp() {
     }
 
     @Test
-    void testEmptyClientSecret() {
+    void testEmptyClientSecretAllowed() {
```
bruce-ricard: Could we name this test:

```
testEmptyClientSecretAllowedForBackwardsCompatibility
```

Because we don't really _want_ this feature. We need to keep it for backwards compatibility. Correct?

uaa/src/main/webapp/WEB-INF/spring/scim-endpoints.xml
```diff
@@ -35,7 +35,7 @@
     </bean>
 
     <bean id="globalPasswordPolicy" class="org.cloudfoundry.identity.uaa.provider.PasswordPolicy">
-        <constructor-arg name="minLength" value="${password.policy.global.minLength:0}"/>
+        <constructor-arg name="minLength" value="${password.policy.global.minLength:1}"/>
```
strehle: Documentation says 0 is allowed , really ?

strehle:
After thinking about I see that minLength for passwords and secrets is 0 by default so then I would go with it, if you have no concerns and completly rely on the PasswordValidor rules, means I will update again to minLength:0

server/src/test/java/org/cloudfoundry/identity/uaa/zone/ZoneAwareClientSecretPolicyValidatorTests.java
```diff
@@ -37,9 +37,15 @@ void setUp() {
...
         zone.getConfig().setClientSecretPolicy(defaultPolicy);
-        assertThrows(InvalidClientSecretException.class, () -> validator.validate(TEST_SECRET_1));
+        validator.validate(TEST_EMPTY_SECRET);
+    }
+
+    @Test
+    void testEmptyClientSecretForbidden() {
```
bruce-ricard: It's a bit odd to have 2 tests that have opposite descriptions.

Now we have a test that says "empty client secret is allowed" and one that says "empty client secret is forbidden".

Maybe `defaultPolicyAcceptsEmptySecret` and `strictPolicyRejectsEmptySecret` ?

uaa/src/main/webapp/WEB-INF/spring/scim-endpoints.xml
bruce-ricard: I still don't understand this change. I don't care too much about it, but requiring a password of minimum length 1 is odd.

We can leave it as is for now.

bruce-ricard (changes requested):
Thanks Markus. This is IMO a lot better. I just have a couple small comments.

uaa/src/main/webapp/WEB-INF/spring/scim-endpoints.xml
bruce-ricard: OK I see that you reverted that change. LGTM.

If someone wants to enforce more characters, they can do that by setting the policy.

peterhaochen47:
This PR is indeed simpler than the first #2571, but it does change the current behavior of UAA (and the 4 failing tests, before [this commit](https://github.com/cloudfoundry/uaa/pull/2572/commits/b30c52f7e1e7f0e2802c9871ed3c50a8567fedc7) was made to remove/modify the existing tests, are evidence). The behavior in question (not allowing user password to be an empty string) seems contradictory to [our uaa-release doc](https://github.com/cloudfoundry/uaa-release/blob/b27bedbe7e5ff1768be28c29ffaab99731e1bc77/jobs/uaa/spec#L620), but its addition seems deliberate based on the [original tracker story](https://www.pivotaltracker.com/n/projects/997278/stories/136465473) associated with this restriction. Hence, I suspect that this change (removing the restriction) might introduce a new regression (possibly the one that the original tracker story tries to address).

strehle:
close this , now, was related to https://github.com/cloudfoundry/uaa/issues/2570 but no successful tests

PR #2573, opened 2023-10-24
fix: cve-scanner failure

hsinn0:
- It was finding CVEs in slate folder, which is for API doc generation.
- We do not generate API docs for v74.5.x anymore.
- So we removed the source codes for API doc generation in this branch.

[#186037475]


bruce-ricard (approved):
LGTM

PR #2574, opened 2023-10-25
validate secrets only with text

### LINKED ISSUE #2570: Create client with empty secret is not possible

strehle:
This fixes client creation rest call with empty secret.
Empty client secret is allowed via YAML setting already, but
in a REST call there is an error:
Client Secret must be at least 1 characters in length.

Why this occurs: There is a policy validator for user and client
policy validation.

For users, a minimum of 1 char for a password might be ok,
for a client not. A secret can be empty.

Before 76.22.0 a missing secret in a client creation call was defaulted
to an empty secret, but with https://github.com/cloudfoundry/uaa/pull/2455
this was fixed. The fix prevented the creation with an empty secret.

Therefore, this here is a fix for a regression introduced with 76.22.0.
It simply prevents the policy validation if the secret is without text
(null or empty).

Fix for issue https://github.com/cloudfoundry/uaa/issues/2570

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/zone/ZoneAwareClientSecretPolicyValidator.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/zone/ZoneAwareClientSecretPolicyValidator.java
@@ -18,6 +18,7 @@ import org.passay.PasswordData;
 import org.passay.PasswordValidator;
 import org.passay.PropertiesMessageResolver;
 import org.passay.RuleResult;
+import org.springframework.util.StringUtils;
 
 import java.util.LinkedList;
 import java.util.List;
@@ -71,7 +72,7 @@ public class ZoneAwareClientSecretPolicyValidator implements ClientSecretValidat
 
     @Override
     public void validate(String clientSecret) throws InvalidClientSecretException {
-        if(clientSecret == null) {
+        if(!StringUtils.hasText(clientSecret)) {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


peterhaochen47:
@strehle In https://github.com/cloudfoundry/uaa/issues/2570, you mentioned that the empty string client secret is not allowed in API because of [this code](https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/util/PasswordValidatorUtil.java#L55). So I'm a little confused on why this PR would solve your problem without touching that code?

strehle:
> @strehle In #2570, you mentioned that the empty string client secret is not allowed in API because of [this code](https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/util/PasswordValidatorUtil.java#L55). So I'm a little confused on why this PR would solve your problem without touching that code?

With this PR we do not reach this code

peterhaochen47 (changes requested):
The general approach is fine. 

Please rewrite the commit to include all the context/motivations behind the change such that future developers would understand why the change was made.

peterhaochen47 (approved):

PR #2575, opened 2023-10-25
return unauthorized_client in oauth2 error code

### LINKED ISSUE #2545: Error response code not compliant to RFC 6749

76.22.0 (latest)

as part of a commercial Cloud Foundry distribution

### What did you do?

I am trying to generate access tokens using the token endpoint - `{authenticationUrl}/oauth/token`

with invalid client credentials, the error response was expected to be as follows:
- Code: 401 Unauthorized
- Body:
  **- “error”: “invalid_grant”**
  - “error_description”:  “Bad credentials”

### What did you see instead?

with invalid client credentials, the error response is as follows:

- Code: 401 Unauthorized
- Body:
  **- “error”: “unauthorized”**
  - “error_description”:  “Bad credentials”

**Question 1**: in [RFC6749 section 5.2](https://www.rfc-editor.org/rfc/rfc6749#section-5.2) -  “unauthorized” is not in the list of allowed ‘error codes’. We should use an error response code from this available list of codes.
**Question 2**: in [RFC6749 section 5.2](https://www.rfc-editor.org/rfc/rfc6749#section-5.2) - suggests a 400 unless specified otherwise. Any reason why we use a 401?

strehle:
return unauthorized_client in oauth2 error response 

Fix for issue https://github.com/cloudfoundry/uaa/issues/2545

### DIFF
--- a/uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/token/TokenMvcMockTests.java
+++ b/uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/token/TokenMvcMockTests.java
@@ -858,7 +858,7 @@ public class TokenMvcMockTests extends AbstractTokenMockMvcTests {
                 .andExpect(
                         header()
                                 .stringValues("WWW-Authenticate",
-                                        "Basic realm=\"UAA/client\", error=\"unauthorized\", error_description=\"Bad credentials\"")
+                                        "Basic realm=\"UAA/client\"")
                 );
     }
 
@@ -3530,7 +3530,7 @@ public class TokenMvcMockTests extends AbstractTokenMockMvcTests {
                 .param(OAuth2Utils.GRANT_TYPE, "password")
                 .param(OAuth2Utils.CLIENT_ID, clientId))
                 .andExpect(status().isUnauthorized())
-                .andExpect(content().string("{\"error\":\"unauthorized\",\"error_description\":\"Bad credentials\"}"));
+                .andExpect(content().string("{\"error\":\"unauthorized_client\",\"error_description\":\"Bad credentials\"}"));
     }
 
     @Test
--- a/uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/token/TokenMvcMockTests.java
+++ b/uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/token/TokenMvcMockTests.java
@@ -858,7 +858,7 @@ public class TokenMvcMockTests extends AbstractTokenMockMvcTests {
                 .andExpect(
                         header()
                                 .stringValues("WWW-Authenticate",
-                                        "Basic realm=\"UAA/client\"")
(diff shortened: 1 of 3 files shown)


uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/token/TokenMvcMockTests.java
```diff
@@ -858,7 +858,7 @@ void token_endpoint_should_return_Basic_WWW_Authenticate_Header() throws Excepti
                 .andExpect(
                         header()
                                 .stringValues("WWW-Authenticate",
-                                        "Basic realm=\"UAA/client\", error=\"unauthorized\", error_description=\"Bad credentials\"")
+                                        "Basic realm=\"UAA/client\"")
```
Tallicia: Is this meant to remove? Instead of replace or adjust the error description?

Tallicia (changes requested):
I want to validate the one question inline before approving.

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/token/TokenMvcMockTests.java
strehle: solved

strehle:
@bruce-ricard since you are active on some issues, can you please check if you see an issue with this, I would like to solve this for next release

neelalex:
@strehle Should this be `invalid_client` as you mentioned in the ticket?

strehle:
> @strehle Should this be `invalid_client` as you mentioned in the ticket?

not necessarily, but currently used unauthorized_client to have it similar to non stand unauthorized code.

My tests with others have shown, that they use invalid_client, but unauthorized_client  is also oauth2 complaint.

But therefore asked the other UAA maintainer for their thouths

neelalex:
> > @strehle Should this be `invalid_client` as you mentioned in the ticket?
> 
> not necessarily, but currently used unauthorized_client to have it similar to non stand unauthorized code.
> 
> My tests with others have shown, that they use invalid_client, but unauthorized_client is also oauth2 complaint.
> 
> But therefore asked the other UAA maintainer for their thouths

The issue addressed in this ticket is **with invalid client credentials**, the error response was expected to be as follows as per https://www.rfc-editor.org/rfc/rfc6749#section-5.2

Code: 401 Unauthorized
Body:
**“error”: “invalid_client”**
“error_description”: “Bad credentials”


whereas ->  unauthorized_client : The authenticated client is not authorized to use this authorization grant type.
This means  this error occurs when the client is **not authorized to request the requested resource, even if the client credentials are valid**. It typically indicates that the client does not have the necessary permissions or scope to access the requested resource.

Tallicia:
Thank you for clarifying this @neelalex

strehle:
close this, because we have clarifid we go with the alternative PR https://github.com/cloudfoundry/uaa/pull/2596

Issue #2550, opened 2023-10-18
CF UAA support for SCIM2.0 as per https://simplecloud.info/#Implementations2


Rohit04061992:
[no description provided]

Rohit04061992:
I am using latest uaa version 76.22.0

strehle:
You created an isssue but this is a feature request and I doubt that we will provide full 2.0 support with one fix, so do have a concrete issue missing SCIM2.0 where a small or medium fix would help. 
And if you would provide such a PR we could solve this in a shorter time frame

Rohit04061992:
We are trying to configure , ad sync with cf uaa , and when azure ad is invoking uaa apis for syncing the users and groups , uaa is giving error which says it only supports scim1.0 schema

Rohit04061992:
In cf uaa for create user api schema is optional , but when someone passes the schema as scim2.0 system breaks , scim 2.0 is defacto standard now , cf uaa should enhance to support this .

@strehle We need basic scim2.0 support for adding users and groups .

Rohit04061992:
Hello @strehle ,
We would like CF UAA to get aligned with https://datatracker.ietf.org/doc/html/rfc7644, so that it can be used with modern IDPS which has capability to sync user in uaa using scim endpoints .

strehle:
@emalm @Tallicia @torsten-sap  FYI.

Agree that SCIM 2.0 is standard, therefore added PM to discuss the prio. 

UAA uses currently https://github.com/pingidentity/scim 
With https://github.com/pingidentity/scim2 an adoption to /v2/Users, /v2/Groups etc. could happen without that much effort, but I have not verified it.

@Rohit04061992 If you would be able to provide a PR it would increase the speed here.

Rohit04061992:
Hi @strehle , 

Yes i have also landed up till https://github.com/pingidentity/scim2, was trying to integrate this as well, but if this comes rom the community it will be best .

Rohit04061992:
@strehle  I have initiated a PR at this location https://github.com/cloudfoundry/uaa/pull/2583
But changes to scim sdk is breaking most of the scim based implementation, which makes sense , but this also means major changes in uaa .

Rohit04061992:
Hi @emalm ,
Can you take this , i feel that this enhancement would be of great value for uaa as a product as well.

strehle:
duplicate to https://github.com/cloudfoundry/uaa/issues/2976

PR #2583, opened 2023-10-30
Update dependencies.gradle to support Scim2.0

Rohit04061992:
Bumped up scim sdk version to support scim2 protocol


strehle:
This approach wont work.
We have to add the new library and new REST endpoints needed to ensure scim 1.1 with existining endpoints

swalchemist:
Closing this as it's not a complete solution and the CLA was not accepted.

We'll still track the SCIM 2 request in  #2550.

PR #2588, opened 2023-11-03
Add label text to all form controls

ystros:
Previously, placeholder text has been used to represent the label for form controls. This allows visual users to tell what content to enter in specific form controls. However, screen readers generally expect an explicit label to be associated with form controls.

Rather redesign all pages to add `<label>` elements for each control, this adds the `aria-label` attribute to each form control. This should allow screen readers to properly announce controls as the page is read.


strehle:
@Tallicia can you retrigger concourse ?

Tallicia:
It looks like these need to be created with local braches in this repo.

strehle (approved):
LGTM

PR #2590, opened 2023-11-06
Add config to control whether to perform "OpenID Connect RP-Initiated Logout" when using an external OIDC provider

### LINKED ISSUE #2589: No config to turn off "OpenID Connect RP-Initiated Logout" when using an external OIDC provider

Latest UAA v76.25.0

I am deploying the UAA

- locally only using gradlew
- using cf-deployment

### What did you do?
* Configure UAA to use an [external OIDC provider](https://docs.cloudfoundry.org/api/uaa/version/76.25.0/index.html#oauth-oidc)
* Log into UAA via the external OIDC login
* Click UAA logout button
* See that the user is redirected to the external OIDC provider logout endpoint

This "OpenID Connect RP-Initiated Logout" behavior is sometimes not what UAA operators want. [This Okta discussion](https://developer.okta.com/blog/2020/03/27/spring-oidc-logout-options) talks about how whether an operator wants this behavior highly depends on their use cases (in general, they likely do not want this under a one-OP-multiple-RPs system). 

Currently, UAA offers no config (API, uaa bosh release, or uaa.yml) to opt out of this behavior.

### Proposal
Add an [external OIDC provider](https://docs.cloudfoundry.org/api/uaa/version/76.25.0/index.html#oauth-oidc) config field that controls whether "OpenID Connect RP-Initiated Logout" is performed after logging out of UAA. Default to the current behavior (does perform the RP-Initiated Logout) to avoid a breaking change.

### Prior context
Equivalent issue on the uaa-release side: https://github.com/cloudfoundry/uaa-release/issues/701 (see discussion there).

peterhaochen47:
Fixes https://github.com/cloudfoundry/uaa/issues/2589

Notes for reviewer:
* refactors of the impacted classes/code path will be in a different PR, so this PR would be easier to review.
* reasons for the config name `performRpInitiatedLogout`:
  * The exact language of [the spec](https://openid.net/specs/openid-connect-rpinitiated-1_0.html)
  * Follow the pattern of the other boolean configs of this endpoint: starting with a verb
  * alternatives considered: `SSOLogout`, `performSSOLogoutOnIdpAfterUAALogout`, `triggerExternalOIDCProviderLogout` etc.

### DIFF
--- a/model/src/main/java/org/cloudfoundry/identity/uaa/provider/AbstractExternalOAuthIdentityProviderDefinition.java
+++ b/model/src/main/java/org/cloudfoundry/identity/uaa/provider/AbstractExternalOAuthIdentityProviderDefinition.java
@@ -46,6 +46,7 @@ public abstract class AbstractExternalOAuthIdentityProviderDefinition<T extends 
     private String userPropagationParameter;
     private OAuthGroupMappingMode groupMappingMode;
     private boolean pkce = true;
+    private boolean performRpInitiatedLogout = true;
 
     public URL getAuthUrl() {
         return authUrl;
@@ -214,6 +215,14 @@ public abstract class AbstractExternalOAuthIdentityProviderDefinition<T extends 
         return this.pkce;
     }
 
+    public boolean isPerformRpInitiatedLogout() {
+        return performRpInitiatedLogout;
+    }
+
+    public void setPerformRpInitiatedLogout(boolean performRpInitiatedLogout) {
+        this.performRpInitiatedLogout = performRpInitiatedLogout;
+    }
+
     @JsonIgnore
     public Class getParameterizedClass() {
         ParameterizedType parameterizedType =
@@ -247,6 +256,7 @@ public abstract class AbstractExternalOAuthIdentityProviderDefinition<T extends 
         if (!Objects.equals(userPropagationParameter, that.userPropagationParameter)) return false;
         if (!Objects.equals(groupMappingMode, that.groupMappingMode)) return false;
         if (pkce != that.pkce) return false;
+        if (performRpInitiatedLogout != that.performRpInitiatedLogout) return false;
         return Objects.equals(responseType, that.responseType);
 
     }
@@ -271,6 +281,7 @@ public abstract class AbstractExternalOAuthIdentityProviderDefinition<T extends 
         result = 31 * result + (groupMappingMode != null ? groupMappingMode.hashCode() : 0);
         result = 31 * result + (responseType != null ? responseType.hashCode() : 0);
         result = 31 * result + (pkce ? 1 : 0);
+        result = 31 * result + (performRpInitiatedLogout ? 1 : 0);
         return result;
     }
 }
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/ZoneAwareWhitelistLogoutHandler.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/ZoneAwareWhitelistLogoutHandler.java
@@ -47,11 +47,12 @@ public class ZoneAwareWhitelistLogoutHandler implements LogoutSuccessHandler {
     public void onLogoutSuccess(HttpServletRequest request, HttpServletResponse response, Authentication authentication) throws IOException, ServletException {
         AbstractExternalOAuthIdentityProviderDefinition oauthConfig = externalOAuthLogoutHandler.getOAuthProviderForAuthentication(authentication);
         String logoutUrl = externalOAuthLogoutHandler.getLogoutUrl(oauthConfig);
+        Boolean shouldPerformRpInitiatedLogout = externalOAuthLogoutHandler.getPerformRpInitiatedLogout(oauthConfig);
 
-        if (logoutUrl == null) {
-            getZoneHandler().onLogoutSuccess(request, response, authentication);
-        } else {
+        if (shouldPerformRpInitiatedLogout && logoutUrl != null) {
             externalOAuthLogoutHandler.onLogoutSuccess(request, response, authentication);
+        } else {
+            getZoneHandler().onLogoutSuccess(request, response, authentication);
         }
     }
 
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/ZoneAwareWhitelistLogoutHandler.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/ZoneAwareWhitelistLogoutHandler.java
@@ -47,7 +47,7 @@ public class ZoneAwareWhitelistLogoutHandler implements LogoutSuccessHandler {
     public void onLogoutSuccess(HttpServletRequest request, HttpServletResponse response, Authentication authentication) throws IOException, ServletException {
         AbstractExternalOAuthIdentityProviderDefinition oauthConfig = externalOAuthLogoutHandler.getOAuthProviderForAuthentication(authentication);
         String logoutUrl = externalOAuthLogoutHandler.getLogoutUrl(oauthConfig);
-        Boolean shouldPerformRpInitiatedLogout = externalOAuthLogoutHandler.getPerformRpInitiatedLogout(oauthConfig);
+        boolean shouldPerformRpInitiatedLogout = externalOAuthLogoutHandler.getPerformRpInitiatedLogout(oauthConfig);
 
         if (shouldPerformRpInitiatedLogout && logoutUrl != null) {
             externalOAuthLogoutHandler.onLogoutSuccess(request, response, authentication);
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/ZoneAwareWhitelistLogoutHandler.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/ZoneAwareWhitelistLogoutHandler.java
@@ -17,6 +17,7 @@ package org.cloudfoundry.identity.uaa.authentication;
 
 import org.cloudfoundry.identity.uaa.oauth.KeyInfoService;
 import org.cloudfoundry.identity.uaa.provider.AbstractExternalOAuthIdentityProviderDefinition;
+import org.cloudfoundry.identity.uaa.provider.OIDCIdentityProviderDefinition;
 import org.cloudfoundry.identity.uaa.provider.oauth.ExternalOAuthLogoutHandler;
 import org.cloudfoundry.identity.uaa.zone.MultitenantClientServices;
 import org.cloudfoundry.identity.uaa.zone.IdentityZoneConfiguration;
@@ -45,7 +46,7 @@ public class ZoneAwareWhitelistLogoutHandler implements LogoutSuccessHandler {
 
     @Override
     public void onLogoutSuccess(HttpServletRequest request, HttpServletResponse response, Authentication authentication) throws IOException, ServletException {
-        AbstractExternalOAuthIdentityProviderDefinition oauthConfig = externalOAuthLogoutHandler.getOAuthProviderForAuthentication(authentication);
+        AbstractExternalOAuthIdentityProviderDefinition<OIDCIdentityProviderDefinition> oauthConfig = externalOAuthLogoutHandler.getOAuthProviderForAuthentication(authentication);
         String logoutUrl = externalOAuthLogoutHandler.getLogoutUrl(oauthConfig);
         boolean shouldPerformRpInitiatedLogout = externalOAuthLogoutHandler.getPerformRpInitiatedLogout(oauthConfig);
 
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthLogoutHandler.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthLogoutHandler.java
@@ -124,7 +124,7 @@ public class ExternalOAuthLogoutHandler extends SimpleUrlLogoutSuccessHandler {
     return config.getLinks().getLogout().getRedirectUrl();
   }
 
-  public boolean getPerformRpInitiatedLogout(AbstractExternalOAuthIdentityProviderDefinition oauthConfig) {
+  public boolean getPerformRpInitiatedLogout(AbstractExternalOAuthIdentityProviderDefinition<OIDCIdentityProviderDefinition> oauthConfig) {
     if (oauthConfig == null) {
       return false;
     }
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthLogoutHandler.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthLogoutHandler.java
@@ -123,4 +123,11 @@ public class ExternalOAuthLogoutHandler extends SimpleUrlLogoutSuccessHandler {
     }
     return config.getLinks().getLogout().getRedirectUrl();
   }
+
+  public Boolean getPerformRpInitiatedLogout(AbstractExternalOAuthIdentityProviderDefinition oauthConfig) {
+    if (oauthConfig == null) {
+      return false;
+    }
+    return oauthConfig.isPerformRpInitiatedLogout();
+  }
 }
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthLogoutHandler.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthLogoutHandler.java
@@ -124,7 +124,7 @@ public class ExternalOAuthLogoutHandler extends SimpleUrlLogoutSuccessHandler {
     return config.getLinks().getLogout().getRedirectUrl();
   }
 
-  public Boolean getPerformRpInitiatedLogout(AbstractExternalOAuthIdentityProviderDefinition oauthConfig) {
+  public boolean getPerformRpInitiatedLogout(AbstractExternalOAuthIdentityProviderDefinition oauthConfig) {
     if (oauthConfig == null) {
       return false;
     }
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/OauthIDPWrapperFactoryBean.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/OauthIDPWrapperFactoryBean.java
@@ -166,6 +166,9 @@ public class OauthIDPWrapperFactoryBean {
         if (idpDefinitionMap.get("clientAuthInBody") instanceof Boolean) {
             idpDefinition.setClientAuthInBody((boolean)idpDefinitionMap.get("clientAuthInBody"));
         }
+        if (idpDefinitionMap.get("performRpInitiatedLogout") instanceof Boolean) {
+            idpDefinition.setPerformRpInitiatedLogout((boolean)idpDefinitionMap.get("performRpInitiatedLogout"));
+        }
     }
 
     private static Map<String, String> parseAdditionalParameters(Map<String, Object> idpDefinitionMap) {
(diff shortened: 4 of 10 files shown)


uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/OIDCLoginIT.java
```diff
@@ -558,6 +560,27 @@ public void testResponseTypeRequired() {
...
+        identityProvider.getConfig().setPerformRpInitiatedLogout(false);
+        updateProvider();
+
+        validateSuccessfulOIDCLogin(zoneUrl, testAccounts.getUserName(), testAccounts.getPassword());
+
+        String externalOIDCProviderLoginPage = baseUrl;
+        webDriver.get(externalOIDCProviderLoginPage);
+        Assert.assertThat(webDriver.getPageSource(), containsString("Where to?"));
+    }
+
```
bruce-ricard: I don't understand which part of the tests check whether the OIDC logout is triggered or not.

It appears that you are testing that through the currentUrl and the PageSource of the webDriver?

If the integration tests aren't more precisely checking that we are indeed logged in/our of the OIDC provider, I don't really see the added value compared to the tests in [ZoneAwareWhitelistLogoutHandlerTests.java](https://github.com/cloudfoundry/uaa/pull/2590/files#diff-05ab05c97b6fc7c55ecccd6c1f17705b1a9d3e2beec5e2ff5e3c0d28974dcdac)

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/OIDCLoginIT.java
```diff
@@ -558,6 +560,27 @@ public void testResponseTypeRequired() {
         assertThat(webDriver.getCurrentUrl(), containsString("error_description=Missing%20response_type%20in%20authorization%20request"));
     }
 
+    @Test
+    public void successfulUaaLogoutTriggersExternalOIDCProviderLogout() {
+        validateSuccessfulOIDCLogin(zoneUrl, testAccounts.getUserName(), testAccounts.getPassword());
+
+        String externalOIDCProviderLoginPage = baseUrl;
+        webDriver.get(externalOIDCProviderLoginPage);
+        Assert.assertThat("URL validation failed", webDriver.getCurrentUrl(), endsWith("/login"));
```
bruce-ricard: Why did you decide to add "URL validation failed" as the response to a failed assertion here? I don't feel like that's going to help debug.

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/OIDCLoginIT.java
```diff
@@ -558,6 +560,27 @@ public void testResponseTypeRequired() {
         assertThat(webDriver.getCurrentUrl(), containsString("error_description=Missing%20response_type%20in%20authorization%20request"));
     }
 
+    @Test
+    public void successfulUaaLogoutTriggersExternalOIDCProviderLogout() {
+        validateSuccessfulOIDCLogin(zoneUrl, testAccounts.getUserName(), testAccounts.getPassword());
```
bruce-ricard, line 568: I think that I would like to see

```
identityProvider.getConfig().setPerformRpInitiatedLogout(true);
```

here, like you did in the other, opposite test.

I don't care about testing the default setting in these tests, and I think that it makes the test easier to read on its own.

server/src/test/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthLogoutHandlerTest.java
```diff
@@ -134,4 +134,15 @@ void getOAuthProviderForAuthentication() {
   void getNullOAuthProviderForAuthentication() {
     assertEquals(null, oAuthLogoutHandler.getOAuthProviderForAuthentication(null));
   }
+
+  @Test
+  void getPerformRpInitiatedLogout() {
+    oAuthIdentityProviderDefinition.setPerformRpInitiatedLogout(true);
+    assertEquals(true, oAuthLogoutHandler.getPerformRpInitiatedLogout(oAuthIdentityProviderDefinition));
```
bruce-ricard, line 141: I'm pretty sure that `assertTrue` and `assertFalse` are Junit methods that you can use to assert on booleans, instead of using `assertEquals(true/false,...)`

server/src/test/java/org/cloudfoundry/identity/uaa/authentication/ZoneAwareWhitelistLogoutHandlerTests.java
```diff
@@ -163,12 +163,21 @@ public void test_external_client_redirect() {
     }
 
     @Test
-    public void test_exteral_logout() throws ServletException, IOException {
+    public void test_external_logout() throws ServletException, IOException {
         when(oAuthLogoutHandler.getLogoutUrl(null)).thenReturn("");
+        when(oAuthLogoutHandler.getPerformRpInitiatedLogout(null)).thenReturn(true);
         handler.onLogoutSuccess(request, response, null);
         verify(oAuthLogoutHandler, times(1)).onLogoutSuccess(request, response, null);
```
bruce-ricard, line 170: I try to avoid using `verify`. Ideally we should be testing whether we are logged in the OIDC provider or not. Checking that we are calling a specific method is testing the implementation detail.

But, I think that this is one of the best possible examples of this pattern. Here, being logged out is pretty much equivalent to `oAuthLogoutHandler.onLogoutSuccess` being called. The parameters could change though...
Maybe better would be:

```
.onLogoutSuccess(any(), any(), any())
```

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/OauthIDPWrapperFactoryBean.java
```diff
@@ -166,6 +166,9 @@ protected void setCommonProperties(Map<String, Object> idpDefinitionMap, Abstrac
         if (idpDefinitionMap.get("clientAuthInBody") instanceof Boolean) {
             idpDefinition.setClientAuthInBody((boolean)idpDefinitionMap.get("clientAuthInBody"));
         }
+        if (idpDefinitionMap.get("performRpInitiatedLogout") instanceof Boolean) {
+            idpDefinition.setPerformRpInitiatedLogout((boolean)idpDefinitionMap.get("performRpInitiatedLogout"));
+        }
```
bruce-ricard, line 171: I would like to have an `else` block here, throwing a type error exception. 

Couldn't a user set the value to some silly string like `tralse` in the configuration file, and the code just wouldn't say anything, just stay on the default value?

server/src/main/java/org/cloudfoundry/identity/uaa/authentication/ZoneAwareWhitelistLogoutHandler.java
```diff
@@ -47,11 +47,12 @@ public ZoneAwareWhitelistLogoutHandler(MultitenantClientServices clientDetailsSe
     public void onLogoutSuccess(HttpServletRequest request, HttpServletResponse response, Authentication authentication) throws IOException, ServletException {
         AbstractExternalOAuthIdentityProviderDefinition oauthConfig = externalOAuthLogoutHandler.getOAuthProviderForAuthentication(authentication);
         String logoutUrl = externalOAuthLogoutHandler.getLogoutUrl(oauthConfig);
+        Boolean shouldPerformRpInitiatedLogout = externalOAuthLogoutHandler.getPerformRpInitiatedLogout(oauthConfig);
 
-        if (logoutUrl == null) {
-            getZoneHandler().onLogoutSuccess(request, response, authentication);
```
bruce-ricard, line 52: Do you happen to know what this does? I was expecting nothing to be done in this case.

Why don't we need to call this method if we do the oidc logout instead?

model/src/main/java/org/cloudfoundry/identity/uaa/provider/AbstractExternalOAuthIdentityProviderDefinition.java
```diff
@@ -214,6 +215,14 @@ public boolean isPkce() {
         return this.pkce;
     }
 
+    public boolean isPerformRpInitiatedLogout() {
```
bruce-ricard, line 218: I know that `isPerform` is a common pattern in the UAA, but we have a chance to make the world a better place here :).

Would you object to calling this method `doPerformRpInitiatedLogout` instead? Or maybe just `performsRpInitiatedLogout`. The object _isn't_ a performed RP initiated logout, it does one, or enables one.

bruce-ricard (commented):
I'm quite confident that the change works as expected.

Just added a few minor readability comments inline.

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/OIDCLoginIT.java
peterhaochen47: I'm just following the [newest existing pattern](https://github.com/cloudfoundry/uaa/blob/9a9537e439bde5c4ea071234413bb0617373416e/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/pageObjects/Page.java#L17) brought by the "page objects" refactor. I assumed there's some benefit to this pattern? At least shouldn't harm.

server/src/test/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthLogoutHandlerTest.java
peterhaochen47: Yes, also `assertNull`. Intellij highlights all these opportunities for simplification for us. My thought was that I'm gonna following exactly the existing pattern, and then issue a separate PR later for refactors (simplifying not just the tests added here but also the other tests, all simplifications in one go), so that this PR is minimal & the code pattern is consistent within the same class.

server/src/test/java/org/cloudfoundry/identity/uaa/authentication/ZoneAwareWhitelistLogoutHandlerTests.java
peterhaochen47: The actual checking of "whether we are logged in the OIDC provider or not" is done in `OIDCLoginIT.java` which actually sets up the external OIDC integration. This test is just a lower-value but simpler unit test. 

"The parameters could change though...Maybe better would be: .onLogoutSuccess(any(), any(), any())"
Maybe? I'm trying to avoid non-essential refactors in this PR. The `onLogoutSuccess(request, response, null)` assertion is existing code, I don't wanna loosen the assertion in this PR.

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/OIDCLoginIT.java
peterhaochen47:
```
        webDriver.get(externalOIDCProviderLoginPage);
        Assert.assertThat("URL validation failed", webDriver.getCurrentUrl(), endsWith("/login"));
```
^ This is going to the external OIDC provider login page (which is also a UAA), and verifies that you are redirected to the UAA `/login` endpoint. (which only happens if the user is unauthenticated/logged-out).

```
        Assert.assertThat(webDriver.getPageSource(), containsString("Where to?"));
```
This "Where to?" string only shows up on the external OIDC provider login page (which is also a UAA), if you are still logged in as an authenticated user. 

I understand that this way of checking is sort of hard to read, but it's consistent with how we perform the same type of check elsewhere. The "page object" refactor sort of addresses this readability issue by [including this check inside of page objects](https://github.com/cloudfoundry/uaa/blob/9a9537e439bde5c4ea071234413bb0617373416e/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/pageObjects/HomePage.java#L16), but I don't wanna include the page objects refactor in this PR, but I make it easy to perform that refactor in the future.

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/OIDCLoginIT.java
peterhaochen47: That's a fair point. 

My intention: this test was added in commit 1 (where no prod code changes exist), and not modified in commit 2 (where prod code change happened & the new config added) and still passing. I was trying to demonstrate that this PR is backward compatible. But I guess that has been proven now by the fact that both commits passed tests. So sure, I think it's fair to add a commit to add `identityProvider.getConfig().setPerformRpInitiatedLogout(true);` for readability.

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/OauthIDPWrapperFactoryBean.java
peterhaochen47: The lack of error exception of malformed boolean config fields is an existing problem (if it is proven to be a problem, there could be some code elsewhere doing this erroring), so I would consider it out of scope of this PR, which I'm trying to scope minimally. 

But you are right. it could be a problem. I'll consider that for the "refactor PR" later. But the solution will need to apply to all the boolean configs (like `clientAuthInBody`).

model/src/main/java/org/cloudfoundry/identity/uaa/provider/AbstractExternalOAuthIdentityProviderDefinition.java
peterhaochen47: Yeah, grammatically, `isPerformRpInitiatedLogout` does not make sense. However:
1. the other boolean configs in the same class follow this pattern:
```
isShowLinkText
isClientAuthInBody
isPkce
```
I don't wanna create an inconsistency within the same class (or I want a refactor to apply to all "getters" like this).

2. intellij's auto getter/setter generation will follow this `isSomeBoolean` pattern, which suggests that this pattern is quite standard. 

Yeah, if this function is a more custom function, I would feel more free to name it however I see fit. But since this is a standard required getter, I feel the need to follow the standard pattern / pattern in the same class for now.

server/src/main/java/org/cloudfoundry/identity/uaa/authentication/ZoneAwareWhitelistLogoutHandler.java
peterhaochen47: Good question. I just spent a good 10 mins looking at this, can't figure it out exactly...

My guess would be that in the "OIDC RP-initiated logout" case:
- the user clicks logout button on the client app
- gets redirected to UAA logout.do
- [codepath diverges here]
- gets redirected to the external OIDC logout page
- and then gets redirected back to UAA page
- Does a redirection back to the client app happen? Not sure.

And for the other case:
- the user clicks logout button on the client app
- gets redirected to UAA logout.do
- [codepath diverges here]
- and then gets redirected back to the client app according to the [post_logout_redirect_uri param](https://docs.cloudfoundry.org/api/uaa/version/76.25.0/index.html#logout-do)

So I'm not too surprised that the codepath diverges completely here for these 2 cases.

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/OIDCLoginIT.java
peterhaochen47: Commit pushed.

model/src/main/java/org/cloudfoundry/identity/uaa/provider/AbstractExternalOAuthIdentityProviderDefinition.java
bruce-ricard: OK I'll add a topic to our next tech forum about the general idea of "creating inconsistencies by doing the right thing".

It's not like it's very important anyways :).

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthLogoutHandler.java
```diff
@@ -123,4 +123,11 @@ private String getZoneDefaultUrl() {
     }
     return config.getLinks().getLogout().getRedirectUrl();
   }
+
+  public Boolean getPerformRpInitiatedLogout(AbstractExternalOAuthIdentityProviderDefinition oauthConfig) {
```
bruce-ricard: It looks like this `Boolean` could actually be a `boolean`. That would make sure that it's never `null`.

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/OIDCLoginIT.java
bruce-ricard: Hmm I don't like the word "validation" in this context.

> shouldn't harm

I feel like the word "validation" shouts "I'm not test code". In test code we typically "assert", "test", "check". Validation is IMO when typically when real production code needs to validate something before doing something else with it.
It "harmed" my understanding when I read the code at first. I was thinking that it was a bad copy and paste from some functional code. I didn't think we could validate in tests.

@swalchemist what do you think?

server/src/main/java/org/cloudfoundry/identity/uaa/authentication/ZoneAwareWhitelistLogoutHandler.java
bruce-ricard: Alright. We are not changing that logic so it doesn't matter much.

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthLogoutHandler.java
peterhaochen47: You are right! commit pushed.

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/OIDCLoginIT.java
swalchemist: I think I should improve that message. We should only add these assertion messages if there's no better way to improve our understanding of what the test is doing and/or what happens when it fails.

In this case, I think "We're at the right URL" works better as the message, e.g., "AssertThat ... we're at the right URL."

peterhaochen47: Ok, how about for the purpose of this PR, I'll forget about refactoring to using the shared "validateUrl" function in the future, and just formulate a `reason` param that works for the test here specifically (how about `reason` = "Not redirected to external OIDC provider login page." as that is what I'm actually testing here), or just taking the `reason` param out for now.

peterhaochen47: Commit pushed.

strehle:
Review.
LGTM

2 remarks from sonar 
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2590

1 smell about raw type usage https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&types=CODE_SMELL&pullRequest=2590&id=cloudfoundry-identity-parent&open=AYunIauC0iXdSBD0Oshu
1 additional test might increase coverage on zone logout, e.g. https://sonarcloud.io/component_measures?id=cloudfoundry-identity-parent&pullRequest=2590&metric=new_coverage&view=list

strehle (approved):
see my comment

bruce-ricard (approved):

peterhaochen47:
> Review. LGTM
> 
> 2 remarks from sonar https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2590
> 
> 1 smell about raw type usage https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&types=CODE_SMELL&pullRequest=2590&id=cloudfoundry-identity-parent&open=AYunIauC0iXdSBD0Oshu 1 additional test might increase coverage on zone logout, e.g. https://sonarcloud.io/component_measures?id=cloudfoundry-identity-parent&pullRequest=2590&metric=new_coverage&view=list

Thanks! I've addressed all the ones I can.

PR #2592, opened 2023-11-07
fix sonar finding

strehle:
first only the found smell
https://sonarcloud.io/project/issues?impactSeverities=HIGH&resolved=false&sinceLeakPeriod=true&types=CODE_SMELL&id=cloudfoundry-identity-parent

later also implemented recommendations from sonar linter


strehle:
re-scan

https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2592

server/src/main/java/org/cloudfoundry/identity/uaa/zone/MultitenantJdbcClientDetailsService.java
```diff
@@ -147,33 +148,30 @@ public void addClientDetails(ClientDetails clientDetails, String zoneId) throws
 
     private boolean exists(String clientId, String zoneId) {
         List<String> idResults = jdbcTemplate.queryForList(SINGLE_SELECT_STATEMENT, String.class, clientId, zoneId);
-        if (idResults != null && idResults.size() == 1) {
-            return true;
-        }
-        return false;
+        return idResults.size() == 1;
```
adrianhoelzl-sap: Don't we need the null check still? Or is it guaranteed that `idResults` will never be null?

strehle: see here, I reverted, but Sonar see the null check as smell

https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&sinceLeakPeriod=true&pullRequest=2592&id=cloudfoundry-identity-parent

Smell is minor, so we can let it, or remove it, you can decide

adrianhoelzl-sap (approved):

PR #2606, opened 2023-11-17
Add allowedGroups to UserConfig in IdZ configuration

### LINKED ISSUE #2505: Missing user management from custom IdZ to CC

## Context

## Proposal

image

klaus-sap:
https://github.com/cloudfoundry/uaa/issues/2505 -> Proposal 3.
We want to add an optional attribute "allowedGroups" to the UserConfig in IdZ configuration. If this attribute is not set (i.e., allowedGroups == null), then all groups are allowed as before. If this attribute contains an array of groups, then only the groups from defaultGroups plus allowedGroups are allowed. All other groups can neither be created nor assigned to users.


strehle:
@klaus-sap here you can see the smells from sonar
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2606

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupProvisioning.java
```diff
@@ -222,8 +227,29 @@ public ScimGroup retrieve(String id, final String zoneId) throws ScimResourceNot
...
     }
 
+    private List<String> getAllowedUserGroups(String zoneId) {
+        List<String> zoneAllowedGroups = null; // default: all groups allowed
+        if (!hasText(zoneId)) {
+            return zoneAllowedGroups;
+        }
+        IdentityZone currentZone = IdentityZoneHolder.get();
+        try {
+            zoneAllowedGroups = (zoneId.equals(currentZone.getId())) ?
```
strehle: You should add this only if currentZone,.getId != "uaa" , because the default zone we dont want /need this check here?

model/src/main/java/org/cloudfoundry/identity/uaa/zone/UserConfig.java
```diff
@@ -27,4 +27,14 @@ public List<String> getDefaultGroups() {
     public void setDefaultGroups(List<String> defaultGroups) {
         this.defaultGroups = defaultGroups;
     }
+
+    private List<String> allowedGroups = null;
+
```
strehle, line 35: Combine the default and the allowed Groups
e.g. if you have entries in defaultGroups, they are automatically in allowed as far as I understood?

strehle (commented):
This should work in all zones != uaa

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupProvisioning.java
```diff
@@ -74,6 +78,7 @@ public JdbcScimGroupProvisioning(
         super(jdbcTemplate, pagingListFactory, new ScimGroupRowMapper());
 
         this.jdbcTemplate = jdbcTemplate;
+        jdbcIdentityZoneProvisioning = new JdbcIdentityZoneProvisioning(jdbcTemplate); // TODO inject
```
strehle: please enhance
https://github.com/cloudfoundry/uaa/blob/develop/uaa/src/main/webapp/WEB-INF/spring/scim-endpoints.xml#L112
with jdbcIdentityZoneProvisioning

klaus-sap: no bean jdbcIdentityZoneProvisioning found for ref

strehle (dismissed):
Open
a) combine default and allowed groups. So from understanding, if you have default groups defined, then these entries are automatically part of allowedGroups. So this behaviour should be implemented.
b) Check if the limitation should be really done always or only in Identity Zone NOT EQUAL to 'uaa'

klaus-sap:
> Open a) combine default and allowed groups. So from understanding, if you have default groups defined, then these entries are automatically part of allowedGroups. So this behaviour should be implemented. b) Check if the limitation should be really done always or only in Identity Zone NOT EQUAL to 'uaa'

a) commit [6e6ab18](https://github.com/cloudfoundry/uaa/pull/2606/commits/6e6ab189b2be5f60f2b3200ede998a04c5253aa8)
b) I would prefer to handle the IdZ configuration of all zones in the same way. If you don't want allowedGroups in the "uaa" zone, simply don't configure them

model/src/main/java/org/cloudfoundry/identity/uaa/zone/UserConfig.java
klaus-sap: done

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupProvisioning.java
klaus-sap: I would prefer to handle the configuration of all zones in the same way

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaAuthorizationRequestManager.java
```diff
@@ -249,7 +249,11 @@ private Set<String> checkUserScopes(Set<String> requestedScopes,
         Set<String> allowed = new LinkedHashSet<>(AuthorityUtils.authorityListToSet(authorities));
         // Add in all default requestedScopes
         Collection<String> defaultScopes = IdentityZoneHolder.get().getConfig().getUserConfig().getDefaultGroups();
+        Collection<String> allowedScopes = IdentityZoneHolder.get().getConfig().getUserConfig().resultingAllowedGroups();
         allowed.addAll(defaultScopes);
+        if (allowedScopes != null) {
+            allowed.retainAll(allowedScopes);
```
strehle, line 256: Test missing? or did not found one
Situation. Your authorities contain 3 entries but in your zone you allow only 1 of them

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupProvisioning.java
```diff
@@ -222,10 +231,32 @@ public ScimGroup retrieve(String id, final String zoneId) throws ScimResourceNot
         }
     }
 
+    @SuppressWarnings("java:S1874")
+    private Set<String> getAllowedUserGroups(String zoneId) {
+        Set<String> zoneAllowedGroups = null; // default: all groups allowed
+        if (!hasText(zoneId)) {
+            return zoneAllowedGroups;
```
strehle: No Test found for this

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupProvisioning.java
```diff
@@ -222,10 +231,32 @@ public ScimGroup retrieve(String id, final String zoneId) throws ScimResourceNot
...
     @Override
     public ScimGroup create(final ScimGroup group, final String zoneId) throws InvalidScimResourceException {
+        Set<String> allowedGroups = getAllowedUserGroups(zoneId);
+        if ((allowedGroups != null) && (!allowedGroups.contains(group.getDisplayName()))) {
+            throw new InvalidScimResourceException("The group with displayName: " + group.getDisplayName()
+                    + " is not allowed in Identity Zone " + zoneId);
+        }
         final String id = UUID.randomUUID().toString();
-        logger.debug("creating new group with id: " + id);
+        logger.debug("creating new group with id: %s", id);
```
strehle: here you can do {} instead of %s

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupProvisioning.java
```diff
@@ -248,6 +279,11 @@ public ScimGroup create(final ScimGroup group, final String zoneId) throws Inval
     @Override
     public ScimGroup update(final String id, final ScimGroup group, final String zoneId) throws InvalidScimResourceException,
             ScimResourceNotFoundException {
+        Set<String> allowedGroups = getAllowedUserGroups(zoneId);
+        if ((allowedGroups != null) && (!allowedGroups.contains(group.getDisplayName()))) {
+            throw new InvalidScimResourceException("The group with displayName: " + group.getDisplayName()
```
strehle: No test found with this error

model/src/main/java/org/cloudfoundry/identity/uaa/zone/UserConfig.java
```diff
@@ -27,4 +29,26 @@ public List<String> getDefaultGroups() {
...
+        return allowedGroups;
+    }
+
+    public void setAllowedGroups(List<String> allowedGroups) {
+        this.allowedGroups = allowedGroups;
+    }
+
+    // return defaultGroups plus allowedGroups
+    public Set<String> resultingAllowedGroups() {
+        if (allowedGroups == null) {
```
strehle: Test coverage in this area should be increased

strehle (commented):
see here new sonar

https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2606

coverage
https://sonarcloud.io/component_measures?id=cloudfoundry-identity-parent&pullRequest=2606&metric=new_coverage&view=list

smells
not all need to be fixed, but please check them
https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&sinceLeakPeriod=true&pullRequest=2606&id=cloudfoundry-identity-parent

strehle:
@klaus-sap Please add some scenario tests into
https://github.com/cloudfoundry/uaa/blob/develop/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/ScimGroupEndpointsIntegrationTests.java

you can create an extra zone or stay in default , btw. zone creation https://github.com/cloudfoundry/uaa/blob/develop/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/util/IntegrationTestUtils.java#L671

Scenario I see
a) Add allowedGroups to zone and perform positive and negative test , means create the group 
b) Update or deleted one entry in defaultGroups (because this has influence to allowedGroups) and then do the same create the groups and verify result

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupProvisioning.java
klaus-sap: done

model/src/main/java/org/cloudfoundry/identity/uaa/zone/UserConfig.java
klaus-sap: added unit test UserConfigTest.java with commit [8cde93e](https://github.com/cloudfoundry/uaa/pull/2606/commits/8cde93e0abb522f5a23552c547eb9d7228715a85)

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaAuthorizationRequestManager.java
klaus-sap: added testScopesIncludesAllowedAuthoritiesForUser() [a464bab](https://github.com/cloudfoundry/uaa/pull/2606/commits/a464babd7399d39fa3eaa6778b9b12dec00ecee7)

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupProvisioning.java
klaus-sap: zoneId must not be null, changed logic: [370be52](https://github.com/cloudfoundry/uaa/pull/2606/commits/370be522393608696adab86ba3888269b8ed235d)

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupProvisioning.java
klaus-sap: check error message in cannotUpdateNotAllowedGroup: [370be52](https://github.com/cloudfoundry/uaa/pull/2606/commits/370be522393608696adab86ba3888269b8ed235d)

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupProvisioning.java
```diff
@@ -222,10 +231,32 @@ public ScimGroup retrieve(String id, final String zoneId) throws ScimResourceNot
         }
     }
 
+    @SuppressWarnings("java:S1874")
+    private Set<String> getAllowedUserGroups(String zoneId) {
+        if (zoneId == null) {
+            throw new IllegalArgumentException("zoneId is null");
```
strehle: this here is not tested, but if you change 
 (zoneId.equals(currentZone.getId())) 
to
 (currentZone.getId(equals(zoneId))

you dont need the null check for zoneId

klaus-sap: done with [ab2d5cf](https://github.com/cloudfoundry/uaa/pull/2606/commits/ab2d5cf74037ce00effc502395c8b2578be87c2e)
(line 241) "jdbcIdentityZoneProvisioning.retrieve(zoneId)" seems to be null-safe as well

strehle (approved):
### sonar ok,
- almost 100 coverage.
- Integration Tests available

https://sonarcloud.io/component_measures?id=cloudfoundry-identity-parent&pullRequest=2606&metric=new_coverage&view=list

@hsinn0 @Tallicia FYI please check who can do a review from VMware side in next week , thanks

tack-sap (approved):

server/src/main/java/org/cloudfoundry/identity/uaa/zone/UserConfigValidator.java
```diff
@@ -0,0 +1,23 @@
...
+
+    public static void validate(UserConfig config) throws InvalidIdentityZoneConfigurationException {
+        Set<String> allowedGroups = (config == null) ? null : config.resultingAllowedGroups();
+        if ((allowedGroups != null) && (allowedGroups.isEmpty())) {
+            String message = "At least one group must be allowed";
+            logger.error(message);
+            throw new InvalidIdentityZoneConfigurationException(message);
+        }
+    }
+}
```
bruce-ricard: It feels like this class doesn't need to exist, and the method could be part of `UserConfig`. I like to avoid util classes whenever I can.

server/src/main/java/org/cloudfoundry/identity/uaa/zone/UserConfigValidator.java
```diff
@@ -0,0 +1,23 @@
...
+
+    // add a private constructor to hide the implicit public one
+    private UserConfigValidator() {
+    }
+
+    public static void validate(UserConfig config) throws InvalidIdentityZoneConfigurationException {
+        Set<String> allowedGroups = (config == null) ? null : config.resultingAllowedGroups();
+        if ((allowedGroups != null) && (allowedGroups.isEmpty())) {
+            String message = "At least one group must be allowed";
+            logger.error(message);
```
bruce-ricard: This is not an error. This is info. Error means the UAA is having an error. Here it's behaving as expected.

server/src/main/java/org/cloudfoundry/identity/uaa/zone/UserConfigValidator.java
```diff
@@ -0,0 +1,23 @@
...
+            throw new InvalidIdentityZoneConfigurationException(message);
```
bruce-ricard, line 15: I'm surprised that Sonar isn't flagging this. Usually it says that you should either log the error, or throw the exception.

Now that I think about it, I think that Sonar only catches it when it's about catching and rethrowing an exception.

But I still think that it applies here. If feels like your message is going to be logged 2 times. Once here, and wherever the exception caught.

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupProvisioning.java
```diff
@@ -324,4 +349,12 @@ protected void validateOrderBy(String orderBy) throws IllegalArgumentException {
         super.validateOrderBy(orderBy, GROUP_FIELDS);
     }
 
+    private void validateAllowedUserGroups(String zoneId, ScimGroup group) {
```
bruce-ricard, line 352: You are only validating the allowed groups in `create` and `update`. It seems to me like it would be possible for a group to have been created in the past with certain user groups, then the admin sets the userconfig to specify certain allowed user groups, and this object still contains the old user groups that haven't been validated. And will never be validated until the object is updated, which could be never.

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupProvisioning.java
```diff
@@ -220,9 +229,24 @@ public ScimGroup retrieve(String id, final String zoneId) throws ScimResourceNot
         }
     }
 
+    @SuppressWarnings("java:S1874")
```
bruce-ricard, line 232: I see lots of these in this PR. That feels like a code smell. Why are we suppressing all these warning?

Also, I don't know what warning this is suppressing. I tried googling this number, I can't find anything. It's not great for being able to understand the code.

model/src/main/java/org/cloudfoundry/identity/uaa/zone/UserConfig.java
```diff
@@ -27,4 +29,27 @@ public List<String> getDefaultGroups() {
...
+
+    public void setAllowedGroups(List<String> allowedGroups) {
+        this.allowedGroups = allowedGroups;
+    }
+
+    // return defaultGroups plus allowedGroups
+    @SuppressWarnings("java:S1168")
+    public Set<String> resultingAllowedGroups() {
+        if (allowedGroups == null) {
+            return null; // null = all groups allowed
```
bruce-ricard, line 51: I like to try to avoid `null`s whenever I can. I think that here we can. At the very least, it's quite confusing to the reader that `null` means when they are calling this method and it returns `null`, assuming that they remember to do the null check.

I have 2 different suggestions:
1. Use `Optional`. The type of this field would be `Optional<Set<String>>`. It solves the null problem, but it leaves open the interpretation of what `empty` is.
2. Have another method `public bool areAllUserGroupsAllowed` to catch that case. If that returns false, you can call the `resultingAllowedGroups` to see which groups are allowed. I understand that it is a bit heavy, but it has the advantage of being very clear and readable.

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupProvisioning.java
```diff
@@ -220,9 +229,24 @@ public ScimGroup retrieve(String id, final String zoneId) throws ScimResourceNot
...
+    private Set<String> getAllowedUserGroups(String zoneId) {
+        Set<String> zoneAllowedGroups = null; // default: all groups allowed
+        try {
+            IdentityZone currentZone = IdentityZoneHolder.get();
+            zoneAllowedGroups = (currentZone.getId().equals(zoneId)) ?
+                currentZone.getConfig().getUserConfig().resultingAllowedGroups() :
+                jdbcIdentityZoneProvisioning.retrieve(zoneId).getConfig().getUserConfig().resultingAllowedGroups();
```
bruce-ricard, line 239: Why do you have the conditional here? Wouldn't it be simpler to just always do the `retrieve(zoneId)`, instead of checking if we are in the current zone?

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupProvisioning.java
```diff
@@ -324,4 +349,12 @@ protected void validateOrderBy(String orderBy) throws IllegalArgumentException {
         super.validateOrderBy(orderBy, GROUP_FIELDS);
     }
 
+    private void validateAllowedUserGroups(String zoneId, ScimGroup group) {
+        Set<String> allowedGroups = getAllowedUserGroups(zoneId);
+        if ((allowedGroups != null) && (!allowedGroups.contains(group.getDisplayName()))) {
```
bruce-ricard, line 354: I don't like that we are doing the filtering based on the group display name. That feels quite dangerous.

Is it because in this specific case the "display name" is actually the unique string that represents it as well? If that's the case, could we then rename that method to be `group.identifier()` instead of `getDisplayName()`?

bruce-ricard (dismissed):
I have one major concern about the validation only happening at group write time, see inline comments.

And a couple sub-major implementation design comments.

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupProvisioning.java
strehle: Hi @bruce-ricard ok, so vote to stay here with the logic . We may can reactor readability, but the current zone is in the local thread already, no DB lookup needed, but other zones need to be retrieved from DB. if we simply retrieve from DB, then we have always a DB lookup

model/src/main/java/org/cloudfoundry/identity/uaa/zone/UserConfig.java
strehle: I vote for 2, which might simplify it

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupProvisioning.java
strehle: Ok, but this is a old discussion old. Must of the S1874 warnings from from deprecated code which we cause, e.g. IdentityZoneHoler or https://github.com/cloudfoundry/uaa/milestone/38 but this annotation simply cleanup  
our own deprecations.....  why is IdentityZoneHoler deprecated but nobody cares about it or removes it

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupProvisioning.java
strehle: ok, but the groups in UAA what in reality is used. UAA uses the SCIM group for authorizations and the display name is the scope name. The identitfier is a not used entity , also in other usage of it

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupProvisioning.java
strehle: you want the validation in token creation ?

server/src/main/java/org/cloudfoundry/identity/uaa/zone/UserConfigValidator.java
strehle: we will have more to validate later, so this validator classes are a pattern in other parts of UAA

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupProvisioning.java
bruce-ricard: OK so it's a performance optimization. It's what I was thinking.

I like to try to make code self explanatory, whenever possible. For example here, I don't want future readers of the code to ask themselves this question I asked, until the end of times. 
Typically, I would have created a new method for this operation, and tried to pass the performance optimization explanation to the reader through good variable and method names.

Another different argument would be to say: this method is only called at group update and creation time. This is quite rare. I don't think that we should care about a single DB lookup made at group write time.

Something else: we do the test `currentZone.getId().equals(zoneId)`, and on the next line we get the current zone config. Isn't it possible that between these 2 instructions, another thread changes the current zone, and we end up looking at allowed groups from another zone?

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupProvisioning.java
bruce-ricard: IMO, we should suppress the warning when we believe the warning is incorrect. We shouldn't suppress the warnings because we don't like the noise, and because we don't want to fix the underlying issue. 

Here, a real issue is reported by Sonar. We shouldn't hide it.

I am not saying that we need to fix the deprecation problem right now. But IMO we need to remove the tags, and allow Sonar to show that there is a problem.

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupProvisioning.java
bruce-ricard: I am not sure if that makes sense. That would require the validation to be done at every single token creation time, that feels wrong from the performance point of view. But it seems that you are correct, that it would solve this issue.

I was thinking of another idea: whenever the user config is updated, we need to revalidate all groups. That would make the config update more computationally heavy, but that is a lot less frequent than token creation.

server/src/main/java/org/cloudfoundry/identity/uaa/zone/UserConfigValidator.java
bruce-ricard: > validator classes are a pattern in other parts of UAA

I know they are a commonly used pattern in UAA. I think that this is a bad pattern, I'm trying to move away from it.

IMHO it would be much better for this static `validate()` method to be moved as an instance method of the `UserConfig` class.

But I can understand if you really want to have the validation code in another class. If that's the case, let me suggest this pattern:
In UserConfig, you have:

```
public static void validate(UserConfigValidator validator) {
    validator.validate(this)
}
```
And in the `UserConfigValidator`, the method becomes an instance one. The class is no longer a "util" class with static methods. 

I think that this is a common OO design pattern, I just don't know its name.

klaus-sap: I would prefer to stay with a separate class UserConfig, because we now have a matching test class UserConfigTest. There are no member variables in UserConfig, so I would like to leave it static

server/src/main/java/org/cloudfoundry/identity/uaa/zone/UserConfigValidator.java
klaus-sap: changed log level with cc27f6661a541d1462b07af49c245803869d4bac

server/src/main/java/org/cloudfoundry/identity/uaa/zone/UserConfigValidator.java
klaus-sap: changed with ef94dea12c7af9305d06317acb0daf34fd5c7c6f

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupProvisioning.java
klaus-sap: Ok, this would be a change in "Update Identity Zone", right? We might add a check if there are groups which would be not allowed anymore after updating the IdZ. In this case, we would raise an UnprocessableEntityException

model/src/main/java/org/cloudfoundry/identity/uaa/zone/UserConfig.java
klaus-sap: ok, but what should resultingAllowedGroups() return if areAllUserGroupsAllowed == true? An empty optional?

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupProvisioning.java
klaus-sap: Regarding the last paragraph: the same may happen if you use an if-statement instead of a conditional. And the time elapsed between the two instructions is really tiny :-)

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupProvisioning.java
klaus-sap: Added check: [56adc44](https://github.com/cloudfoundry/uaa/pull/2606/commits/56adc44dbf4b0281bf8044e852fce5ce3d60253a)

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimGroupProvisioning.java
strehle: removed

hsinn0 (approved):

hsinn0:
@bruce-ricard & I agreed that I would take over this PR review from our side as he will be unavailable for a couple of weeks. I am approving this PR.

PR #2608, opened 2023-11-17
Miscellaneous refactors

peterhaochen47:
- particularly codes related to https://github.com/cloudfoundry/uaa/pull/2590

### DIFF
--- a/model/src/main/java/org/cloudfoundry/identity/uaa/provider/AbstractExternalOAuthIdentityProviderDefinition.java
+++ b/model/src/main/java/org/cloudfoundry/identity/uaa/provider/AbstractExternalOAuthIdentityProviderDefinition.java
@@ -16,12 +16,14 @@ package org.cloudfoundry.identity.uaa.provider;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
+import lombok.Getter;
 
 import java.lang.reflect.ParameterizedType;
 import java.net.URL;
 import java.util.List;
 import java.util.Objects;
 
+@Getter
 @JsonIgnoreProperties(ignoreUnknown = true)
 public abstract class AbstractExternalOAuthIdentityProviderDefinition<T extends AbstractExternalOAuthIdentityProviderDefinition> extends ExternalIdentityProviderDefinition {
     public enum OAuthGroupMappingMode {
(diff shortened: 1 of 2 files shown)


model/src/main/java/org/cloudfoundry/identity/uaa/provider/AbstractExternalOAuthIdentityProviderDefinition.java
```diff
@@ -16,12 +16,14 @@
...
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
+import lombok.Getter;
 
 import java.lang.reflect.ParameterizedType;
 import java.net.URL;
 import java.util.List;
 import java.util.Objects;
 
+@Getter
```
swalchemist, line 26: Nice! I'm a big fan of Lombok.

swalchemist (approved):
Great improvements.

model/src/main/java/org/cloudfoundry/identity/uaa/provider/AbstractExternalOAuthIdentityProviderDefinition.java
hsinn0, line 26: I am not really a fan of converting existing class to use Lombok. To me, it is most useful for new code. Regardless, if you wanted to change it to use Lombok, why stop at Getter? Why not also use @Setter?

peterhaochen47: For the Setters, our existing implementations of the setters in this class is not "standard", e.g.:
```
    public T setTokenKeyUrl(URL tokenKeyUrl) {
        this.tokenKeyUrl = tokenKeyUrl;
        return (T) this;
    }
```
So I can't use the lombok setter without more code changes. It can be done though, just leaving it aside for now.

As of the motivation of this refactor - it's mainly sort of a response to [this prior discussion with Bruce](https://github.com/cloudfoundry/uaa/pull/2590#discussion_r1385608347). With this PR, I'm making the point that the "isSomeBoolean" getter pattern is now a common pattern, since it's what lombok would generate, though it might not be necessarily grammatically correct.

hsinn0: I see.

hsinn0 (approved):

strehle:
sonar happy, I am happy

https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2608

strehle (approved):

PR #2613, opened 2023-11-23
Update documentation for adding SAP IAS as OIDC Provider

vlast3k:


docs/OIDC-Provider-Examples/sap-public-oidc-provider.md
```diff
@@ -4,33 +4,54 @@ SAP IAS can be setup as an [OIDC provider](https://help.sap.com/viewer/6d6d63354
...
+8. (Optional) Use e-mail for Login Id instead of P-user
+   1. In IAS Admin Page, under "Trust / Single Sign-on / Subject Name Identifier / Basic Configuration"
+      * Select "Select a basic attribute" : "Email"
+   2. In `uaa.yml` append the following configuration, to the `login.oauth.providers.{origin}` section (at the same level as the other properties from the example above):
+      ```
+      attributeMappings:
+          given_name: "first_name"
+          family_name: "last_name"
+          email: "mail"
+          user_name: "mail"
```
strehle: user_name: email

docs/OIDC-Provider-Examples/sap-public-oidc-provider.md
```diff
@@ -4,33 +4,54 @@ SAP IAS can be setup as an [OIDC provider](https://help.sap.com/viewer/6d6d63354
...
+9. (optional) For CF Login, use `cf login --sso` and select the provider.
+    * Trying to login with User/Pass requires a confidential OAuth Client, creating a Secret in the Client Authentication tab, adding it as `relyingPartySecret` property and disabling "Enforce PKCE"
+  
+8. (Optional) Use e-mail for Login Id instead of P-user
+   1. In IAS Admin Page, under "Trust / Single Sign-on / Subject Name Identifier / Basic Configuration"
+      * Select "Select a basic attribute" : "Email"
+   2. In `uaa.yml` append the following configuration, to the `login.oauth.providers.{origin}` section (at the same level as the other properties from the example above):
+      ```
+      attributeMappings:
+          given_name: "first_name"
```
strehle: please remove the 3 lines for
given_name
family_name
email

because UAA has requested email and profile scope and with that the SAP OIDC provider will return the standard OIDC claims. These Claims are used

In generel if you set 8 the email to outgoing login id then you dont need 
user_name: email

because then in sub the email is there

docs/OIDC-Provider-Examples/sap-public-oidc-provider.md
```diff
@@ -4,33 +4,54 @@ SAP IAS can be setup as an [OIDC provider](https://help.sap.com/viewer/6d6d63354
...
 
-4. Ensure that the scope `openid`, `email` and `profile` is included in the`scopes` property. Then UAA shadow user (if addShadowUserOnLogin=true) is created
-   with all properties. 
+7. Restart UAA.
+   * You may see `Login with IAS-Public` link on your login page.
+   * Or if the link is not displayed, you need to enter the `{origin}` manually and then login against it
 
-5. Restart UAA. You will see `Login with IAS-Public` link on your login page.
+9. (optional) For CF Login, use `cf login --sso` and select the provider.
+    * Trying to login with User/Pass requires a confidential OAuth Client, creating a Secret in the Client Authentication tab, adding it as `relyingPartySecret` property and disabling "Enforce PKCE"
```
strehle, line 45: UAA by default uses now always PKCE

docs/OIDC-Provider-Examples/sap-public-oidc-provider.md
```diff
@@ -4,33 +4,54 @@ SAP IAS can be setup as an [OIDC provider](https://help.sap.com/viewer/6d6d63354
...
                   - profile
                 linkText: Login with IAS-Public
                 showLinkText: true
-                relyingPartyId: 3feb7ecb-d106-4432-b335-aca2689ad123
+                relyingPartyId: <client_id>
+                addShadowUserOnLogin: true
+
+6. Ensure that the scope `openid`, `email` and `profile` is included in the `scopes` property. Then UAA shadow user (if addShadowUserOnLogin=true) is created with all properties. 
 
-4. Ensure that the scope `openid`, `email` and `profile` is included in the`scopes` property. Then UAA shadow user (if addShadowUserOnLogin=true) is created
```
strehle, line 33: addShadowUserOnLogin: true 
is only needed if you have no user in UAA

So you can also use SAP IPS to provision the user into UAA

remark: the addShadowUserOnLogin is a security setting. With true you allow ALL users from a provider and no pre-provision needed (easy part) , but with false you allow only the pre-provisioned users (which is the more secure way)

strehle (commented):
In general thanks for your improvments.

Please see my comments,

docs/OIDC-Provider-Examples/sap-public-oidc-provider.md
vlast3k: i think for an initial configuration it is better to have the users to be created by default, as  having to do this manually or via SAP IPS is more complicated

docs/OIDC-Provider-Examples/sap-public-oidc-provider.md
vlast3k: @strehle , then in point 1., do we need to explicitly state that the new application should be made "public", because this confused me a bit, since there was no place to mark "public".
Maybe we could say there, that the application is by default "public", and a confidential one is required, a secret should be added? Or how the application can be made not public ?

bruce-ricard (commented):
LGTM. I didn't try to follow the steps to see if the documentation is indeed correct. Markus, I trust you know better than me and can approve this PR when you think it's ready.

strehle (approved):

PR #2617, opened 2023-11-23
Check zone id

strehle:
After discussions in https://github.com/cloudfoundry/uaa/pull/2606

Add null checks to improve robustness in this dao

add more tests to cover all edge cases

https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2617


klaus-sap (approved):
looks good to me

strehle:
@swalchemist This is a small refactoring in preparation of a pr from @klaus-sap .

Return now an exception before asking JDBC (or DB) about null for a zone.
EmptyResultDataAccessException("Zone id cannot be null", 1);

tack-sap:
@strehle currently the methods do not throw EmptyResultDataAccessException, but only ZoneDoesNotExistException. Would this maybe be a better Exception to throw, as this is something the calling classes may already expect?

strehle:
> @strehle currently the methods do not throw EmptyResultDataAccessException, but only ZoneDoesNotExistException. Would this maybe be a better Exception to throw, as this is something the calling classes may already expect?

ok

tack-sap (approved):
LGTM

PR #2618, opened 2023-11-24
Limit user creation in zone

### LINKED ISSUE #2581: Limit user creation within a zone
This is a request for a feature to limit the amount of users which can be created within the identity zone.

This is needed for https://github.com/cloudfoundry/uaa/issues/2505 but can used independently.

Scenario.
A foundational operator creates a zone for a department and provides this department access to maintain their zone themselves, e.g. scim.write . The department does not have access to maintain details or policies of the zone. This action is
only possible from uaa.admin. The depart does not have uaa.admin or zones.<idz.>.admin.

The department is trusted, however it must not exceed all resources in persistence by accident, therefore there must be a mechanism to limit scim users. The admin of the department may get zones.<zone id>.read to read own data but no more rights.

Proposal.
Identity-Zone-Configuration has some policy sections, e.g. tokenPolicy, . Here a limitPolicy should be added with such rules.
E.g. maxUsers or maxGroups

strehle:
Addition limit to https://github.com/cloudfoundry/uaa/pull/2606 

Sonar
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2618


model/src/main/java/org/cloudfoundry/identity/uaa/zone/UserConfig.java
```diff
@@ -20,11 +20,21 @@ public class UserConfig {
 
     private List<String> defaultGroups = DEFAULT_ZONE_GROUPS;
 
+    private Integer maxUsers;
+
```
hsinn0, line 26: I do not like the `maxUsers` default value being null. What is the real default number that you want to use? `-1`? In any case, I suggest that we make `maxUsers` type to `int` with some default number.

hsinn0 (commented):
What is the reason for this PR? I do not see any related issue or reasoning for the change.

strehle:
> What is the reason for this PR? I do not see any related issue or reasoning for the change.

we need this (similar to group limitation) for https://github.com/cloudfoundry/uaa/issues/2505 . 

A custom zone should have limits in future in order to run multiple custom zones in one deployment. Wihout limits, one zone could consume all possible resources

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
```diff
@@ -513,6 +527,14 @@ public int getTotalCount() {
         return count;
     }
 
+    public int getZoneCount(String zoneId, int maxLimit) {
```
adrianhoelzl-sap: Can we rename this method to include the word "Users", e.g., `getUsersCountForZone`?

server/src/main/java/org/cloudfoundry/identity/uaa/zone/UserConfigValidator.java
```diff
@@ -9,10 +9,19 @@ private UserConfigValidator() {
...
+        }
+
+        Set<String> allowedGroups = config.resultingAllowedGroups();
+        if (allowedGroups != null && allowedGroups.isEmpty()) {
             String message = "At least one group must be allowed";
             throw new InvalidIdentityZoneConfigurationException(message);
         }
+
+        int maxUsers = config.getMaxUsers();
+        if (maxUsers < -1 || maxUsers == 0 || maxUsers > Integer.MAX_VALUE) {
```
adrianhoelzl-sap: Should we introduce a global limit that is smaller than `Integer.MAX_VALUE`?

server/src/test/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioningTests.java
```diff
@@ -1042,6 +1045,31 @@ void updateLastLogonTime() {
...
+        IdentityZoneHolder.get().getConfig().getUserConfig().setMaxUsers(10);
+        ScimUser scimUser = new ScimUser("user-id-1", "user1@example.com", "User", "Example");
+        ScimUser.Email email = new ScimUser.Email();
+        email.setValue("user@example.com");
+        scimUser.setEmails(Collections.singletonList(email));
+        scimUser.setPassword("password");
+        scimUser.setOrigin(OriginKeys.UAA);
+        for (int i = 2; i < 12; i++) {
+            scimUser = jdbcScimUserProvisioning.create(scimUser, IdentityZoneHolder.getCurrentZoneId());
+            scimUser.setId("user-id-" + 1);
```
adrianhoelzl-sap:
```suggestion
            scimUser.setId("user-id-" + i);
```

server/src/test/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioningTests.java
```diff
@@ -1042,6 +1045,31 @@ void updateLastLogonTime() {
...
+            scimUser.setPassword("password");
+        }
+        scimUser.setPassword("password");
+        ScimUser finalScimUser = scimUser;
+        assertThrowsWithMessageThat(
+            InvalidScimResourceException.class,
+            () -> jdbcScimUserProvisioning.create(finalScimUser, IdentityZoneHolder.getCurrentZoneId()),
+            containsString("The maximum allowed numbers of users: 10 is reached already in Identity Zone")
+        );
+        IdentityZoneHolder.get().getConfig().getUserConfig().setMaxUsers(-1);
```
adrianhoelzl-sap: [nitpick] Should we move this to an `AfterEach` handler?

adrianhoelzl-sap: Otherwise, the value would not be reset if an exception is thrown in one of the previous statements

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
```diff
@@ -522,4 +544,26 @@ protected void validateOrderBy(String orderBy) throws IllegalArgumentException {
...
+            return userConfig.getMaxUsers();
+        } catch (ZoneDoesNotExistsException e) {
+            logger.debug("could not retrieve identity zone with id: {}", zoneId);
+        }
+        return -1;
+    }
+
+    private void validateUserLimit(String zoneId) {
+        int maxAllowedUsers = getAllowedUserLimit(zoneId);
+        if (maxAllowedUsers > 0 && getZoneCount(zoneId, maxAllowedUsers) > (maxAllowedUsers -1)) {
```
adrianhoelzl-sap: Please add a comment here that we compare against `maxAllowedUsers - 1` rather than `maxAllowedUsers` since we need capacity for the one user that is currently being created.

server/src/main/java/org/cloudfoundry/identity/uaa/zone/UserConfigValidator.java
hsinn0: Either that or remove `maxUsers > Integer.MAX_VALUE` from the condition. `maxUsers > Integer.MAX_VALUE` can never be true.
If we decide to keep `Integer.MAX_VALUE` as the actual limit, then I suggest that we just set `UserConfig.maxUsers` to `Integer.MAX_VALUE` instead. I see no benefit to set it as `-1` and do extra logic.

hsinn0 (changes requested):

server/src/main/java/org/cloudfoundry/identity/uaa/zone/UserConfigValidator.java
strehle: omitted Integer.MAX_VALUE

adrianhoelzl-sap (approved):

model/src/main/java/org/cloudfoundry/identity/uaa/zone/UserConfig.java
```diff
@@ -22,6 +22,8 @@ public class UserConfig {
 
     private List<String> defaultGroups = DEFAULT_ZONE_GROUPS;
 
+    private long maxUsers = -1;
```
hsinn0: Why is `maxUsers` `long` instead of `int`? Do we need it to be more than max int value?

server/src/main/java/org/cloudfoundry/identity/uaa/zone/UserConfigValidator.java
```diff
@@ -9,10 +9,19 @@ private UserConfigValidator() {
...
+        Set<String> allowedGroups = config.resultingAllowedGroups();
+        if (allowedGroups != null && allowedGroups.isEmpty()) {
             String message = "At least one group must be allowed";
             throw new InvalidIdentityZoneConfigurationException(message);
         }
+
+        long maxUsers = config.getMaxUsers();
+        if (maxUsers < -1 || maxUsers == 0) {
+            throw new InvalidIdentityZoneConfigurationException("Maximum number of users in the zone is invalid, either use -1 or a value more than 0.");
+        }
```
hsinn0, line 25: Another option is to make the default `maxUsers` value `0` and `0` value to mean unlimited. Then this condition for example can be changed to `maxUsers < 0`. I think doing so will result in more readable code.

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
```diff
@@ -111,6 +116,8 @@ public Logger getLogger() {
 
     public static final String HARD_DELETE_BY_PROVIDER = "delete from users where identity_zone_id = ? and origin = ?";
 
+    public static final String USER_COUNT_BY_ZONE = "select count(*) from users where identity_zone_id = ? limit ?";
```
hsinn0: I think that using `limit` here is wrong. It is my understanding that `select count` query always returns 1 row. `limit` limits the number of returned rows and do nothing for this query. Also looking at the code that uses this query, which requires the 2nd parameter, makes it all confusing. Remove `limit` from this query.

hsinn0 (changes requested):

server/src/main/java/org/cloudfoundry/identity/uaa/zone/UserConfigValidator.java
strehle: This is the validator for the rest interface. I dont see this as a complicated logic. It should simply help to have clean values. 
We use -1 for many other numbers which the meaning of "do nothing", e.g. token or password policy. Here -1 is always no check so I would like to have -1 because we have it already.

Allowing no user maxUsers == 0 is something if you dont want it I can remove, but then no user in a zone could be created... 

So would you agree with 

if (maxUsers < -1) {
            throw new InvalidIdentityZoneConfigurationException("Maximum number of users in the zone is invalid, either use -1 or a value.");
        }

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
strehle: for postgresql I thought it could help ,e.g. https://dba.stackexchange.com/questions/117284/count-gives-more-than-1-with-limit-1 , but yes I will remove it

model/src/main/java/org/cloudfoundry/identity/uaa/zone/UserConfig.java
strehle: because we can have more than 2 mio users and simply the compare functions should always run with long , simply to omit casts
A select count with int as return type means you never expect ot have more values in your DB but we expect to have many values :-)

server/src/main/java/org/cloudfoundry/identity/uaa/zone/UserConfigValidator.java
hsinn0: > We use -1 for many other numbers which the meaning of "do nothing"

OK, then I am find to keep it as it is.

> Allowing no user maxUsers == 0 is something if you dont want it I can remove

No, you should keep it as the default will stay as `-1`.

strehle: thanks close this

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
strehle: changed to int , removed limit

hsinn0 (approved):

Issue #2560, opened 2023-10-19
Spring Security Oauth2 End of Life


strehle:
UAA uses https://github.com/spring-attic/spring-security-oauth
This was no issues in the past but with the Spring Boot 3 update the library needs to be changed or removed

Similar issue in past https://github.com/cloudfoundry/uaa/issues/1728

Plan for Removal.
- copy base classes (most needed) and exceptions into UAA
- migrate other: https://github.com/spring-projects/spring-security/wiki/OAuth-2.0-Migration-Guide

fhanik:
I did a similar migration this week on another project. The biggest helpful thing I did was to create classed `LegacyClientDetails` and `LegacyBaseClientDetails` to retain the JSON structure of clients as a data object.

strehle:
> I did a similar migration this week on another project. The biggest helpful thing I did was to create classed `LegacyClientDetails` and `LegacyBaseClientDetails` to retain the JSON structure of clients as a data object.

@fhanik to me UAA is not legacy, so I will do this but called in UaaClientDetails :-)
Similar things I will to with exceptions, means I will move/copy them into UAA context, because many of them are still in use there.

fhanik:
@strehle - I also found this answer to support the majority of the lifting when creating the new RestTemplates

https://stackoverflow.com/a/60634690

PR #2624, opened 2023-11-28
Refactor: remove spring security jwt and use nimbus jose

strehle:
Intro:
UAA uses https://github.com/spring-attic/spring-security-oauth
See https://github.com/cloudfoundry/uaa/issues/2560 , we need to replace it

Replacement with library com.nimbusds:nimbus-jose-jwt see https://github.com/cloudfoundry/uaa/blob/develop/dependencies.gradle#L125
Nimbus in use already for some fixes in past...
Details :
https://connect2id.com/products/nimbus-jose-jwt

There are some incompatibilities with this replacement
* HMAC signature
  - key type was in UAA always MAC, but standard needs oct. Added workaround in JWK creation
  - key length . The key size must fit to algorithm, e.g. HS256 needs 32 bit key length. Created UaaMacSigner fork to allow old keys
  - JWS header parser. x5c and jwk header had wrong type, use now generic object, because we dont need the keys but we need to parse the header

* RSA signature
  - for small keys (<2KB) there is a deprecated allow option. Added this to have currently support for old keys. Need to have a solution in general. 
* JWS header processing
 - JWT header was not processed as expected, means a JW header with alg HS256 was verified internally with a RSA key. So signature validation was done on internal configured key and the header from JWT was not really checked. If this happed in past then this is a severe security issue which cannot happen with nimbus. Mentioned it only for completeness

Good news.
UAA allows now ESxxx signature family. The spring library had only support for validation of ES, now we can sign and verify all specified JWS signing algorithms!!!


tack-sap (approved):

Tallicia:
Will this be updated soon to address the merge conflicts?

strehle:
> Will this be updated soon to address the merge conflicts?

Do you see a merge conflict in this PR?

strehle:
@hsinn0 I know this is a bigger PR, but this should be approved by vmware . 
It removes the spring-security-jwt library . We use nimbus already, but now we move completly to it, finally this PR is a pre-requisit for the spring boot 3 upgrade

model/src/main/java/org/cloudfoundry/identity/uaa/util/ObjectUtils.java
```diff
@@ -54,4 +54,8 @@ public static int countNonNull( Object... objects ) {
         }
         return count;
     }
+
+    public static boolean isNotEmpty(Object object) {
+        return !org.springframework.util.ObjectUtils.isEmpty(object);
+    }
```
hsinn0, line 60: I do not see much value in addition of this method. Additional method just to rerun NOT result of another method? However, I guess that it is preference matter, so it is OK to me if you want to keep it this way.

model/src/main/java/org/cloudfoundry/identity/uaa/util/UaaStringUtils.java
```diff
@@ -317,4 +318,12 @@ public static String getSafeParameterValue(String[] value) {
...
         return StringUtils.hasText(value[0]) ? value[0] : EMPTY_STRING;
     }
+
+    public static Set<String> getArrayDefaultValue(Set<String> values, String defaultValue) {
+        if (ObjectUtils.isEmpty(values)) {
+            return Set.of(defaultValue);
+        } else {
+            return values;
+        }
+    }
```
hsinn0, line 330: I am having a hard time correlating this method name to what it actually does, and having difficulty to understand the method's purpose. Maybe you can call it `getValuesOrDefaultValue` and add Javadoc about the purpose of this method?

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/InvalidSignatureException.java
```diff
@@ -0,0 +1,11 @@
+package org.cloudfoundry.identity.uaa.oauth;
+
+public class InvalidSignatureException extends RuntimeException {
+  public InvalidSignatureException(String message) {
+    super(message);
+  }
+
+  public InvalidSignatureException(String message, Exception exception) {
+    super(message, exception);
+  }
```
hsinn0, line 16: Usually, this Exception ctor's 2nd argument type is `Throwable`, unless you have a specific reason not to do so.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/InvalidSignatureException.java
```diff
@@ -0,0 +1,11 @@
+package org.cloudfoundry.identity.uaa.oauth;
+
+public class InvalidSignatureException extends RuntimeException {
```
hsinn0, line 3: Technically speaking, `static final serialVersionUID` should be defined for classes that implements `Serializable`. That will include any classes that extend `Throwable`. I know other UAA classes do not have it, but we might want to start doing it for new classes.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/KeyInfo.java
```diff
@@ -188,7 +184,9 @@ private Optional<X509Certificate> getValidX509Certificate(String pemEncoded) {
                 x509Certificate.checkValidity();
                 return Optional.of(x509Certificate);
             }
-        } catch (RuntimeException | CertificateExpiredException | CertificateNotYetValidException e) { } // ignore
+        } catch (RuntimeException | CertificateExpiredException | CertificateNotYetValidException e) {
+            // ignore
+        }
```
hsinn0: I know that is is existing code, but is a good idea to ignore the exceptions here without even any logging?

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/jwt/SignatureVerifier.java
```diff
@@ -0,0 +1,112 @@
...
+ *
+ *     This product is licensed to you under the Apache License, Version 2.0 (the "License").
+ *     You may not use this product except in compliance with the License.
+ *
+ *     This product includes a number of subcomponents with
+ *     separate copyright notices and license terms. Your use of these
+ *     subcomponents is subject to the terms and conditions of the
+ *     subcomponent's license, as noted in the LICENSE file.
+ * ****************************************************************************
+ */
```
hsinn0: This reminds me a question I had before. Is there specific copyright statement that we are supposed to use for UAA code currently?

server/src/main/java/org/cloudfoundry/identity/uaa/util/JwtTokenSignedByThisUAA.java
```diff
@@ -126,12 +128,12 @@ public JwtTokenSignedByThisUAA checkSignature() {
         return checkSignature(fetchSignatureVerifierFromToken(this.tokenJwt));
     }
 
-    public JwtTokenSignedByThisUAA checkSignature(SignatureVerifier verifier) {
+    public JwtTokenSignedByThisUAA checkSignature(Object verifier) {
```
hsinn0: It would be better if we can have a base type that can represent verifiers instead of taking `Object` type here. I suppose that you considered that already?

server/src/main/java/org/cloudfoundry/identity/uaa/util/JwtTokenSignedByThisUAA.java
```diff
@@ -141,30 +143,22 @@ public JwtTokenSignedByThisUAA checkIssuer(String issuer) {
...
-        if (!claims.containsKey(ISS)) {
+        JWTClaimsSet jwtClaimsSet = getJwt().getClaimSet();
+        if (jwtClaimsSet.getIssuer() == null) {
             throw new InvalidTokenException("Token does not bear an ISS claim.", null);
         }
 
-        if (!equals(issuer, claims.get(ISS))) {
+        if (!equals(issuer, jwtClaimsSet.getIssuer())) {
             throw new InvalidTokenException("Invalid issuer (" + claims.get(ISS) + ") for token did not match expected: " + issuer, null);
         }
```
hsinn0, line 153: Any chance that `claims` is null and `claims.get()` call causes NullPointerException here?

hsinn0 (commented):
It is a large PR and I did my best, but had to be selective in what to look into. Especially, I skipped most of the test class changes.

model/src/main/java/org/cloudfoundry/identity/uaa/util/ObjectUtils.java
strehle: yeah many other libraries provide also isEmpty or isNotEmpty but sometimes they deprecate their usage e.g. spring did in past so these utils should make code in uaa cleaner , esp. in stream methods we can do filters with isNotEmpty etc. So I would like to stay with it....and will do more usages with refactoring later.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/jwt/SignatureVerifier.java
strehle: thanks, not wanted

model/src/main/java/org/cloudfoundry/identity/uaa/util/UaaStringUtils.java
strehle: done

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/InvalidSignatureException.java
strehle: done

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/InvalidSignatureException.java
strehle: done

server/src/main/java/org/cloudfoundry/identity/uaa/util/JwtTokenSignedByThisUAA.java
strehle: done

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/InvalidSignatureException.java
```diff
@@ -0,0 +1,17 @@
...
+
+  private static final long serialVersionUID = 1L;
```
hsinn0: `serialVersionUID` should be generated with `serialver` JDK command or similar tool instead. You can also use IntelliJ IDEA to generate it while editing the source code.

strehle: done, but according to https://stackoverflow.com/questions/21850907/what-means-1l-serialversionuid-when-could-i-use-this-default-value-1l 
I used 1L

hsinn0: Hmm, that makes sense, too.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/KeyInfo.java
hsinn0: Not sure if you have seen this question. Can you answer it when you have chance?

hsinn0 (approved):

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/KeyInfo.java
strehle: I removed the changes. We dont have a logger in KeyInfo and adding one for this would be too much , at least this is not important to know, so if there would be a logger in the class I could to a simply log.

So I would like to stay as it is. The meaning is, if the x509 is invalid , it wont be used for JWT keys

server/src/test/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManagerTest.java
```diff
@@ -154,16 +156,19 @@ public void cleanup() {
...
         expectedException.expect(InvalidTokenException.class);
         expectedException.expectMessage("Could not verify token signature.");
 
         Map<String, Object> header = map(
-                entry(HeaderParameterNames.ALGORITHM, JWSAlgorithm.HS256.getName()),
+                entry(HeaderParameterNames.ALGORITHM, JWSAlgorithm.RS256.getName()),
                 entry(HeaderParameterNames.KEY_ID, OIDC_PROVIDER_KEY)
         );
-        Signer signer = new RsaSigner(changedOidcProviderTokenSigningKey);
+        JWSSigner signer = new KeyInfo(OIDC_PROVIDER_KEY, oidcProviderTokenSigningKey, "http://localhost:8080/uaa").getSigner();
```
Tallicia: The hard coded port 8080 looks concerning and should be centralized. we have run into issues were non default ports don't behave as expected or parameters to use different ports are ignored. This is a test and likely under more control, but good practice to keep them centralized and coordinated with startup/config

Tallicia (approved):
This is quite the PR, In the future if possible to reduce this to smaller segments, that would be fantastic. Though I spent a considerable amount of time, I could have easily spent three times as much to deeply scrutinize, I left comments, and are more for refactoring options following this PR than required as part of this PR. 

Overall this is approved, It was good a large majority was duplicated updates and changes and many removals, though the additions require significant deep dives, which I do not feel I completely dove all the way through, but trust based on what I saw and the tests, it was meeting desired objectives.

model/src/main/java/org/cloudfoundry/identity/uaa/util/UaaStringUtils.java
```diff
@@ -49,6 +49,8 @@ public final class UaaStringUtils {
 
     public static final String EMPTY_STRING = "";
 
+    public static final String DEFAULT_UAA_URL = "http://localhost:8080/uaa";
```
Tallicia, line 52: Nice, that is a step in the right direction. Following this PR, the port should be defaulted to 8080 with the capability to override.

Tallicia (approved):
Looks good

Tallicia:
@strehle The PR is indicating "This branch cannot be rebased due to conflicts" Are you planning to rebase to address whatever conflicts github is referring to?

strehle:
> @strehle The PR is indicating "This branch cannot be rebased due to conflicts" Are you planning to rebase to address whatever conflicts github is referring to?

yes, I would like to release soon (after our meeting) whithout this PR and then I merge and work on further refactorings . .... because of our pipeline hickups I waited here

PR #2625, opened 2023-11-28
fixes from non OIDC standard tests

strehle:
* main issue. was: audience is empty in same cases
* tests corrected, there where HS256 was used to create RS256 signature  (new library checks it)
* HMAC key length corrected (new library checks it)


strehle:
sonar 
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2625

strehle:
close this PR, because it was mainly to discuss the incompatibilities with new nimbus-jose-jwt. To make it clear. The problems are in UAA, because UAA was/is not compatible to RFCs according to JWT processing and headers. Nevertheless, we cannot break existing customers, therefore for all issues, I created exceptions in https://github.com/cloudfoundry/uaa/pull/2624

No need to keep
https://github.com/cloudfoundry/uaa/pull/2625

PR #2637, opened 2023-12-06
Alias ID and Alias ZID for Identity Providers

adrianhoelzl-sap:
Issue #2505 

Identity providers may be copied from the custom identity zone to the default identity zone (or vice versa) to enable customers to do user management in own identity zones


klaus-sap:
If alias_zid is set, replicate entry into alias identity zone. From default zone "uaa" you may replicate into any custom zone. From any custom zone, you may replicate into default zone "uaa" only.

model/src/main/java/org/cloudfoundry/identity/uaa/provider/IdentityProvider.java
```diff
@@ -208,6 +232,8 @@ public int hashCode() {
         result = prime * result + ((name == null) ? 0 : name.hashCode());
         result = prime * result + ((originKey == null) ? 0 : originKey.hashCode());
         result = prime * result + ((type == null) ? 0 : type.hashCode());
+        result = prime * result + ((aliasId == null) ? 0 : aliasId.hashCode());
+        result = prime * result + ((aliasZid == null) ? 0 : aliasZid.hashCode());
```
adrianhoelzl-sap, line 234: To be clarified: are these hash codes persisted anywhere and might therefore become a breaking change?

model/src/main/java/org/cloudfoundry/identity/uaa/provider/IdentityProvider.java
```diff
@@ -269,6 +309,21 @@ public String toString() {
         sb.append(", name='").append(name).append('\'');
         sb.append(", type='").append(type).append('\'');
         sb.append(", active=").append(active);
+
```
adrianhoelzl-sap, line 318: To be clarified: do we also want to add the identity zone ID field here?

server/src/main/java/org/cloudfoundry/identity/uaa/provider/IdentityProviderEndpoints.java
```diff
@@ -119,34 +135,56 @@ public ResponseEntity<IdentityProvider> createIdentityProvider(@RequestBody Iden
             samlConfigurator.validateSamlIdentityProviderDefinition(definition);
             body.setConfig(definition);
         }
+
```
adrianhoelzl-sap, line 153: Are there any restrictions on the IdPs we want to allow mirroring for? For example, should we allow mirroring only for IdPs of certain types?

strehle:
@adrianhoelzl-sap please check findings from sonar

https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2637

model/src/main/java/org/cloudfoundry/identity/uaa/provider/IdentityProvider.java
Tallicia: Do we have an answer to this?

server/src/test/java/org/cloudfoundry/identity/uaa/provider/IdentityProviderEndpointsTest.java
```diff
@@ -437,6 +722,47 @@ void testDeleteIdentityProviderExisting() {
         assertEquals(validIDP, deleteResponse.getBody());
     }
 
+    @Test
+    void testDeleteIdentityProviderMirrored() {
```
Tallicia: What does mirrored mean here? Is a duplicate 100% redundant idp? If so, shouldn't that be protected against being created? If mirrored is something else, what does it mean?

Looking through more of the PR, I am trying to evaluate what the term mirrored is intending and alignment with the PR intent.

Tallicia:
@klaus-sap 
> If alias_zid is set, replicate entry into alias identity zone. From default zone "uaa" you may replicate into any custom zone. From any custom zone, you may replicate into default zone "uaa" only.

How is this protected against collisions for multiple concurrent changes occurring in different identity zones? Is there a possibility of a race condition or a cross talk between zones?

server/src/test/java/org/cloudfoundry/identity/uaa/provider/IdentityProviderEndpointsTest.java
adrianhoelzl-sap: For an IdP, its mirrored IdP is the copy of it in another identity zone (see the diagram in issue #2505). The "original" IdP and the mirrored IdP are identical except for the fields `id`, `zid`, `alias_id` and `alias_zid`. In the original IdP, the alias fields reference its mirrored IdP and vice versa.

adrianhoelzl-sap:
> @klaus-sap
> 
> > If alias_zid is set, replicate entry into alias identity zone. From default zone "uaa" you may replicate into any custom zone. From any custom zone, you may replicate into default zone "uaa" only.
> 
> How is this protected against collisions for multiple concurrent changes occurring in different identity zones? Is there a possibility of a race condition or a cross talk between zones?

We perform all operations on identity providers that involve the creation or update of a mirrored IdP in transactions. Therefore, no such inconsistencies should occur.

server/src/test/java/org/cloudfoundry/identity/uaa/provider/IdentityProviderEndpointsTest.java
Tallicia: My confusion lies in the term mirror. that implies it is kept in sync and should remain a "mirror" or identical to the original.  Though in this case, it is a copy/duplicate, which can over time diverge and become different, especially in regards to permissions, yes? This is what is declared in the doc: "We propose copying users instead of creating references, because we want to create separate user records in default and custom identity zone."

It looks like the term mirror was used and has carried through in many places which further creates confusion.

adrianhoelzl-sap: The two IdPs are actually kept in sync. Any property change to one of them is propagated to the other, respectively.

The sentence you mentioned refers to us copying the IdPs (or later also users) instead of using "pointers" to IdPs in the corresponding other identity zone. This way, we do not need to add logic for resolving these pointers whenever an operation on an IdP is performed. Instead, the existing functionalities work out-of-the-box for the mirrored and original IdP, as both of them are - apart from their `aliasId` and `aliasZid` properties - similar to any other IdP. We just need to pay attention that the changes are propagated to the mirrored user during CRUD operations.

server/src/main/java/org/cloudfoundry/identity/uaa/provider/IdentityProviderEndpoints.java
adrianhoelzl-sap: In particular, should it be possible to mirror IdPs of type `UAA`?

adrianhoelzl-sap:
> > @klaus-sap
> > > If alias_zid is set, replicate entry into alias identity zone. From default zone "uaa" you may replicate into any custom zone. From any custom zone, you may replicate into default zone "uaa" only.
> > 
> > 
> > How is this protected against collisions for multiple concurrent changes occurring in different identity zones? Is there a possibility of a race condition or a cross talk between zones?
> 
> We perform all operations on identity providers that involve the creation or update of a mirrored IdP in transactions. Therefore, no such inconsistencies should occur.

For each Identity Provider (and later also user), there is only one "custom" zone that can be the alias zone. These entities can only be mirrored from or to the uaa zone. Therefore, there cannot be multiple custom zones in which the same entity in the uaa zone is changed.
The only remaining potential issue is that an IdP is modified both in the custom zone and in the uaa zone. However, since we use transactions for the update of the original and the mirrored entity, respectively, the database decides which transaction is accepted.

server/src/test/java/org/cloudfoundry/identity/uaa/provider/JdbcIdentityProviderProvisioningTests.java
```diff
@@ -63,6 +66,52 @@ void deleteProvidersInZone() {
         assertThat(jdbcTemplate.queryForObject("select count(*) from identity_provider where identity_zone_id=?", new Object[]{otherZoneId1}, Integer.class), is(0));
     }
 
+    @Test
+    void deleteProvidersInUaaZone_Mirrored() {
+        final IdentityZone mockIdentityZone = mock(IdentityZone.class);
+        when(mockIdentityZone.getId()).thenReturn(otherZoneId1);
+
+        final String originSuffix = RandomStringUtils.randomAlphabetic(5);
```
strehle: because UAA uses random very often, we have AlphanumericRandomValueStringGenerator withint this module

model/src/main/java/org/cloudfoundry/identity/uaa/provider/IdentityProvider.java
strehle: hashCode is needed for serialization and compare (means to get the info about changes in an object)

model/src/main/java/org/cloudfoundry/identity/uaa/provider/IdentityProvider.java
strehle: if the field is set I suggest to show it

server/src/main/java/org/cloudfoundry/identity/uaa/provider/IdentityProviderEndpoints.java
strehle: maybe ignore type UAA (or do the sync only for SAML and OAutth2/OIDC) because internal origins like UAA and LDAP are not really zone aware. They need extra config. Origin UAA is special in zone UAA and maybe not a good idea to sync with other zones

model/src/main/java/org/cloudfoundry/identity/uaa/provider/IdentityProvider.java
Tallicia: They are not persisted anywhere, correct?

server/src/test/java/org/cloudfoundry/identity/uaa/provider/JdbcIdentityProviderProvisioningTests.java
Tallicia: I concur, this should maintain consistency with other patterns, please update.

model/src/main/java/org/cloudfoundry/identity/uaa/provider/IdentityProvider.java
adrianhoelzl-sap: We also assumed that they are not persisted in any way. Therefore, I will mark this discussion as resolved.

model/src/main/java/org/cloudfoundry/identity/uaa/provider/IdentityProvider.java
adrianhoelzl-sap: Fixed

adrianhoelzl-sap: I saw that `identityZoneId` is not used in `hashCode` and `equals`. Should I add it there as well?

server/src/test/java/org/cloudfoundry/identity/uaa/provider/JdbcIdentityProviderProvisioningTests.java
adrianhoelzl-sap: Fixed

server/src/main/java/org/cloudfoundry/identity/uaa/provider/IdentityProviderEndpoints.java
adrianhoelzl-sap: In commit 862405896a1bcb8da76042569c71c72705c1ea35, I added an allowlist that contains SAML, OIDC and OAuth2.0. Are there any values I should add to/remove from there?

server/src/main/java/org/cloudfoundry/identity/uaa/provider/JdbcIdentityProviderProvisioning.java
```diff
@@ -114,12 +116,18 @@ public IdentityProvider update(final IdentityProvider identityProvider, String z
         validate(identityProvider);
         jdbcTemplate.update(UPDATE_IDENTITY_PROVIDER_SQL, ps -> {
             int pos = 1;
+
+            // placeholders in INSERT INTO
             ps.setInt(pos++, identityProvider.getVersion() + 1);
-            ps.setTimestamp(pos++, new Timestamp(new Date().getTime()));
+            ps.setTimestamp(pos++, new Timestamp(new Date().getTime())); // lastmodified
```
Tallicia, line 122: What does this last modified comment mean?

adrianhoelzl-sap: This is just an indication that this statement sets the value for the `lastmodified` column, as it is not obvious from the value itself.

server/src/main/java/org/cloudfoundry/identity/uaa/provider/IdentityProviderEndpoints.java
```diff
@@ -241,7 +344,7 @@ public ResponseEntity<String> testIdentityProvider(@RequestBody IdentityProvider
             status = BAD_REQUEST;
             exception = getExceptionString(x);
         } catch (Exception x) {
-            logger.error("Identity provider validation failed.", x);
+            logger.warn("Identity provider validation failed.", x);
```
Tallicia: Why are we lowering this from an error to a warn?

server/src/main/java/org/cloudfoundry/identity/uaa/provider/IdentityProviderEndpoints.java
Tallicia: I believe the 3 listed are all that is needed - SAML, OIDC and OAuth2.0

server/src/main/java/org/cloudfoundry/identity/uaa/provider/IdentityProviderEndpoints.java
adrianhoelzl-sap: I used the `error` level for some of the log messages I added in the new code and reverted them - and accidentally also this one - back to `warn` later. I will change this particular call back to `error`, thanks for noticing :)

adrianhoelzl-sap: fixed

uaa/src/main/resources/uaa.yml
```diff
@@ -553,6 +553,8 @@ uaa:
         - GET
         - HEAD
         - OPTIONS
+  features:
+#    aliasEntitiesEnabled: true
```
strehle: I would omit this here, because if you want have it for documentation reasons you should add it in uaa-release

Then next question is: should be really call it features.
I would have expected a configuration option simply in login.xxxx name can be aliasEntitiesEnabled because it is a feature yes, but it should also help to configure behaviour... so I dont see this as a pure nice to have...
Many other configuration options in UAA are feature, but we dont use this term

hsinn0: I think that having the `aliasEntitiesEnabled` setting (commented out) here as an example is a good thing. It will certainly help me if I want to check out the behavior when the setting is `true`.
Regarding calling it `features`, I am agreeing to @strehle. I think it can be called something specific instead.

hsinn0: I also think calling this implementation as "feature flag" is not proper. When I first heard it today, I thought about IAM or role-based feature flag, which is turned on/off depending on who you are. However, here we have one configuration option that affects the whole instance.

Tallicia: The feature flag terminology appears aligned with cloud foundry usage as in the docs: https://docs.cloudfoundry.org/adminguide/listing-feature-flags.html

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/providers/IdentityProviderEndpointsAliasMockMvcTests.java
```diff
@@ -0,0 +1,1208 @@
...
+                        null,
+                        null,
+                        existingIdpInZone2.getOriginKey(), // same origin key
+                        OIDC10
+                );
+                final IdentityProvider<?> providerInZone1 = createIdp(zone1, idp);
+
+                // update the alias ZID to zone 2, where an IdP with this origin already exists -> should fail
+                providerInZone1.setAliasZid(zone2.getId());
+                shouldRejectUpdate(zone1, providerInZone1, HttpStatus.CONFLICT);
```
Tallicia, line 552: I like the use of a 409 response here.

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/providers/IdentityProviderEndpointsAliasMockMvcTests.java
```diff
@@ -0,0 +1,1208 @@
...
+                passwordPolicy.setRequireUpperCaseCharacter(1);
+                passwordPolicy.setRequireLowerCaseCharacter(1);
+                passwordPolicy.setRequireSpecialCharacter(1);
+                passwordPolicy.setPasswordNewerThan(new Date(System.currentTimeMillis()));
+                return new UaaIdentityProviderDefinition(passwordPolicy, null);
+            default:
+                throw new IllegalArgumentException("IdP type not supported.");
+        }
+    }
+}
```
Tallicia, line 1211: There appears to be good coverage here, though I wonder if something is missed. I would prefer if there was a doc or matrix indicating conditions that maps to the tests to validate if any gaps exist.

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/providers/IdentityProviderEndpointsAliasMockMvcTests.java
```diff
@@ -0,0 +1,1208 @@
...
+        assertThat(referencedIdp.getIdentityZoneId()).isNotBlank().isEqualTo(idp.getAliasZid());
+    }
+
+    private void assertOtherPropertiesAreEqual(final IdentityProvider<?> idp, final IdentityProvider<?> aliasIdp) {
+        // the configs should be identical
+        final OIDCIdentityProviderDefinition originalIdpConfig = (OIDCIdentityProviderDefinition) idp.getConfig();
+        final OIDCIdentityProviderDefinition aliasIdpConfig = (OIDCIdentityProviderDefinition) aliasIdp.getConfig();
+        assertThat(aliasIdpConfig).isEqualTo(originalIdpConfig);
+
+        // check if remaining properties are equal
```
Tallicia, line 965: Are these 3 remaining properties the only relevant properties, if so, should this test be called assertImportantPropertiesAreEqual or Relevant, Primary, or Significant Properties? There are many other properties like LastModified, version, identityZoneID and such that are not checked.

model/src/main/java/org/cloudfoundry/identity/uaa/provider/IdentityProvider.java
adrianhoelzl-sap: (will be done in a separate PR)

uaa/src/main/resources/uaa.yml
adrianhoelzl-sap: I moved the flag to `login`

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/providers/IdentityProviderEndpointsAliasMockMvcTests.java
adrianhoelzl-sap: I checked all properties again and found out that I missed to check for the `active` field. The only properties that won't be equal are the two timestamps `created` and `lastmodified` as well as the `version` field, which is fine IMO.

I added the missing check as well as a comment about the other properties in f12dc96.

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/providers/IdentityProviderEndpointsAliasMockMvcTests.java
swalchemist: There is a way to see the test coverage on the new code [in SonarCloud](https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2637). Overall coverage on new code is reported as 80.7% currently, just over the minimum threshold.

If I drill into IdentityProviderEndpoints.java, I see some `if` statement bodies with non-trivial code that would be nice to get test coverage for, plus some exceptions that aren't tested.

Also, Sonar indicates that there is one new code smell introduced in the PR that I'd like to see addressed.

adrianhoelzl-sap: I resolved the Sonar finding with e5f5791 and added some unit tests to increase the coverage with fafe033.

server/src/main/java/org/cloudfoundry/identity/uaa/provider/IdpAliasFailedException.java
```diff
@@ -15,24 +15,51 @@ public IdpAliasFailedException(
...
+            @Nullable final String aliasId,
+            @Nullable final String aliasZid,
+            @NonNull final Reason reason,
+            @Nullable final Throwable cause
+    ) {
+        super(cause, ERROR, buildMessage(idpId, idzId, aliasId, aliasZid, reason), reason.responseCode.value());
     }
 
-    private static String buildMessagePrefix(@NonNull final IdentityProvider<?> idp) {
+    private static String buildMessagePrefix(
```
swalchemist, line 39: I'm not sure what code smell this particular change is addressing, but I think having 4 parameters vs. a single parameter class is also a code smell.

server/src/main/java/org/cloudfoundry/identity/uaa/provider/IdentityProviderEndpoints.java
```diff
@@ -541,27 +541,43 @@ private <T extends AbstractIdentityProviderDefinition> IdentityProvider<?> ensur
             return originalIdp;
         }
 
+        final IdentityProvider<?> persistedAliasIdp = createNewAliasIdp(aliasIdp, originalIdp.getAliasZid());
```
swalchemist, line 544: I like the change reducing the length of a long method. I'm curious what the Sonar smells are that all of these changes are addressing.

swalchemist (approved):
Approved. If you want to make further refactoring from my comments here, I recommend merging this as-is and put additional changes in a new PR, so you can get this PR merged.

server/src/main/java/org/cloudfoundry/identity/uaa/provider/IdpAliasFailedException.java
adrianhoelzl-sap: The reason I changed this is the new method `createNewAliasIdp` in `IdentityProviderEndpoints`. With just the other constructor of `IdpAliasFailedException` being present, we would have to pass the original IdP as a parameter just for the exception case (e.g., https://github.com/cloudfoundry/uaa/pull/2637/commits/e5f579179e6d1c0486cd814ac6b817cbae848b21#diff-90da5e48423637b46ed07b703063b4fca4a0ee47cca8286548c91941b532fd07R566), even though all of the information for the exception message is also present in the alias IdP.

server/src/main/java/org/cloudfoundry/identity/uaa/provider/IdentityProviderEndpoints.java
adrianhoelzl-sap: This is the fix to the Sonar warning about the "Cognitive Complexity" being too high in this method.

PR #2652, opened 2023-12-21
fix: check for existing groups

klaus-sap:
Update IdZ: check for existing groups was not working correctly. Fixed it and added an IT
@strehle please add to release


strehle:
empty now

PR #2660, opened 2023-12-27
upgrade spring security version 5.8.9

strehle:
prepare for spring boot upgrade 3.x

see https://docs.spring.io/spring-security/reference/5.8/migration/index.html


peterhaochen47:
Hi @strehle, I interpret this PR as doing the following 2 things:
- allow us to explicitly define the versions of springframework and spring-security related libraries (before, it was just whatever versions springboot provides)
- upgrade spring-security to 5.8.9. 

Could you help me understand why these 2 changes are "[preparing] for spring boot upgrade 3.x"? Are we trying to upgrade spring-security before the springboot upgrade?

strehle:
@peterhaochen47 yes, this PR includes  https://github.com/cloudfoundry/uaa/pull/2659

Reason:
- https://github.com/cloudfoundry/uaa/pull/2659 is simply a preparation for the time where we have no spring boot updates anymore, but we will have spring framework and security updates, e.g. https://github.com/spring-projects/spring-framework/milestone/343 , same for spring security see https://github.com/spring-projects/spring-security/milestone/329.

- this update from 5.7x to 5.8.x is an extra PR to have a discussion about the step of upgrade to spring security to 5.8.x before going to spring boot 3. see https://docs.spring.io/spring-security/reference/5.8/migration/index.html#_update_to_spring_security_5_8 . the spring security update is no pre-requisit but I would do it as being prepared simply

peterhaochen47:
@strehle I see. Makes sense. So #2659 would be a temporary measure, and after spring boot 3 upgrade we can reverse this refactor (since we'll be getting the spring security updates automatically from spring boot 3).

peterhaochen47 (approved):

PR #2663, opened 2024-01-03
Alias ID and Alias ZID for Users

### LINKED ISSUE #2505: Missing user management from custom IdZ to CC

adrianhoelzl-sap:
Issue #2505 

Users may be copied from a custom identity zone to the default identity zone (or vice versa) to enable customers to do user management in own identity zones.

Prerequisites:

- #2637 
- #2769


Tallicia:
@adrianhoelzl-sap Will you be updating this so it is able to be merged?

adrianhoelzl-sap:
> @adrianhoelzl-sap Will you be updating this so it is able to be merged?

I resolved the merge conflicts earlier today, does it still show conflicts for you?

server/src/main/java/org/cloudfoundry/identity/uaa/scim/endpoints/ScimUserEndpoints.java
```diff
@@ -224,7 +250,34 @@ public ScimUser createUser(@RequestBody ScimUser user, HttpServletRequest reques
...
+        user.setZoneId(identityZoneManager.getCurrentIdentityZoneId());
+
+        if (!aliasHandler.aliasPropertiesAreValid(user, null)) {
+            throw new ScimException("Alias ID and/or alias ZID are invalid.", HttpStatus.BAD_REQUEST);
+        }
+
+        // create the user and an alias for it if necessary
+        ScimUser scimUser;
+        try {
+            scimUser = transactionTemplate.execute(txStatus -> {
```
strehle, line 279: you know I dont like transactions in UAA, but if you really want use them, can you please use them in scenario where this feature is enabled ( aliasEntitiesEnabled == true ) and all other UAA users (aliasEntitiesEnabled == false) create a user with one call only.

server/src/main/java/org/cloudfoundry/identity/uaa/scim/endpoints/ScimUserEndpoints.java
```diff
@@ -161,11 +184,14 @@ public ScimUserEndpoints(
...
         this.codeStore = codeStore;
         this.approvalStore = approvalStore;
+        this.aliasEntitiesEnabled = aliasEntitiesEnabled;
         this.userMaxCount = userMaxCount;
         this.membershipManager = membershipManager;
         this.messageConverters = new HttpMessageConverter[] {
                 new ExceptionReportHttpMessageConverter()
         };
+        this.aliasHandler = aliasHandler;
+        this.transactionTemplate = new TransactionTemplate(transactionManager);
```
strehle: hard to test if you create the TransactionTemplate in code here , so not easy to mock then

server/src/main/java/org/cloudfoundry/identity/uaa/scim/endpoints/ScimUserEndpoints.java
adrianhoelzl-sap: fixed with commit ced0186

server/src/main/java/org/cloudfoundry/identity/uaa/scim/endpoints/ScimUserEndpoints.java
adrianhoelzl-sap: fixed with commit b94a96d

server/src/main/java/org/cloudfoundry/identity/uaa/scim/endpoints/ScimUserEndpoints.java
strehle: ok

server/src/main/java/org/cloudfoundry/identity/uaa/scim/endpoints/ScimUserEndpoints.java
strehle: ok

strehle (approved):
LGTM

PR #2669, opened 2024-01-09
Dependabot cleanup

swalchemist:
Removed an ignore clause that doesn't have any effect, and commented on another with a relevant issue link.

[#186765938]

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -27,9 +27,6 @@ updates:
   ignore:
   - dependency-name: "org.apache.tomcat:*"
     update-types: ["version-update:semver-major"] # Stay in Tomcat 9 because it is still supported, likely until 2027 (https://endoflife.date/tomcat), and bumping to Tomcat 10 requires some major code changes
-  - dependency-name: org.apache.directory.api:api-ldap-model
-    versions:
-    - ">= 3"
   - dependency-name: org.flywaydb:flyway-core
     versions:
     - ">= 8" # Version 8+ no longer supports MySQL 5.7, we're not ready to drop MySQL 5.7 support yet. Tracker story: https://www.pivotaltracker.com/story/show/180879913.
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -27,15 +27,12 @@ updates:
   ignore:
   - dependency-name: "org.apache.tomcat:*"
     update-types: ["version-update:semver-major"] # Stay in Tomcat 9 because it is still supported, likely until 2027 (https://endoflife.date/tomcat), and bumping to Tomcat 10 requires some major code changes
-  - dependency-name: org.apache.directory.api:api-ldap-model
-    versions:
-    - ">= 3"
```
strehle, line 32: I remember that this was added after problem with major upgrade from 1 to 2.x so it simply should ignore further major updates but we can remove it

.github/dependabot.yml
```diff
@@ -27,15 +27,12 @@ updates:
...
   - dependency-name: org.flywaydb:flyway-core
     versions:
     - ">= 8" # Version 8+ no longer supports MySQL 5.7, we're not ready to drop MySQL 5.7 support yet. Tracker story: https://www.pivotaltracker.com/story/show/180879913.
   - dependency-name: com.icegreen:greenmail
     versions:
-    - ">= 2" # Version 2+ was causing test flakiness, wait for it to stabilize. See https://github.com/cloudfoundry/uaa/pull/2314.
+    - ">= 2" # Version 2+ was causing test flakiness, wait for it to stabilize. See https://github.com/cloudfoundry/uaa/pull/2314. Story: https://www.pivotaltracker.com/story/show/186803724.
```
strehle: many others (including myself) dont have access to pivotal tracker stories, so it maybe confuses more than it helps

btw. after using spring boot 3 we must update this here as well and then we can remove it

swalchemist: I opened a GitHub issue and pointed to that instead. 

(Of course, the issue also points to Tracker.)

.github/dependabot.yml
swalchemist: Okay, let's err on the side of experimenting and see if the next several bumps work ok. We have a monthly process to review these exceptions, so I was trying to either document the reasons that we could continue to review, or get rid of them.

strehle (approved):

.github/dependabot.yml
strehle: ok 👌

PR #2676, opened 2024-01-11
bump go to 1.21.6

swalchemist:
Routine bump in the minimum Go version.

[#186765938]

### DIFF
--- a/k8s/go.mod
+++ b/k8s/go.mod
@@ -1,6 +1,6 @@
 module github.com/cloudfoundry/uaa
 
-go 1.21
+go 1.21.6
 
 require (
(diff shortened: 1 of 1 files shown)


peterhaochen47:
There's a check fail: https://github.com/cloudfoundry/uaa/actions/runs/7483273784/job/20368260650?pr=2676
Similar error message as https://github.com/golang/go/issues/61888
Seems like the solution is that the go installation in the environment that ran this failed check needs to be upgraded.

strehle:
see https://github.com/cloudfoundry/uaa/pull/2677

Tallicia:
This has been in fixed here: https://github.com/cloudfoundry/uaa/pull/2677

PR #2679, opened 2024-01-12
Post jwt token library refactoring

strehle:
- remove string serialization and simplify token creation. create a JWT from a map instead of a string
- reduce complexity in token creation (map to string, then string to JWT)
- reduce complexity in token read (only once)


model/src/test/java/org/cloudfoundry/identity/uaa/oauth/token/IntrospectionClaimsTest.java
```diff
@@ -36,5 +37,6 @@ void isActive() {
   @Test
   void testSerialize() {
     assertTrue(JsonUtils.writeValueAsString(INTROSPECTION_PAYLOAD).contains(TokenConstants.CLIENT_AUTH_NONE));
+    assertNotNull(INTROSPECTION_PAYLOAD.getClaimObject());
```
adrianhoelzl-sap: Can we here also check for some of the contents of the Map?

model/src/main/java/org/cloudfoundry/identity/uaa/oauth/token/Claims.java
```diff
@@ -363,4 +366,9 @@ public String getClientAuth() {
     public void setClientAuth(final String clientAuth) {
         this.clientAuth = clientAuth;
     }
+
+    @JsonIgnore
+    public Map<String, Object> getClaimObject() {
```
hsinn0: This method name does not make much sense. Different names I can think of: `getClaimValues()` `getClaimMap()`, or `obtainClaimsAsMap()`?

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/openid/IdToken.java
```diff
@@ -156,4 +157,9 @@ public Long getAuthTimeInSeconds() {
     public String userId() {
         return sub;
     }
+
+    @JsonIgnore
+    public Map<String, Object> getClaimObject() {
```
hsinn0: The same as in `Claims` class. I think that the method name can be improved.

hsinn0 (approved):

PR #2688, opened 2024-01-22
fix: check origin of user

### LINKED ISSUE #2576: SCIM user creation works when providing invalid origin

76.21.0 , issue is also present in older versions

Via bosh release. However, issue is independent of deployment

### What did you do?

Create or update a shadow user with invalid origin.

```
curl -X POST 'https://<uaaurl>/Users' -H "Authorization: bearer <admintoken>" -H "Content-type: application/json" -d '{"userName": "user-does-not-exist",  "origin": "does-not-exist", "emails": [{"value": [test@email.com](mailto:test@email.com), "primary": false}]}'
```

Creation or update is rejected because the origin does not map to a known identity provider.

### What did you see instead?

User is created or updated.

klaus-sap:
issue: https://github.com/cloudfoundry/uaa/issues/2576
check must be switched on in idz configuration to avoid incompatible change

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
@@ -588,7 +588,8 @@ public class JdbcScimUserProvisioning extends AbstractQueryable<ScimUser>
 
     private void checkOrigin(String origin, String zoneId) {
         Integer count = jdbcTemplate.queryForObject("select count(*) from identity_provider where origin_key=? and identity_zone_id=? ", Integer.class, origin, zoneId);
-        if (count == 0) {
+        int c = (count == null) ? 0 : count;
+        if (c == 0) {
             throw new InvalidScimResourceException("Invalid origin " + origin + " in Identity Zone " + zoneId);
         }
     }
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
@@ -62,6 +62,8 @@ import java.util.Map;
 import java.util.UUID;
 import java.util.regex.Pattern;
 
+import javax.annotation.Nullable;
+
 import static java.sql.Types.VARCHAR;
 import static org.springframework.util.StringUtils.hasText;
 
@@ -192,11 +194,12 @@ public class JdbcScimUserProvisioning extends AbstractQueryable<ScimUser>
 
     @Override
     public ScimUser create(final ScimUser user, String zoneId) {
-        validateUserLimit(zoneId);
+        UserConfig userConfig = getUserConfig(zoneId);
+        validateUserLimit(zoneId, userConfig);
         if (!hasText(user.getOrigin())) {
             user.setOrigin(OriginKeys.UAA);
         }
-        if (isCheckOriginEnabled(zoneId)) {
+        if (isCheckOriginEnabled(userConfig)) {
             checkOrigin(user.getOrigin(), zoneId);
         }
         logger.debug("Creating new user: {}", UaaStringUtils.getCleanedUserControlString(user.getUserName()));
@@ -548,23 +551,22 @@ public class JdbcScimUserProvisioning extends AbstractQueryable<ScimUser>
         jdbcTemplate.update(UPDATE_LAST_LOGON_TIME_SQL, timeService.getCurrentTimeMillis(), id, zoneId);
     }
 
-    private int getAllowedUserLimit(String zoneId) {
+    @Nullable
+    private UserConfig getUserConfig(String zoneId) {
         try {
-            UserConfig userConfig;
             IdentityZone currentZone = identityZoneManager.getCurrentIdentityZone();
-            userConfig = (currentZone.getId().equals(zoneId)) ?
+            return (currentZone.getId().equals(zoneId)) ?
                 currentZone.getConfig().getUserConfig() :
                 jdbcIdentityZoneProvisioning.retrieve(zoneId).getConfig().getUserConfig();
-            return userConfig.getMaxUsers();
         } catch (ZoneDoesNotExistsException e) {
             logger.debug("could not retrieve identity zone with id: {}", zoneId);
         }
-        return -1;
+        return null;
     }
 
-    private void validateUserLimit(String zoneId) {
+    private void validateUserLimit(String zoneId, UserConfig userConfig) {
         // get current limit of allowed users
-        long maxAllowedUsers = getAllowedUserLimit(zoneId);
+        long maxAllowedUsers = (userConfig == null) ? -1 : userConfig.getMaxUsers();
         // check, if there is a limit (>0), that the limit is not reached with one user more (getUsersCountForZone + 1)
         if (maxAllowedUsers > 0 && maxAllowedUsers < (getUsersCountForZone(zoneId) + 1)) {
             throw new InvalidScimResourceException("The maximum allowed numbers of users: " + maxAllowedUsers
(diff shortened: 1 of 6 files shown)


server/src/test/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioningTests.java
```diff
@@ -1048,26 +1053,66 @@ void updateLastLogonTime() {
...
-        for (int i = 2; i < 12; i++) {
+        try {
+            idzManager.getCurrentIdentityZone().getConfig().getUserConfig().setMaxUsers(10);
+            for (int i = 1; i < 12; i++) {
+                scimUser.setId("user-id-" + i);
+                scimUser.setUserName("user" +i+ "@example.com");
+                scimUser.setPassword(randomString());
+                scimUser = jdbcScimUserProvisioning.create(scimUser, IdentityZoneHolder.getCurrentZoneId());
+            }
+        } catch (InvalidScimResourceException e) {
```
adrianhoelzl-sap: Either use `assertThrows` or add `fail` before this line

server/src/test/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioningTests.java
```diff
@@ -1048,26 +1053,66 @@ void updateLastLogonTime() {
...
+        ScimUser scimUser = new ScimUser("user-id-"+randomId, "user@example.com", "User", "Example");
+        ScimUser.Email email = new ScimUser.Email();
+        email.setValue("user-"+randomId+"@example.com");
+        scimUser.setEmails(Collections.singletonList(email));
+        scimUser.setPassword(randomString());
+        scimUser.setOrigin(invalidOrigin);
+        try {
+            idzManager.getCurrentIdentityZone().getConfig().getUserConfig().setCheckOriginEnabled(true);
+            scimUser = jdbcScimUserProvisioning.create(scimUser, IdentityZoneHolder.getCurrentZoneId());
+        } catch (InvalidScimResourceException e) {
```
adrianhoelzl-sap: Same here

klaus-sap: fixed

server/src/test/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioningTests.java
klaus-sap: fixed

strehle:
see sonar issues, https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2688

strehle (commented):
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2688

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
```diff
@@ -271,6 +274,9 @@ public ScimUser update(final String id, final ScimUser user, String zoneId) thro
         logger.debug("Updating user " + user.getUserName());
         final String origin = hasText(user.getOrigin()) ? user.getOrigin() : OriginKeys.UAA;
         user.setOrigin(origin);
+        if (isCheckOriginEnabled(zoneId)) {
+            checkOrigin(origin, zoneId);
```
strehle, line 281: sonar does not show coverage for this here, can you please add a test and I would prefere an IT for negative testing as well

klaus-sap: There is already a negative test for create: cannotCreateUserWithInvalidOrigin

klaus-sap: Added test cannotUpdateUserWithInvalidOrigin

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
```diff
@@ -588,7 +588,8 @@ private boolean isCheckOriginEnabled(String zoneId) {
 
     private void checkOrigin(String origin, String zoneId) {
         Integer count = jdbcTemplate.queryForObject("select count(*) from identity_provider where origin_key=? and identity_zone_id=? ", Integer.class, origin, zoneId);
-        if (count == 0) {
+        int c = (count == null) ? 0 : count;
```
hsinn0: It seems that the query string is existing code. This query operation can be optimized by using `exists` instead of `count` though, because we just need to know whether there is at least 1 such record that meets the condition or not. For example, in mysql, the query can be as follows instead:

`select exists (select * from identity_provider where origin_key=? and identity_zone_id=?)`

And then the result can be assigned to a `boolean` variable, removing the need to check for null Integer value.

hsinn0 (dismissed):

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
klaus-sap: In this case, count can only be 0 or 1, and there is a matching index on table identity_provider.
https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/resources/org/cloudfoundry/identity/uaa/db/postgresql/V2_0_0__Multitenancy.sql#L37
So I don't think that the query can be optimized with respect to runtime.

hsinn0: @klaus-sap , do you not agree that exists query is more efficient than count query? Even if the count can be only 1 or 0, count query needs to go through all records while exists query does not, right?

hsinn0: I see. There is unique index on `identity_provider (identity_zone_id,origin_key)`. So in theory, not much to gain on the SQL side then. In that case, the only benefit of changing the query is likely not having to covert from Integer to int, and I am OK if you don't think that is important enough to make further change.

hsinn0 (approved):

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
```diff
@@ -196,6 +196,9 @@ public ScimUser create(final ScimUser user, String zoneId) {
         if (!hasText(user.getOrigin())) {
             user.setOrigin(OriginKeys.UAA);
         }
+        if (isCheckOriginEnabled(zoneId)) {
```
strehle: This here is in createUser, where we have some lines before this call validateUserLimit , 
Both potentially perform a SQL to resolve the identityzone. Can we combine it and do the  identityZoneManager.getCurrentIdentityZone check once?

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
```diff
@@ -565,4 +571,26 @@ private void validateUserLimit(String zoneId) {
...
+            return userConfig.isCheckOriginEnabled();
+        } catch (ZoneDoesNotExistsException e) {
+            logger.debug("could not retrieve identity zone with id: {}", zoneId);
+        }
+        return false;
+    }
+
+    private void checkOrigin(String origin, String zoneId) {
+        Integer count = jdbcTemplate.queryForObject("select count(*) from identity_provider where origin_key=? and identity_zone_id=? ", Integer.class, origin, zoneId);
+        int c = (count == null) ? 0 : count;
```
strehle: line 591 + 592 -> if (count == null || count == 0) { throw xxx

klaus-sap: fixed with commit https://github.com/cloudfoundry/uaa/pull/2688/commits/e382e49e6a53b04441332d272412036aee4e9260

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
klaus-sap: fixed with [d9f8f25](https://github.com/cloudfoundry/uaa/pull/2688/commits/d9f8f25bfb0d791fcb5dfd8ea1d5c1e432c72428)

strehle (approved):

PR #2689, opened 2024-01-24
Import refactor for SAML

swalchemist:
These refactors will prepare the code for removing the old Spring SAML extension library. There are several libraries that are brought in only as a result of having the SAML extension library. These changes decouple these libraries that aren't directly related to SAML, by declaring them explicitly.

Changes made with the help of @peterhaochen47.

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -129,6 +129,7 @@ libraries.nimbusJwt = "com.nimbusds:nimbus-jose-jwt:9.37.3"
 libraries.xmlSecurity = "org.apache.santuario:xmlsec:4.0.1"
 libraries.orgJson = "org.json:json:20231013"
 libraries.spingSamlEsapiDependencyVersion = "org.owasp.esapi:esapi:2.5.3.1"
+libraries.jodaTime = "joda-time:joda-time:2.12.6"
 
 // gradle plugins
 libraries.testRetryPlugin = "org.gradle:test-retry-gradle-plugin:1.5.8"
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -128,7 +128,7 @@ libraries.zxing = "com.google.zxing:javase:3.5.2"
 libraries.nimbusJwt = "com.nimbusds:nimbus-jose-jwt:9.37.3"
 libraries.xmlSecurity = "org.apache.santuario:xmlsec:4.0.1"
 libraries.orgJson = "org.json:json:20231013"
-libraries.spingSamlEsapiDependencyVersion = "org.owasp.esapi:esapi:2.5.3.1"
+libraries.owaspEsapi = "org.owasp.esapi:esapi:2.5.3.1"
 libraries.jodaTime = "joda-time:joda-time:2.12.6"
 libraries.commonsHttpClient = "commons-httpclient:commons-httpclient:3.1"
 
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -130,6 +130,7 @@ libraries.xmlSecurity = "org.apache.santuario:xmlsec:4.0.1"
 libraries.orgJson = "org.json:json:20231013"
 libraries.spingSamlEsapiDependencyVersion = "org.owasp.esapi:esapi:2.5.3.1"
 libraries.jodaTime = "joda-time:joda-time:2.12.6"
+libraries.commonsHttpClient = "commons-httpclient:commons-httpclient:3.1"
 
 // gradle plugins
 libraries.testRetryPlugin = "org.gradle:test-retry-gradle-plugin:1.5.8"
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -31,7 +31,7 @@ dependencies {
         exclude(module: "xalan")
     }
     implementation(libraries.jodaTime)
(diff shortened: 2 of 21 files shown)


dependencies.gradle
```diff
@@ -128,7 +128,9 @@ libraries.zxing = "com.google.zxing:javase:3.5.2"
 libraries.nimbusJwt = "com.nimbusds:nimbus-jose-jwt:9.37.3"
 libraries.xmlSecurity = "org.apache.santuario:xmlsec:4.0.1"
 libraries.orgJson = "org.json:json:20231013"
-libraries.spingSamlEsapiDependencyVersion = "org.owasp.esapi:esapi:2.5.3.1"
+libraries.owaspEsapi = "org.owasp.esapi:esapi:2.5.3.1"
```
strehle, line 131: Can you please add comments to these libraries... because httpClient is not needed finally so we need these libraries only for now

strehle (approved):
OK, but some comments

dependencies.gradle
swalchemist: So you're asking about `commons-httpclient:commons-httpclient:3.1` not esapi?

While the develop branch works fine without this explicit dependency, on our [branch that has the SAML extension library removed](new-saml-lib-186822654), the code doesn't compile without this httpclient dependency. 

Specifically, it's needed in `ExternalOAuthAuthenticationFilter.java`

> import org.apache.commons.httpclient.util.URIUtil;
...
>     final String origin = URIUtil.getName(String.valueOf(request.getRequestURL()));

And in `TokenMvcMockTests.java` in the invalidScopeErrorMessageIsNotShowingAllClientScopes and 
invalidScopeErrorMessageIsNotShowingAllUserScopes test methods, for the same import.

So unless we decide to change this code to use a different library, the need to declare this dependency is not temporary.

Cc: @peterhaochen47

strehle: > So you're asking about `commons-httpclient:commons-httpclient:3.1` not esapi?

Yes httpclient:3.1 is a very only version of apache http and this wont be needed if we have no opensaml2, so I recommend this to mark with a todo or comment to remove it later...
In our CVE scanners this library is excluded for many years now ... 
For esapi I am not sure if newer opensaml4 has it or not

swalchemist: Our intention was to declare the dependency that the develop branch was already using for this import. I'm trying to double-check what that is.

strehle: I know, but this will be obsolete ... however for me it is Ok as it is,

swalchemist: In that case, isn't it already obsolete? Are you saying this is already a problem on develop? We're not trying to change the version.

strehle: no , only later if we move away from opensaml2 -> opensaml3 or 4 then the apache http client 3 is no longer needed.

peterhaochen47: > if we move away from opensaml2 -> opensaml3 or 4 then the apache http client 3 is no longer needed

The particular usage of apache http client (see [usage here](https://github.com/cloudfoundry/uaa/blob/aaeb747e26887a69ede6864ccbdc9bfcb5cba60a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationFilter.java#L95)) is NOT related to saml or used in the context of opensaml. So we could stop using this apache http client in the future if it's problematic, but it seems to not dependent on any saml change.

strehle: maybe you check our SBOM we have apache http client in 2 different versions , httpclient-4.5.14 .... needed from many components, and then commons-httpclient-3.1 which is needed only for opensaml2 for automatic SAML metadata update

strehle: > ent (see [usage here](https://github.com/cloudfoundry/uaa/blob/aaeb747e26887a69ede6864ccbdc9bfcb5cba60a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationFilter.java#L95))

the usage in our code is something we should refactor.... 
but the other usage cannot be refactored without opensaml upgrade, e.g.
https://github.com/search?q=repo%3Acloudfoundry%2Fuaa%20org.apache.commons.httpclient&type=code

swalchemist: We looked in httpclient 4.5.14, and it doesn't seem to have the same API, and we didn't see a getName() method there at all, which is what we're using from httpclient 3.1. Do you see one there?

strehle: https://mvnrepository.com/artifact/commons-httpclient/commons-httpclient
vs.
https://mvnrepository.com/artifact/org.apache.httpcomponents/httpclient

But we can close this discussion and you can proceed....

swalchemist: I think in this particular case, our best option is to migrate to using java.net.URI to replace URIUtil.getName, which then removes the need for commons-httpclient:3.1. I tried to do that earlier and found that handling the checked exceptions that the library throws was a pain.

https://marc.info/?l=httpclient-users&m=125425095705062&w=2

peterhaochen47: We are going to plan to merge this PR. Created this issue to track the future work to replace the old httpclient lib: https://github.com/cloudfoundry/uaa/issues/2691

PR #2693, opened 2024-01-27
Refactor BouncyCastleProvider to BouncyCastleFipsProvider

### LINKED ISSUE #2230: Support FIPS compliant BC
Support https://www.bouncycastle.org/fips-java/ instead of only https://www.bouncycastle.org/java.html

CF seems to provide FIPS containers, e.g. https://github.com/cloudfoundry/uaa-release/issues/358 but then UAA should support FIPS in runtime

strehle:
Solves issue https://github.com/cloudfoundry/uaa/issues/2230

Replaced BC libraries towards https://www.bouncycastle.org/fips-java/, e.g. 
https://www.bouncycastle.org/fips_faq.html

Some refactoring's were needed, used the BC cookbook, e.g. 
https://downloads.bouncycastle.org/fips-java/BC-FJA-UserGuide-1.0.2.pdf


peterhaochen47:
Thanks & makes sense! Just to confirm, this change wouldn't introduce any breaking changes right (e.g. like changes in any publicly exposed keys like the token keys that would impact UAA clients)?

strehle:
> Thanks & makes sense! Just to confirm, this change wouldn't introduce any breaking changes right (e.g. like changes in any publicly exposed keys like the token keys that would impact UAA clients)?

correct, no breaking change.

peterhaochen47 (approved):

PR #2695, opened 2024-01-30
exclude vulnerable indirect dep xalan

peterhaochen47:
- a vulnerable xalan version (2.7.2; CVE-2022-34169) is brought in by the saml library we use: org.springframework.security.extensions:spring-security-saml2-core which has reached EOL (we are replacing it for develop branch, but not 74.5.x branch). So this xalan would not be bumped by spring-security-saml2-core anymore.
- so to address this CVE scan result (CVE-2022-34169), exclude this indirect dep, like develop branch does: https://github.com/cloudfoundry/uaa/commit/061bee94cc3bc0e4d40ade7dc2d8949af46e9274
- note: excluding a dep of a library we use carries the risk of failure if we trigger a code path in the library that depends on the dep. The fact all tests are passing + the fact that this exclude has been applied in develop branch for more than 1 year give us enough confidence that this exclude would not introduce a failure.

[#186948853]

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -27,6 +27,7 @@ dependencies {
     }
     implementation(libraries.springSecuritySaml) {
         exclude(module: "bcprov-ext-jdk15on")
+        exclude(module: "xalan") // exclude vulnerable xalan 2.7.2 (see: CVE-2022-34169)
     }
     implementation(libraries.springSessionJdbc)
 
--- a/uaa/build.gradle
+++ b/uaa/build.gradle
@@ -80,7 +80,9 @@ dependencies {
     testImplementation(libraries.springTest)
     testImplementation(libraries.springSecurityJwt)
     testImplementation(libraries.springSecurityLdap)
-    testImplementation(libraries.springSecuritySaml)
+    testImplementation(libraries.springSecuritySaml) {
+        exclude(module: "xalan") // exclude vulnerable xalan 2.7.2 (see: CVE-2022-34169)
+    }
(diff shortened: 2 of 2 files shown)


swalchemist (commented):
Would it makes sense to add comments in the build.gradle files explaining why the exclude is there? Seems important since it's keeping a CVE out of the package.

swalchemist:
Wouldn't the build fail if we actually needed this dependency? Seems pretty low-risk to me.

I know Spring has a lot of potential for runtime errors that are surprising, but perhaps not for this type of issue.

peterhaochen47:
> Wouldn't the build fail if we actually needed this dependency? Seems pretty low-risk to me.
> 
> I know Spring has a lot of potential for runtime errors that are surprising, but perhaps not for this type of issue.

After reading this [gradle doc](https://docs.gradle.org/current/userguide/dependency_downgrade_and_exclude.html#sec:excluding-transitive-deps), I got the impression that the failure might happen at runtime instead of build time. But in any case, if all our tests are passing, it's pretty good evidence that exclude works. Usually, we should use exclude very sparingly, in this case the direct dep (the saml extension lib) has reached EOL, so it can't be bumped, so our options are limited.

peterhaochen47:
> Would it makes sense to add comments in the build.gradle files explaining why the exclude is there? Seems important since it's keeping a CVE out of the package.

Sure, wouldn't hurt to add that in addition to commit message.

swalchemist (approved):

PR #2700, opened 2024-01-31
Refactor saml dependencies 186822654

bruce-ricard:
We have a couple refactors that will be helpful when replacing the SAML extension library.

They should also help with readability and complexity, in general.

Also added a couple tests, when we found missing coverage.

@swalchemist and me


strehle:
sonar is not happy with it, therefore I have to check in more details 
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2700

bruce-ricard:
Thanks for pointing that out Markus. We'll look into it.

swalchemist:
Why is the Sonar action passing when SonarCloud says it failed?

swalchemist:
I have pushed fixes for some of the code smells that Sonar found. However, the Sonar scan still fails because of its code coverage rules. The code we added to PasscodeInformation.java is 100% covered, but Sonar wants us to add coverage for the code that we didn't change, perhaps because we moved it to a new package.

We could possibly improve coverage by removing an unused constructor, but we'd need to figure out what to do with the Jackson annotations on that constructor.

strehle (approved):

PR #2638, opened 2023-12-06
Remove SAML IDP feature

hsinn0:
- In preparation for the SAML library replacement

~~Approve or disapprove but do not merge to develop yet.~~ We want merge this to `develop` now. As this is a breaking change, ~~we have to decide which UAA version this change should go with.~~ we also want to bump UAA major version after the merge.

PR #2701, opened 2024-02-01
feat: clean up unused DB table service_provider

peterhaochen47:
- This table was added for the UAA-as-SAML-IDP feature (https://github.com/cloudfoundry/uaa/commit/b93c87aa210097539a5778448cec5ca281b5de17)
- This feature has been removed: https://github.com/cloudfoundry/uaa/pull/2638. Hence this table is now unused.
- The "DROP TABLE IF EXISTS" syntax would not error out if the table does not exist, compared to just "DROP TABLE".
- Also clean up docs and a test util that reference this table.

[#182118433]

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,7 +197,6 @@ List of relations
  public | schema_version                | table    | root
  public | sec_audit                     | table    | root
  public | sec_audit_id_seq              | sequence | root
- public | service_provider              | table    | root
  public | spring_session                | table    | root
  public | spring_session_attributes     | table    | root
  public | user_google_mfa_credentials   | table    | root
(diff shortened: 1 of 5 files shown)


Tallicia (approved):
Looks good

Tallicia:
@strehle FYI - I believe this is expected to be included before the v77 release as part of the breaking changes for the SAML IdP removal.

strehle:
@peterhaochen47 I have seen this also during review of https://github.com/cloudfoundry/uaa/pull/2638 but I thought you wanted wait with the removal.

My assumption was that maybe we wait an additional release until removing the DB table because if we remove data we cannot revert to the older version.
Assume: someone does a CF update but then realizes that SAML IDP is not available anymore. If they then go back to the UAA before they can run without disruption. If we remove the DB table we remove this possiblity.

Therefore I thought it was wanted to not removing the table, however from my side we can also remove it and then release or 
release now 77.v

strehle (approved):
OK for SAP, but see my comments about this step.... if we delete the table, then the way back is not possible

hsinn0:
I am also for not removing the table right away but waiting for a couple of releases or so.

peterhaochen47:
Given that this PR has 2 approvals (and that I just discussed this with my team and the team was mostly neutral but leaning toward merging), so I will opt to merge this & include it in UAA v77.0.0.

PR #2702, opened 2024-02-01
Fix filter check in ids/Users endpoint

### LINKED ISSUE #2703: Potential Error in the SCIM Filter Validation of the /ids/Users Endpoint
The /ids/Users endpoint allows passing a SCIM filter as a URL parameter. For validating this filter string, the value is parsed and subsequently traversed recursively, checking each clause for validity.

When handling clauses joined by the operator `or` or `and`, we saw that an OR-operator is used for combining the validation result of the left-hand side and the right-hand-side of the expression (see https://github.com/cloudfoundry/uaa/blob/d962dd67cc08de7d43aeef9cf0ae92f8c4b5cb85/server/src/main/java/org/cloudfoundry/identity/uaa/scim/endpoints/UserIdConversionEndpoints.java#L123).

If the RHS of the expression is valid (i.e., `checkFilter` returned `true` for it) while the LHS is invalid (i.e., `checkFilter` returned `false` for it), the method will return `true` for the joined expression.

Instead, the two clauses should be combined by using an AND operator, so that both clauses are required to be valid.

If the current state of the code is however correct, this shall be made clear in the comments of the code.

adrianhoelzl-sap:
Fix an inconsistency in the SCIM filter check used in the `/ids/Users` endpoint where only one of the operands of an `and` or `or` clause must be valid.

As indicated in the commit https://github.com/cloudfoundry/uaa/commit/0c51eab31bac7fee03ddc1b659090b6999253d44, the or operator is actually correct since the method should check if ANY of the clauses in the SCIM filter contains a comparison with the "id" or "userName" property.


strehle (dismissed):
thanks, LGTM

Tallicia (approved):
This looks good, nice clear function renaming and behavior

PR #2704, opened 2024-02-02
Optimize "/ids/Users" endpoint

### LINKED ISSUE #2705: Performance Issues in SCIM User Lookup of "/ids/Users" Endpoint
The endpoint "/ids/Users" allows filtering the users of an identity zone by ID and/or username by using a SCIM filter. Furthermore, it (optionally) allows restricting the result to only those users that come from an active IdP.

In the endpoint handler, this filter for active origins is implemented by constructing an additional clause in the SCIM filter: first, all active IdPs in the current identity zone are read. Second, for each of their origins, the clause `origin eq <origin key>` is added to the SCIM filter (combined by using the `or` operator).

This modified SCIM filter is then passed to the endpoint handler of the regular SCIM users endpoint, where an SQL query is generated from it. Analogous to the SCIM filter, this query will contain a `WHERE` clause containing one `origin = <origin key>` condition for each origin of an active IdP.

Especially for large number of active IdPs in an identity zone, the execution of this query will become less performant.

Due to the nature of the original SCIM filter, i.e., only allowing the username and/or the ID of the user, there will be relatively few matches. Therefore, the following approach is more efficient:

First, fetch users by applying the initial filter (i.e., the one containing only conditions regarding the user's ID and/or username). Then, if necessary (i.e., if only users from active IdPs shall be returned), these IdPs are to be read and the users list finally filtered for the origins of the active IdPs in Java code.

adrianhoelzl-sap:
see issue #2705 

Prerequisite: PR #2702


peterhaochen47:
> see issue #2705
> 
> Prerequisite: PR #2703

Hi @adrianhoelzl-sap, #2703 is an issue, did you mean PR #2702?

adrianhoelzl-sap:
> > see issue #2705
> > Prerequisite: PR #2703
> 
> Hi @adrianhoelzl-sap, #2703 is an issue, did you mean PR #2702?

You are right, I updated it. Thanks :)

bruce-ricard:
The first commit 955cd2b02ab2f9de4ac734df2b460b388015437e  is already on `develop` . Should we rebase the PR branch?

bruce-ricard:
Also, github is saying

> This branch has conflicts that must be resolved

bruce-ricard (dismissed):
Rebase and merge conficts need to be resolved.

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
```diff
@@ -170,6 +176,79 @@ public List<ScimUser> retrieveByUsernameAndZone(String username, String zoneId)
         return jdbcTemplate.query(USER_BY_USERNAME_AND_ZONE_QUERY , mapper, username, zoneId);
     }
 
+    /**
+     * This method should behave similar to {@link JdbcScimUserProvisioning#query(String, String, boolean, String)},
+     * with the filtering for active IdPs (if required) being the only difference.
+     */
+    @Override
+    public List<ScimUser> retrieveByScimFilter(
```
swalchemist: Is there a refactor you can do to reduce the number of arguments here to 3 or 4? For example, in the implementation, the path for includeInactive true and false and completely independent, so it would make sense to make those two different methods.

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
```diff
@@ -170,6 +176,79 @@ public List<ScimUser> retrieveByUsernameAndZone(String username, String zoneId)
...
+            final String filter,
+            final boolean includeInactive,
+            final String sortBy,
```
swalchemist, line 182: Is there a better type than String that we can use for sortBy so it's better constrained to valid values?

swalchemist (changes requested):
I'm partway through the PR.

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
adrianhoelzl-sap: Fixed with commit 709a372

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
adrianhoelzl-sap: The method allows passing any of the fields of the `ScimUser` class for this parameter (except for `password` and `salt`). We can introduce an enum for all the options, but I would do this in another issue/PR. What do you think?

swalchemist: Yeah, it looks like other existing code is already depending on this free string, so sure, let's save that improvement for another PR.

server/src/test/java/org/cloudfoundry/identity/uaa/scim/endpoints/UserIdConversionEndpointsTests.java
```diff
@@ -96,46 +120,78 @@ public void testBadFilterWithGroup() {
...
+        // check different page sizes
+        assertEndpointReturnsCorrectResult(filter, 1, expectedUsers, false);
+        assertEndpointReturnsCorrectResult(filter, 2, expectedUsers, false);
+        assertEndpointReturnsCorrectResult(filter, 3, expectedUsers, false);
+        assertEndpointReturnsCorrectResult(filter, 4, expectedUsers, false);
+        assertEndpointReturnsCorrectResult(filter, 10, expectedUsers, false);
+    }
+
     @Test
     public void testGoodFilter1() {
```
swalchemist: Could we get more descriptive test method names instead of numbers on all of these?

server/src/test/java/org/cloudfoundry/identity/uaa/scim/endpoints/UserIdConversionEndpointsTests.java
```diff
@@ -222,16 +281,93 @@ public void testBadFilter11() {
...
+
+    private void arrangeCurrentIdentityZone(final String idzId) {
+        final IdentityZone identityZone = new IdentityZone();
+        identityZone.setId(idzId);
+        idzHolderMockedStatic.when(IdentityZoneHolder::get).thenReturn(identityZone);
+        idzHolderMockedStatic.when(IdentityZoneHolder::getCurrentZoneId).thenReturn(idzId);
+        when(identityZoneManager.getCurrentIdentityZoneId()).thenReturn(idzId);
+    }
+
+    private void arrangeScimUsersForFilter(
```
swalchemist, line 227: This method is really complex. Is it really necessary to validate all of these details for all of the tests that call this?

server/src/test/java/org/cloudfoundry/identity/uaa/scim/endpoints/UserIdConversionEndpointsTests.java
```diff
@@ -13,65 +13,89 @@
...
+    private final IdentityZoneManager identityZoneManager = mock(IdentityZoneManager.class);
 
     @SuppressWarnings("rawtypes")
-    private Collection authorities = AuthorityUtils
-                    .commaSeparatedStringToAuthorityList("orgs.foo,uaa.user");
+    private final Collection authorities = AuthorityUtils
+            .commaSeparatedStringToAuthorityList("orgs.foo,uaa.user");
+
+    public UserIdConversionEndpointsTests() {
+        this.idzHolderMockedStatic = mockStatic(IdentityZoneHolder.class);
```
swalchemist: Sad that we have to use mockStatic here. But I'll trust that this won't cause any additional flakiness in the test results.

adrianhoelzl-sap: I was able to get rid of this mock entirely (commit 2644f53).

server/src/test/java/org/cloudfoundry/identity/uaa/scim/endpoints/UserIdConversionEndpointsTests.java
adrianhoelzl-sap: I migrated the entire test class to JUnit 5. With that, we can use `ParameterizedTest`s to group the tests for the different success and error cases. (see commit 2644f53)

server/src/test/java/org/cloudfoundry/identity/uaa/scim/endpoints/UserIdConversionEndpointsTests.java
adrianhoelzl-sap: I assume that you were referring to the method below, i.e., `assertEndpointReturnsCorrectResult`. I simplified it a bit in commit d0cf81f.

swalchemist (approved):

Tallicia:
@adrianhoelzl-sap @strehle This PR is where I am very likely going to revert in the next half hour to hour as this looks to be the culprit for the performance drop.

Thoughts?

PR #2712, opened 2024-02-05
Inconsistent Update Behavior for SCIM "/Users/{userId}"

### LINKED ISSUE #2671: Inconsistent update behavior for SCIM /Users/{userId}

latest 

not relevant, but docker, cf deployment

### What did you do?

Update the origin of a specific useruser via SCIM API /Users/{userId} , either via PATCH or PUT.

Consistent behavior that both http methods either allow or disallow the update of the origin of a user via scim.

### What did you see instead?

- PUT allows update of the origin of a user
- PATCH does not allow the update of the origin of a user

```sh
# create a new user
curl -X POST '<uaadomain>/Users' -H "Authorization: bearer $TOKEN" -H "Content-type: application/json" -d '{"userName": "user-does-not-exist",  "origin": "does-not-exist", "emails": [{"value": "test@email.com", "primary": false}]}'

# try to update origin via patch, returns successfully but with old origin
curl -X PATCH '<uaadomain>/Users/<userid>' -H "Authorization: bearer $TOKEN" -H "Content-type: application/json" -H 'If-Match: *' -d '{"origin":"still-not-existing"}'

# update origin via put, returns successfully with new origin
curl -X PUT '<uaadomain>/Users/<userid>' -H "Authorization: bearer $TOKEN" -H "Content-type: application/json" -H 'If-Match: *' -d '{"userName": "user-does-not-exist",  "origin": "still-does-not-exist", "emails": [{"value": "test@email.com", "primary": false}], "name": {"familyName":"familyName", "givenName":"givenName"}}'

```

Why ?

PATCH does not take into consideration the origin while applying the provided delta.

https://github.com/cloudfoundry/uaa/blob/4dae99997cb27c3fea97a07ced3a2a2402f2e037/model/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUser.java#L711

PUT simply takes the provided user from the body and updates

https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/scim/endpoints/ScimUserEndpoints.java#L262

Current documentation mentions origin as update field for both methods.

Further, this suffers from the same issue as #2576 , i.e., i can PUT a non-existing origin.

adrianhoelzl-sap:
See Issue #2671


uaa/src/test/java/org/cloudfoundry/identity/uaa/scim/endpoints/ScimUserEndpointsMockMvcTests.java
```diff
@@ -1068,6 +1069,23 @@ void testUpdateUser_No_Username_Returns_400() throws Exception {
...
+
+        // change origin
+        final String initialOrigin = scimUser.getOrigin();
+        scimUser.setOrigin(UUID.randomUUID().toString());
+
+        scimUser.setName(new ScimUser.Name("Some New", "Name"));
+
+        updateUser(scimReadWriteToken, HttpStatus.BAD_REQUEST.value(), scimUser);
+
+        // should work again when origin was not changed
```
swalchemist: This last part doesn't seem necessary for the test, and should be covered by a different test.

server/src/test/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioningTests.java
```diff
@@ -1170,6 +1200,30 @@ private static void addUser(final JdbcTemplate jdbcTemplate,
         jdbcTemplate.execute(addUserSql);
     }
 
+    private static void addUser(final JdbcTemplate jdbcTemplate,
```
swalchemist: Could this signature be simplified with a parameter class, like ScimUser? It's a large number of parameters.

swalchemist (changes requested):
Am I reading this right, that PATCH will refuse to modify the origin with no error, but PUT will give an error if a different origin is given? That doesn't meet the need to give them the same behavior.

Also, will the API documentation be updated to indicate that the origin cannot be changed?

server/src/test/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioningTests.java
adrianhoelzl-sap: fixed with 0dec2a7

uaa/src/test/java/org/cloudfoundry/identity/uaa/scim/endpoints/ScimUserEndpointsMockMvcTests.java
adrianhoelzl-sap: fixed with 0dec2a7

swalchemist (approved):

swalchemist (approved):

uaa/src/test/java/org/cloudfoundry/identity/uaa/scim/endpoints/ScimUserEndpointsMockMvcTests.java
swalchemist, line 1: Much easier to read now. Thanks!

klaus-sap (approved):

PR #2718, opened 2024-02-08
feat: Use Database availability as indicator for /healthz response

tack-sap:
In one of our landscapes we recently saw, that the UAA /healthz endpoint will respond with ok even if the UAA Database is completely down (and hence none of the other requests to the UAA will work).

This PR aims to consider the DB availability in the response of the /healthz endpoint. 
As the requests to /healthz are time critical and should not take over a second, the actual DB check is done in a scheduled background task that is trying to establish a new DB connection and execute a statement. This task is executed every 10 seconds by default (time is configurable) and the /healthz endpoint will read the flag whether the last connection was successful.

When the DB connection works, the status is kept as it was before. When there has not been a DB check (usually only during startup), the status will still be 200, but the message will change until there has been a DB connection check.
A failed DB connection will result in a 503 status code along with an error message, indicating that the DB connection check failed.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/health/HealthzEndpoint.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/health/HealthzEndpoint.java
@@ -55,7 +55,7 @@ public class HealthzEndpoint {
             return "stopping\n";
         } else {
             if (wasLastConnectionSuccessful == null) {
-                return "No Check done.\n";
+                return "UAA running. Database status unknown.\n";
             }
 
             if (wasLastConnectionSuccessful) {
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/health/HealthzEndpoint.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/health/HealthzEndpoint.java
@@ -59,7 +59,7 @@ public class HealthzEndpoint {
             }
 
             if (wasLastConnectionSuccessful) {
-                return "ok\n";
+                return "ok. Database connection successful.\n";
             } else {
                 response.setStatus(503);
                 return "Database Connection failed.\n";
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/health/HealthzEndpoint.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/health/HealthzEndpoint.java
@@ -1,13 +1,18 @@
 package org.cloudfoundry.identity.uaa.health;
 
+import javax.servlet.http.HttpServletResponse;
+import javax.sql.DataSource;
+
+import java.sql.Connection;
+import java.sql.Statement;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.scheduling.annotation.Scheduled;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.ResponseBody;
-
-import javax.servlet.http.HttpServletResponse;
 
 /**
  * Simple controller that just returns "ok" in a request body for the purposes
@@ -18,10 +23,13 @@ import javax.servlet.http.HttpServletResponse;
 public class HealthzEndpoint {
     private static Logger logger = LoggerFactory.getLogger(HealthzEndpoint.class);
     private volatile boolean stopping = false;
+    private volatile Boolean wasLastConnectionSuccessful = null;
+    private DataSource dataSource;
 
     public HealthzEndpoint(
             @Value("${uaa.shutdown.sleep:10000}") final long sleepTime,
-            final Runtime runtime) {
+            final Runtime runtime,
+            final DataSource dataSource) {
         Thread shutdownHook = new Thread(() -> {
             stopping = true;
             logger.warn("Shutdown hook received, future requests to this endpoint will return 503");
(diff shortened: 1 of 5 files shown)


swalchemist (commented):
I tried sampling the healthz endpoint while starting up UAA connected with mysql. As the server came up, it took 33 seconds for the endpoint to respond. Is it safe to assume that's beyond what's considered normal?

```
% while true; do date; time curl localhost:8080/uaa/healthz; sleep 3; done
Thu Feb  8 17:52:43 PST 2024
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (7) Failed to connect to localhost port 8080 after 0 ms: Couldn't connect to server

real    0m0.017s
user    0m0.004s
sys     0m0.008s
Thu Feb  8 17:52:46 PST 2024
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100     3    0     3    0     0      0      0 --:--:--  0:00:33 --:--:--     0
ok

real    0m33.452s
user    0m0.007s
sys     0m0.012s
```

server/src/main/java/org/cloudfoundry/identity/uaa/health/HealthzEndpoint.java
```diff
@@ -45,8 +54,28 @@ public String getHealthz(HttpServletResponse response) {
             response.setStatus(503);
             return "stopping\n";
         } else {
-            return "ok\n";
+            if (wasLastConnectionSuccessful == null) {
+                return "No Check done.\n";
```
swalchemist: Would this message cause confusion? Maybe it should mention that UAA is up but the database status is unknown.

Also, should this return 503 instead of 200 for this case?

server/src/main/java/org/cloudfoundry/identity/uaa/health/HealthzEndpoint.java
```diff
@@ -45,8 +54,28 @@ public String getHealthz(HttpServletResponse response) {
...
+            } else {
+                response.setStatus(503);
+                return "Database Connection failed.\n";
+            }
         }
     }
 
-}
\ No newline at end of file
+    @Scheduled(fixedRateString = "${uaa.health.db.rate:10000}")
```
swalchemist, line 70: Do we know if the schedule will run this right at startup? I'm just curious whether we're likely to get "No Check done" in the 10 seconds after startup.

tack-sap: From my experience with this, the first run is done directly after the app is running (if there is no configuration of the "initialDelay" in the annotation. So the time where you would be able to see the No Check done should be rather small.

server/src/main/java/org/cloudfoundry/identity/uaa/health/HealthzEndpoint.java
tack-sap: I would prefer to keep 200 here, as otherwise you could theoretically be very unlucky and the UAA may be restarted because of the error from /healthz, while the check was simply not yet done. 

I updated the message according to your feedback.

server/src/main/java/org/cloudfoundry/identity/uaa/health/HealthzEndpoint.java
```diff
@@ -45,8 +54,28 @@ public String getHealthz(HttpServletResponse response) {
             response.setStatus(503);
             return "stopping\n";
         } else {
-            return "ok\n";
+            if (wasLastConnectionSuccessful == null) {
+                return "UAA running. Database status unknown.\n";
+            }
+
+            if (wasLastConnectionSuccessful) {
+                return "ok\n";
```
Tallicia: This should indicate database connection is successful, especially from a searching logs/responses.

Tallicia (changes requested):
Response on successful database connection state should be explicit

server/src/main/java/org/cloudfoundry/identity/uaa/health/HealthzEndpoint.java
tack-sap: Initially I left this message the same so anyone checking on the current message would not see a difference. 
But per your request I extended the message, mentioning the DB connection was successful.

swalchemist:
Looking better. Still curious about your thoughts on the long delay in response at startup.

tack-sap:
> Looking better. Still curious about your thoughts on the long delay in response at startup.

When starting the uaa locally I have the following behavior: 
- when the cargoRunLocal step begins there is no response yet for a few seconds (<5s, i suppose this is where the container is started)
- afterwards any request will wait for a response until the container becomes healthy (e.g. flyway finished work, all beans are created, all required spring magic is done)
- after about 30 seconds the request will return with a status, in my case it was already returning that the DB connection was successful (so the DB check already finished until my request came to the point where the result is checked)

Is this the same behavior for you or do you see any other behavior here?

From my perspective the above is the normal behavior, it has always been like this and this is not only like this for the healthz endpoint, but also if you e.g. want to access the login page during startup. i.e. this is nothing that is changed by this PR, the startup just takes some time where requests are accepted, but not yet processed. Also the 33s is about the amount of time that I would still see as normal.

swalchemist:
@tack-sap My apologies, I didn't think to test the current behavior on develop, and I do see that this delay is already how the healthz endpoint works.

swalchemist (approved):

Tallicia (approved):
Looks good.

PR #2719, opened 2024-02-09
refactor: reduce test dependency on EOL lib

peterhaochen47:
- instead of using org.springframework.security.saml.trust.X509TrustManager which is provided by a saml lib that has reached EOL, use a more standard/generic X509TrustManager.

[#186958651]


swalchemist (approved):

Tallicia (approved):
Looks good!

PR #2721, opened 2024-02-09
Maintain tomcat version

strehle:
As long as we have no spring boot upgrade


strehle:
Check the resulted libs from uaa bosh , e.g. https://github.com/cloudfoundry/uaa-release/releases/download/v77.0.0/uaa-77.0.0.tgz -> uaa -> cloudfoundry-identity-uaa.war -> web-inf/... 
... search for tomcat. Should have 9.0.85, but contain versions from 9.0.83 (last spring boot dependency)
tomcat-jdbc-9.0.83

PR #2731, opened 2024-02-15
Revert "Ignore failing integration test"

peterhaochen47:
This reverts commit 3f891393dd34591342dd4cbd477d819da4a9c030.

- this test is passing; we shouldn't ignore valid tests on develop branch (or in branches we intend to release)


swalchemist (approved):
LGTM. Also, the combination of `org.junit.Test` with `org.junit.jupiter.api.extension.ExtendWith` is a concern - mixing jUnit4 and jUnit5 libraries can be problematic.

PR #2737, opened 2024-02-20
Move IdP Alias Handling to separate Class

adrianhoelzl-sap:
see issue #2505 

Move the handling of alias IdPs to a separate class so that it can be re-used for the future alias feature for SCIM users.


server/src/main/java/org/cloudfoundry/identity/uaa/alias/EntityAliasHandler.java
```diff
@@ -0,0 +1,236 @@
...
+                        existingEntity.getAliasDescription()
+                ));
+            }
+
+            // both properties must be left unchanged in the operation
+            return existingEntity.getAliasId().equals(requestBody.getAliasId())
+                    && existingEntity.getAliasZid().equals(requestBody.getAliasZid());
+        }
+
+        // alias ID must not be set when no alias existed already
```
klaus-sap, line 57: caller must not choose alias ID, instead new alias ID will be generated

klaus-sap (approved):

model/src/main/java/org/cloudfoundry/identity/uaa/EntityWithAlias.java
```diff
@@ -0,0 +1,45 @@
...
+
+    String getZoneId();
+
+    @Nullable
+    String getAliasId();
+
+    void setAliasId(String aliasId);
+
+    @Nullable
+    String getAliasZid();
```
hsinn0, line 32: I suppose that aliasZid is aliasZoneId, right? Might be better to either document it (javadoc) or rename the property to something clearer?

adrianhoelzl-sap: Added Javadoc comments in commit 0197893

hsinn0 (approved):

PR #2747, opened 2024-02-26
avoid NPE

klaus-sap:


adrianhoelzl-sap (approved):

server/src/main/java/org/cloudfoundry/identity/uaa/logging/LogSanitizerUtil.java
```diff
@@ -1,10 +1,15 @@
 package org.cloudfoundry.identity.uaa.logging;
 
+import javax.annotation.Nullable;
```
strehle: Please check other parts uses org.springframework.lang.Nullable;

In general I would omit the annotation

strehle (commented):
please add a test
https://github.com/cloudfoundry/uaa/blob/develop/server/src/test/java/org/cloudfoundry/identity/uaa/logging/LogSanitizerUtilTest.java

server/src/main/java/org/cloudfoundry/identity/uaa/logging/LogSanitizerUtil.java
klaus-sap: changed import to org.springframework.lang.Nullable, which is used in IdentityProviderEndpoints.java and IdpAliasFailedException.java

hsinn0 (approved):

PR #2749, opened 2024-02-28
doc: clarify token revocation

peterhaochen47:


uaa/slateCustomizations/source/index.html.md.erb
```diff
@@ -378,6 +378,29 @@ _Response Fields_
   Added in UAA 3.3.0
 </aside>
 
+When an access token is revoked using this endpoint, the UAA Introspect Token endpoint (``/introspect``)
```
bruce-ricard: > an access token is revoked using this endpoint

We are trying to document what it means to "revoke a token using this endpoint", so we should try to not use this phrasing in the documentation.

I'm thinking of something like

> When an access token is passed to the `/revoke` endpoint, and the UAA responds positively to the request by returning a 200 HTTP code, the same token then passed to the `/introspect` endpoint will respond with `active: false`. We will say that the token has been revoked.

uaa/slateCustomizations/source/index.html.md.erb
```diff
@@ -378,6 +378,29 @@ _Response Fields_
...
+will respond with ``"active": false`` when presented with the revoked access token.
+
+If the access token is in the JWT format (as opposed to the opaque format), the server config ``uaa.jwt.revocable`` or
+the Identity Zone config ``config.tokenPolicy.jwtRevocable`` must be set to ``true`` for
+this feature to work. However, OAuth resource servers are generally not required to call the UAA Introspect
```
bruce-ricard: > this feature to work

maybe "for the revocation to work"?

I may also be useful to document what the UAA returns if these settings are `false` and someone tries to revoke a token?

bruce-ricard: > resource servers are generally not required to call the UAA Introspect

They are *never* required to call the introspect endpoint. It's just that when they have an opaque token, they don't have an option. And when they have a jwt one, they shouldn't, though they may.

uaa/slateCustomizations/source/index.html.md.erb
```diff
@@ -378,6 +378,29 @@ _Response Fields_
...
+Token endpoint to validate the status of the token. Once issued, a valid access token
```
bruce-ricard: > the status of the token

the validity of the token, maybe?

uaa/slateCustomizations/source/index.html.md.erb
```diff
@@ -378,6 +378,29 @@ _Response Fields_
...
+in the JWT format is generally considered valid until its expiry. Hence, we do not recommend
```
bruce-ricard: > is generally considered valid until its expiry

IMO the "generally" is confusing. I would phrase it:

> A JWT token should be considered valid and irrevocable until its expiry.

uaa/slateCustomizations/source/index.html.md.erb
```diff
@@ -378,6 +378,29 @@ _Response Fields_
...
+relying on this endpoint to revoke access tokens in the JWT format. If the ability
+to remove/limit access after the tokens are issued is important to you, we recommend the following instead:
+
+* Ask the OAuth client to use opaque tokens only, so that the OAuth resource server is required to use
+the UAA Introspect Token endpoint to validate that the tokens have not been revoked.
```
bruce-ricard, line 396: I personally understand this sentence, though I feel like it could confuse novice users through its length and complexity.

uaa/slateCustomizations/source/index.html.md.erb
```diff
@@ -378,6 +378,29 @@ _Response Fields_
...
+
+If the access token is in the JWT format (as opposed to the opaque format), the server config ``uaa.jwt.revocable`` or
+the Identity Zone config ``config.tokenPolicy.jwtRevocable`` must be set to ``true`` for
+this feature to work. However, OAuth resource servers are generally not required to call the UAA Introspect
+Token endpoint to validate the status of the token. Once issued, a valid access token
+in the JWT format is generally considered valid until its expiry. Hence, we do not recommend
+relying on this endpoint to revoke access tokens in the JWT format. If the ability
+to remove/limit access after the tokens are issued is important to you, we recommend the following instead:
+
+* Ask the OAuth client to use opaque tokens only, so that the OAuth resource server is required to use
```
bruce-ricard, line 395: > Ask the OAuth client to use opaque tokens only

That's another thing that IMHO is ridiculous. Why is the client the one deciding the token format? But that's beyond the scope of this work.

uaa/slateCustomizations/source/index.html.md.erb
```diff
@@ -378,6 +378,29 @@ _Response Fields_
...
+the UAA Introspect Token endpoint to validate that the tokens have not been revoked.
+* If the access tokens are in the JWT format, configure the access tokens to be short-lived
+(e.g. a few minutes), and when needed, revoke the more long-lived refresh tokens so that they
+may no longer be used to obtain refreshed access tokens.
+
+When a refresh token is in the opaque format and revoked using this endpoint, the refresh token
+will no longer be considered valid when used to perform the Refresh Token grant.
+When a refresh token is in the JWT format, the server config ``uaa.jwt.revocable`` or
+the Identity Zone config ``config.tokenPolicy.jwtRevocable`` must be set to ``true`` for
+this feature to work.
```
bruce-ricard: Same comments as for the similar access token section above

bruce-ricard (commented):
Maybe could you start the section by saying something like

> Both access and refresh tokens can potentially be revoked

and add 2 subsections for each token type?


Sorry if I'm too nit-picky, I think that it's important to be precise and clear so that readers can't misunderstand what we are trying to explain.
But I technically agree with everything you wrote.

uaa/slateCustomizations/source/index.html.md.erb
bruce-ricard: Also, I would like us to be more clear and specific and add something like

> Even-though the `/revoke` endpoint may return a 200 OK when calling it with a JWT token, a JWT token should be considered valid and irrevocable until its expiry, because resource servers may not know that the token was revoked, and hence still grant access to the resource to a client presenting such token.

uaa/slateCustomizations/source/index.html.md.erb
peterhaochen47: Point discussed and/or edits made.

uaa/slateCustomizations/source/index.html.md.erb
peterhaochen47: Point discussed and/or edits made.

peterhaochen47: Point discussed and/or edits made.

uaa/slateCustomizations/source/index.html.md.erb
peterhaochen47: Point discussed and/or edits made.

uaa/slateCustomizations/source/index.html.md.erb
peterhaochen47: Point discussed and/or edits made.

uaa/slateCustomizations/source/index.html.md.erb
peterhaochen47: Point discussed and/or edits made.

uaa/slateCustomizations/source/index.html.md.erb
peterhaochen47: Point discussed and/or edits made.

uaa/slateCustomizations/source/index.html.md.erb
peterhaochen47: Point discussed and/or edits made.

uaa/slateCustomizations/source/index.html.md.erb
```diff
@@ -398,11 +398,18 @@ the UAA Introspect Token endpoint to validate that the tokens have not been revo
 (e.g. a few minutes), and when needed, revoke the more long-lived refresh tokens so that they
 may no longer be used to obtain refreshed access tokens.
 
-When a refresh token in the opaque format is successfully passed to the ``/revoke`` endpoint,
+When a refresh token is successfully passed to the ``/revoke`` endpoint,
```
bruce-ricard: Now I think that I found a better wording for this.

> When the `/revoke` endpoint is successfully invoked with a refresh token...

Because the successfulness should be about the endpoint invocation, not the passing of the token. One can successfully pass a token, and make an unsuccessful request with it.

uaa/slateCustomizations/source/index.html.md.erb
```diff
@@ -398,11 +398,18 @@ the UAA Introspect Token endpoint to validate that the tokens have not been revo
...
 the refresh token can no longer be used to perform the Refresh Token grant.
-When a refresh token is in the JWT format, the server config ``uaa.jwt.revocable`` or
-the Identity Zone config ``config.tokenPolicy.jwtRevocable`` must be set to ``true`` for
-the "Revoke a single token" endpoint (``/oauth/token/revoke/{tokenId}``) to work.
+
+Refresh tokens in any format can be revoked using the "Revoke all tokens for a user" endpoint (``/oauth/token/revoke/user/{userId}``),
+the "Revoke all tokens for a client" endpoint (``/oauth/token/revoke/client/{clientId}``), or
+the "Revoke all tokens for a user and client combination" endpoint (``/oauth/token/revoke/user/{userId}/client/{clientId}``).
+
```
bruce-ricard, line 407: Maybe link to the API docs could be useful for all the different endpoints?

uaa/slateCustomizations/source/index.html.md.erb
```diff
@@ -398,11 +398,18 @@ the UAA Introspect Token endpoint to validate that the tokens have not been revo
...
+Refresh tokens in the opaque format can be individually revoked using
+the "Revoke a single token" endpoint (``/oauth/token/revoke/{tokenId}``).
+However, refresh tokens in the JWT format can only be individually revoked using
```
bruce-ricard, line 410: I forgot if you said you validated this one or not. Because this would be out of spec. As we validated yesterday, the spec says that refresh tokens MUST be revocable.

bruce-ricard (approved):
Looks quite good! Just a couple thoughts added inline.

uaa/slateCustomizations/source/index.html.md.erb
peterhaochen47: This is validated (see story comment). This scenario would indeed be out of spec. However, UAA can be configured to be spec-compliant (when refresh token format is set to opaque). There are many other examples in UAA where it can be configured to be either spec compliant or out of spec (aka the previous behavior, and hence no breaking change). So the fact that UAA can be configured to be out of spec is not an immediate problem in my mind. 

It is a problem that the default value of refresh token format is jwt (which would make UAA out of spec, when jwt.revocable = false, which is also the default value). Because UAA should be spec-compliant by default. But in any case, I see that as out of scope of the current doc improvement PR).

uaa/slateCustomizations/source/index.html.md.erb
peterhaochen47: change pushed

uaa/slateCustomizations/source/index.html.md.erb
peterhaochen47: I find that more than trivial to execute, but please feel free to make further improvements to the doc yourself.

uaa/slateCustomizations/source/index.html.md.erb
peterhaochen47: issue filed: https://github.com/cloudfoundry/uaa-release/issues/813

strehle (approved):
Thanks.
We should think about to prevent JWT for refresh tokens but for now good doc

peterhaochen47:
> Thanks. We should think about to prevent JWT for refresh tokens but for now good doc

Thanks! Yes, one issue filed here: https://github.com/cloudfoundry/uaa-release/issues/813

PR #2756, opened 2024-03-01
[do not merge - will fail] Add test to demonstrate JWT creation issue

strehle:
Using JWTClaimsSet has impact that aud content might be a simple string. UAA return always an array before.


server/src/test/java/org/cloudfoundry/identity/uaa/oauth/jwt/ChainedSignatureVerifierTests.java
```diff
@@ -88,31 +88,31 @@ public void setup() {
     @Test
     public void test_single_key_valid() {
         verifier = new ChainedSignatureVerifier(new JsonWebKeySet<>(Collections.singletonList(validKey)));
-        signedValidContent.verifySignature(verifier);
+        JwtHelper.decode(signedValidContent.getEncoded()).verifySignature(verifier);
```
strehle: these corrections are for a better API unterstanding because singing and verification should be done with extra API objects ... before it was with internal state of the objects.... and therefore not really doing a signature check...

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/jwt/CommonSignerTest.java
```diff
@@ -109,4 +115,32 @@ public void test_mac_signing_options() throws JOSEException, ParseException {
...
+        CommonSigner signer = new CommonSigner("id", rsaSigningKey, "http://localhost/uaa");
+        assertEquals("RS256", signer.algorithm());
+        assertEquals("id", signer.keyId());
+        JWSHeader header = new JWSHeader.Builder(JWSAlgorithm.RS256).type(JOSEObjectType.JWT).build();
+        JWTClaimsSet claimsSet = JWTClaimsSet.parse(objectMap);
+        SignedJWT resultedJwt = new SignedJWT(header, claimsSet);
+        resultedJwt.sign(signer);
+        String payLoadString = JWTParser.parse(resultedJwt.serialize()).getParsedParts()[1].decodeToString();
+        // then
+        assertThat(payLoadString, Matchers.containsString("\"aud\":\"single\""));
```
strehle: this plain decode and match is done to ensure type.... if we could parse it with own Claims class or JWTClaimsSet then we would get an array...

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/jwt/CommonSignerTest.java
```diff
@@ -109,4 +115,32 @@ public void test_mac_signing_options() throws JOSEException, ParseException {
...
+
+    @Test
+    public void test_uaa_singing_with_single_aud_value() throws JOSEException, ParseException {
+        // given
+        Map<String, Object> objectMap = Map.of("sub","1234567890", "name", "John Doe", "aud", Arrays.asList("single"));
+        // when
+        String uaaResultedJwt = JwtHelper.encode(objectMap, KeyInfoBuilder.build("id", rsaSigningKey, "http://localhost/uaa")).getEncoded();
+        String payLoadString = JWTParser.parse(uaaResultedJwt).getParsedParts()[1].decodeToString();
+        // then
+        assertThat(payLoadString, Matchers.containsString("\"aud\":[\"single\"]"));
```
strehle: A fix must meet here the expectation

strehle (commented):
see inline

hsinn0 (approved):

strehle:
can be closed after merging of https://github.com/cloudfoundry/uaa/pull/2757

PR #2758, opened 2024-03-04
Load jwtClientAuthentication in uaa.yml 

### LINKED ISSUE #2752: Support reuse of OIDC identity provider configuration jwtClientAuthentication on a global level

77.1.0

I am deploying the UAA
- using cf-deployment

### What did you do?

Creation of multiple "related" OIDC identity providers in the default identity zones via jwtClientAuthentication. These "related" OIDC identity providers share some of its configuration (e.g. kid). According to the [documentation](https://docs.cloudfoundry.org/api/uaa/version/77.1.0/index.html#oauth-oidc) this configuration has the following sub-parameters:

- kid: UAA 76.18.0 Custom key from your defined keys, defaults to activeKeyId
- iss: Custom issuer, see RFC 7523, defaults to relyingPartyId
- aud: Custom audience, see RFC 7523, defaults to tokenUrl

After the credential rotation for the configured OIDC identity providers, some shared configuration entries require updates. I want to update this shared configuration only once.

### What did you see instead?

After the credential rotation for the configured OIDC identity provider, it is required the shared configuration in multiple identity providers of the default zones.

#### Idea

* Support "central" maintenance of the shared configuration (e.g. kid) for OIDC identity providers in the default zone (via uaa.yml)
* Possibility to reference "centrally" maintained configuration (from jwtClientAuthentication on identity provider level) in the default identity zone

strehle:
By mistake the variable jwtclientAuthentication was used in PR for clientAuthentication. 
Documentation explains jwtClientAuthentication as entry

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/OauthIDPWrapperFactoryBean.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/OauthIDPWrapperFactoryBean.java
@@ -23,6 +23,7 @@ import org.cloudfoundry.identity.uaa.util.JsonUtils;
 
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
@@ -88,24 +89,28 @@ public class OauthIDPWrapperFactoryBean {
             idpDefinitionMap.get("passwordGrantEnabled") == null ? false : (boolean) idpDefinitionMap.get("passwordGrantEnabled"));
         oidcIdentityProviderDefinition.setSetForwardHeader(idpDefinitionMap.get("setForwardHeader") == null ? false : (boolean) idpDefinitionMap.get("passwordGrantEnabled"));
         oidcIdentityProviderDefinition.setPrompts((List<Prompt>) idpDefinitionMap.get("prompts"));
-        setJwtClientAuthentication("jwtclientAuthentication", idpDefinitionMap, oidcIdentityProviderDefinition);
+        setJwtClientAuthentication(idpDefinitionMap, oidcIdentityProviderDefinition);
         oauthIdpDefinitions.put(alias, oidcIdentityProviderDefinition);
         rawDef = oidcIdentityProviderDefinition;
         provider.setType(OriginKeys.OIDC10);
         return rawDef;
     }
 
-    private static void setJwtClientAuthentication(String entry, Map<String, Object> map, OIDCIdentityProviderDefinition definition) {
-        if (map.get(entry) != null) {
-            if (map.get(entry) instanceof Boolean) {
-                boolean jwtClientAuthentication = (Boolean) map.get(entry);
-                if (jwtClientAuthentication) {
-                    definition.setJwtClientAuthentication(new HashMap<>());
+    private static void setJwtClientAuthentication(Map<String, Object> map, OIDCIdentityProviderDefinition definition) {
+        Object jwtClientAuthDetails = getJwtClientAuthenticationDetails(map, List.of("jwtClientAuthentication", "jwtclientAuthentication"));
+        if (jwtClientAuthDetails != null) {
+            if (jwtClientAuthDetails instanceof Boolean boolValue) {
+                if (boolValue.booleanValue()) {
+                    definition.setJwtClientAuthentication(Collections.emptyMap());
                 }
-            } else if (map.get(entry) instanceof HashMap) {
-                definition.setJwtClientAuthentication(map.get(entry));
+            } else if (jwtClientAuthDetails instanceof Map) {
+                definition.setJwtClientAuthentication(jwtClientAuthDetails);
             }
         }
+    }
+
+    private static Object getJwtClientAuthenticationDetails(Map<String, Object> uaaYamlMap, List<String> entryInUaaYaml) {
+        return entryInUaaYaml.stream().filter(e -> uaaYamlMap.get(e) != null).findFirst().map(uaaYamlMap::get).orElse(null);
     }
 
     public static IdentityProviderWrapper getIdentityProviderWrapper(String origin, AbstractExternalOAuthIdentityProviderDefinition rawDef, IdentityProvider provider, boolean override) {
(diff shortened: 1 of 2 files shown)


peterhaochen47:
Before I look at this PR, I want to understand the original issue first. See [comment](https://github.com/cloudfoundry/uaa/issues/2752#issuecomment-1982068000).

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/OauthIDPWrapperFactoryBean.java
```diff
@@ -88,26 +89,30 @@ private AbstractExternalOAuthIdentityProviderDefinition getExternalOIDCIdentityP
...
     }
 
-    private static void setJwtClientAuthentication(String entry, Map<String, Object> map, OIDCIdentityProviderDefinition definition) {
-        if (map.get(entry) != null) {
-            if (map.get(entry) instanceof Boolean) {
-                boolean jwtClientAuthentication = (Boolean) map.get(entry);
-                if (jwtClientAuthentication) {
-                    definition.setJwtClientAuthentication(new HashMap<>());
+    private static void setJwtClientAuthentication(Map<String, Object> map, OIDCIdentityProviderDefinition definition) {
+        Object jwtClientAuthDetails = getJwtClientAuthenticationDetails(map, List.of("jwtClientAuthentication", "jwtclientAuthentication"));
```
strehle, line 100: Have a list of 2 possible entries in yaml

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/OauthIDPWrapperFactoryBean.java
```diff
@@ -88,26 +89,30 @@ private AbstractExternalOAuthIdentityProviderDefinition getExternalOIDCIdentityP
...
                 }
-            } else if (map.get(entry) instanceof HashMap) {
-                definition.setJwtClientAuthentication(map.get(entry));
+            } else if (jwtClientAuthDetails instanceof Map) {
+                definition.setJwtClientAuthentication(jwtClientAuthDetails);
             }
         }
     }
 
+    private static Object getJwtClientAuthenticationDetails(Map<String, Object> uaaYamlMap, List<String> entryInUaaYaml) {
```
strehle, line 112: this method is using the list and returns the first found

server/src/test/java/org/cloudfoundry/identity/uaa/provider/oauth/OauthIdentityProviderDefinitionFactoryBeanTest.java
```diff
@@ -167,6 +168,24 @@ public void jwtClientAuthenticationWithCustomSetting() {
         assertEquals("issuer", (((Map<String, String>)((OIDCIdentityProviderDefinition) factoryBean.getProviders().get(0).getProvider().getConfig()).getJwtClientAuthentication()).get("iss")));
     }
 
+    @Test
+    public void jwtClientAuthenticationWith2EntriesButNewOneMustWin() {
```
strehle, line 172: this test is showing the theoretic problem

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/OauthIDPWrapperFactoryBean.java
```diff
@@ -88,26 +89,30 @@ private AbstractExternalOAuthIdentityProviderDefinition getExternalOIDCIdentityP
...
-    private static void setJwtClientAuthentication(String entry, Map<String, Object> map, OIDCIdentityProviderDefinition definition) {
-        if (map.get(entry) != null) {
-            if (map.get(entry) instanceof Boolean) {
-                boolean jwtClientAuthentication = (Boolean) map.get(entry);
-                if (jwtClientAuthentication) {
-                    definition.setJwtClientAuthentication(new HashMap<>());
+    private static void setJwtClientAuthentication(Map<String, Object> map, OIDCIdentityProviderDefinition definition) {
+        Object jwtClientAuthDetails = getJwtClientAuthenticationDetails(map, List.of("jwtClientAuthentication", "jwtclientAuthentication"));
+        if (jwtClientAuthDetails != null) {
+            if (jwtClientAuthDetails instanceof Boolean boolValue) {
```
strehle, line 102: Sonar refactorings only....

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/OauthIDPWrapperFactoryBean.java
```diff
@@ -88,26 +89,30 @@ private AbstractExternalOAuthIdentityProviderDefinition getExternalOIDCIdentityP
...
+                if (boolValue.booleanValue()) {
+                    definition.setJwtClientAuthentication(Collections.emptyMap());
                 }
-            } else if (map.get(entry) instanceof HashMap) {
-                definition.setJwtClientAuthentication(map.get(entry));
+            } else if (jwtClientAuthDetails instanceof Map) {
```
strehle, line 106: Sonar refactorings only

strehle (commented):
@peterhaochen47

peterhaochen47:
Thanks @strehle! I personally think that fixing the code to be aligned with the correct case-sensitive field `jwtClientAuthentication` as documented in our API doc (without being compatible with the incorrect `jwtclientAuthentication`) is okay, as it can be justified as a bug fix (and not as a breaking change). 

What do you think? Are there a lot of adoptions of the incorrect `jwtclientAuthentication` on your side already?

strehle:
> What do you think? Are there a lot of adoptions of the incorrect `jwtclientAuthentication` on your side already?

No I am sure, that not many adoptions out with incorrect setting, but with this PR both settings would be used

But I asked me to change this PR so that only the correct setting is used?

peterhaochen47:
> > What do you think? Are there a lot of adoptions of the incorrect `jwtclientAuthentication` on your side already?
> 
> No I am sure, that not many adoptions out with incorrect setting, but with this PR both settings would be used
> 
> But I asked me to change this PR so that only the correct setting is used?

Simply correcting the code to comply with the doc (without being compatible with the incorrect `jwtclientAuthentication`) would be my preference, as it's technically a bug fix, rather than an intentional field name migration. But if maintaining compatibility with the incorrect field is important to you, I'm okay with the current code too.

strehle:
> I'm okay with the current code too.

Ok, then I would like to stay with current code. In current code jwtClientAuthentication is used first, so no extra step if correct name is used, but for others jwtclientAuthentication would still work

peterhaochen47 (approved):

PR #2763, opened 2024-03-04
feat: Use Database availability as indicator for /healthz response

tack-sap:
As the original PR #2718 got reverted, here the next try to implement a /healthz that uses database availability as idicator for the response.

The difference in this PR is, that the response for the "happy case" is not changed and thus backwards compatible (like the original PR initially also tried before the more detailed status messages were introduced).


torsten-sap:
@Tallicia @swalchemist Could you please review as you were already involved in the initial PR?

peterhaochen47:
I will take look.

swalchemist (approved):

PR #2765, opened 2024-03-05
Add 'aliasId' and 'aliasZid' Fields to ScimUser

adrianhoelzl-sap:
see issue #2505 

This PR is a preparation for the alias handling for SCIM users:

- added fields and the corresponding getters/setters to `ScimUser` class
- FlyWay migrations for adding the necessary columns to the `users` table
- adjustments of the queries used in `JdbcScimUserProvisioning`


model/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUser.java
swalchemist, line 1: There is a Sonar warning about adding an equals() method on ScimUser. It was dismissed, but there is still no equals() method on this class, only on some of its inner classes.

model/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUser.java
```diff
@@ -839,6 +862,9 @@ public void patch(ScimUser patch) {
             }
             setPhoneNumbers(current);
         }
+
+        ofNullable(patch.getAliasId()).ifPresent(this::setAliasId);
```
swalchemist: This is a very long method and it really shouldn't get longer. Could we have a refactor to make it at least a little smaller?

model/src/test/java/org/cloudfoundry/identity/uaa/scim/ScimUserTests.java
```diff
@@ -451,14 +451,11 @@ public void testPhoneNumber() {
         assertEquals("type", p1.getType());
         ScimUser user = new ScimUser();
         user.setPhoneNumbers(Collections.singletonList(p1));
-        try {
+
+        assertThatIllegalArgumentException().isThrownBy(() -> {
             p1.setType(null);
```
swalchemist: Nice use of AssertJ! 

Could the call to setType() move outside this block, so we're sure the exception comes from addPhoneNumber()?

model/src/test/java/org/cloudfoundry/identity/uaa/scim/ScimUserTests.java
```diff
@@ -495,6 +492,33 @@ public void test_patch_previous_logon() {
         assertNull(user.getPreviousLogonTime());
     }
 
+    @Test
+    public void testPatchAliasId() {
+        final String aliasId = UUID.randomUUID().toString();
+        patch.setAliasId(aliasId);
+        user.patch(patch);
+        Assertions.assertThat(user.getAliasId()).isEqualTo(aliasId);
```
swalchemist: We already have several calls to assertEquals() and no uses of isEqualTo(). I recommend using only one or the other in each test class.

server/src/test/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioningTests.java
```diff
@@ -101,6 +114,8 @@ void setUp(@Autowired LimitSqlAdapter limitSqlAdapter) {
 
         jdbcScimUserProvisioning = new JdbcScimUserProvisioning(jdbcTemplate, pagingListFactory, passwordEncoder, idzManager);
 
+        ReflectionTestUtils.setField(jdbcScimUserProvisioning, "jdbcIdentityZoneProvisioning", jdbcIdentityZoneProvisioning);
```
swalchemist: How about adding a way to properly inject this dependency via a constructor or setter so we don't have to use reflection?

uaa/src/test/java/org/cloudfoundry/identity/uaa/scim/endpoints/ScimUserEndpointsMockMvcTests.java
```diff
@@ -397,6 +398,109 @@ void create_user_without_email() throws Exception {
                                         .put("error", "invalid_scim_resource"))));
     }
 
+    @Test
```
swalchemist: I can't figure out from the test code why the alias is supposed to be ignored in all the tests in this class. Maybe update the comments here to explain?

swalchemist (changes requested):

model/src/test/java/org/cloudfoundry/identity/uaa/scim/ScimUserTests.java
adrianhoelzl-sap: Thanks for the hint, I fixed it in commit 5e89664.

model/src/test/java/org/cloudfoundry/identity/uaa/scim/ScimUserTests.java
adrianhoelzl-sap: fixed in commit 1f2614c

model/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUser.java
adrianhoelzl-sap: I refactored the method in commit 45f13d0.

server/src/test/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioningTests.java
adrianhoelzl-sap: fixed with commit 60cb8ca

uaa/src/test/java/org/cloudfoundry/identity/uaa/scim/endpoints/ScimUserEndpointsMockMvcTests.java
adrianhoelzl-sap: fixed with commit 22f80aa

model/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUser.java
adrianhoelzl-sap: added `equals` method with commit 3061277

adrianhoelzl-sap: After some investigation, I decided to revert the added `equals` method since this is a non-trivial change of behavior.

Prior to the change, all objects from classes extending `ScimCore` (e.g., `ScimUser` and `ScimGroup`) were considered equal if only their `id` fields are equal (see also [this test case](https://github.com/cloudfoundry/uaa/blob/1860cd47d6921be651737fbbdf229188fc18f158/server/src/test/java/org/cloudfoundry/identity/uaa/scim/ScimCoreTests.java#L29)). By also considering the other fields, we would change this behavior throughout the entire application.

I propose to address this in a separate issue/PR. What do you think?

swalchemist: Sure, sounds good, if the equals() behavior is supposed to be the same and the tests already validate that.

swalchemist (approved):
Thanks for making improvements in the code.

PR #2767, opened 2024-03-06
Fix uaa start. Prevent exception if encryption section missing

### LINKED ISSUE #2766: Either fix missing encryption_keys or remove it

Latest UAA , develop

I am deploying the UAA

- locally only using gradlew
- using cf-deployment
- other (please explain)

remove entries from [encryption_keys](https://github.com/cloudfoundry/uaa/blob/develop/scripts/cargo/uaa.yml#L7-L11)
start UAA
./gradlew run

### What did you do?

started uaa without https://github.com/cloudfoundry/uaa/blob/develop/scripts/cargo/uaa.yml#L7-L11

A started UAA

### What did you see instead?

UAA does not start because of
Caused by: java.lang.IllegalArgumentException: Could not resolve placeholder 'encryption.active_key_label' in value "${encryption.active_key_label}"

But the EncryptionService bean is not in use (if should server MFA). The infrastructure itself is maybe re-usable for other configuration options later

Currently we force all operators in UAA to set 

encryption:
  active_key_label: CHANGE-THIS-KEY
  encryption_keys:
  - label: CHANGE-THIS-KEY
    passphrase: CHANGEME

see uaa.yaml from [cargo ](https://github.com/cloudfoundry/uaa/blob/develop/scripts/cargo/uaa.yml) but we dont use it

strehle:
The encryption service is not usable if no passphrase set, but that service is not in use at all - currently.


peterhaochen47:
Summary of OSS sync: this was for MFA feature (which is already removed); hence these fields are currently unused. But this PR keeps them in case there is future usage.

strehle:
@peterhaochen47 @Tallicia can you please take this PR in review. For us it means less Ops but for UAA there is no risk to break something, because these classes are not in use - but they need a secret key currently

strehle:
sonar : 
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2767

server/src/test/java/org/cloudfoundry/identity/uaa/cypto/EncryptionServiceTest.java
```diff
@@ -44,4 +46,9 @@ public void decrypt_shouldNotDecryptWithInvalidPassphrase() throws EncryptionSer
         assertThat(ciphertext, is(notNullValue()));
         new EncryptionService("invalid-password").decrypt(ciphertext);
     }
+
+    @Test
+    public void initShouldWorkWithEmptyKey() {
```
Tallicia, line 51: Good to see the test!

server/src/test/java/org/cloudfoundry/identity/uaa/cypto/EncryptionKeyServiceTest.java
```diff
@@ -58,11 +59,16 @@ public void shouldThrowErrorIfActiveEncryptionKeyHasNotBeenProvided() {
...
-        expectedException.expect(NoActiveEncryptionKeyProvided.class);
-        expectedException.expectMessage("UAA cannot be started without encryption key value uaa.encryption.active_key_label");
-
+    public void shouldThrowErrorIfEmptyActiveKeyLabelIsProvided() {
         encryptionKeyService = new EncryptionKeyService("", new ArrayList<>());
+        assertNull(encryptionKeyService.getActiveKey());
+    }
+
+    @Test
+    public void shouldNotThrowErrorIfNoActiveKeyLabelIsProvided() {
```
Tallicia, line 68: good test

Tallicia (approved):
This looks good.

PR #2769, opened 2024-03-07
Alias Handler for SCIM Users

adrianhoelzl-sap:
see issue #2505 

Preparation for alias feature for SCIM users; adds subclass of `EntityAliasHandler` that handles alias entities of SCIM users.

Prerequisites:

- #2765
- #2824


server/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandler.java
```diff
@@ -0,0 +1,158 @@
...
+        return EntityAliasHandler.isValidAliasPair(idpInCurrentZone.get(), idpInAliasZone.get());
+    }
+
+    @Override
+    protected void setPropertiesFromExistingAliasEntity(
+            final ScimUser newAliasEntity,
+            final ScimUser existingAliasEntity
+    ) {
+        // these three timestamps should not be overwritten by the timestamps of the original user
+        newAliasEntity.setPasswordLastModified(existingAliasEntity.getPasswordLastModified());
```
Tallicia, line 61: This should be setPasswordLastModifiedTime to align to the naming of the two following Time setters.

server/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandler.java
```diff
@@ -0,0 +1,158 @@
...
+        newAliasEntity.setLastLogonTime(existingAliasEntity.getLastLogonTime());
+        newAliasEntity.setPreviousLogonTime(existingAliasEntity.getPreviousLogonTime());
+    }
```
Tallicia, line 64: Why are these setting the Time to the existingAliasEntity instead of "now" or even null for the 2 logon times -  a new Alias should have it's own timestamps. Setting them to the existingAlias seems it would be confusing when it was actually Modified or logged into.

server/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandler.java
```diff
@@ -0,0 +1,158 @@
...
+    private final ScimUserProvisioning scimUserProvisioning;
+    private final IdentityProviderProvisioning identityProviderProvisioning;
+    private final IdentityZoneManager identityZoneManager;
+
+    protected ScimUserAliasHandler(
+            @Qualifier("identityZoneProvisioning") final IdentityZoneProvisioning identityZoneProvisioning,
+            final ScimUserProvisioning scimUserProvisioning,
+            final IdentityProviderProvisioning identityProviderProvisioning,
+            final IdentityZoneManager identityZoneManager,
+            @Value("${login.aliasEntitiesEnabled:false}") final boolean aliasEntitiesEnabled
```
strehle: the 2nd place for aliasEntitiesEnabled , maybe create an extra bean or component and use here a ref

server/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandler.java
```diff
@@ -0,0 +1,158 @@
...
+        return scimUserProvisioning.update(entity.getId(), entity, zoneId);
+    }
+
+    @Override
+    protected ScimUser createEntity(final ScimUser entity, final String zoneId) {
+        try {
+            return scimUserProvisioning.createUser(entity, entity.getPassword(), zoneId);
+        } catch (final ScimResourceAlreadyExistsException e) {
+            final String errorMessage = String.format(
+                    "Could not create %s. A user with the same username already exists in the alias zone.",
```
strehle, line 146: did not found a test which cover this part

server/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandler.java
```diff
@@ -0,0 +1,158 @@
...
+        aliasUser.setName(new ScimUser.Name(originalEntity.getGivenName(), originalEntity.getFamilyName()));
+
+        aliasUser.setEmails(originalEntity.getEmails());
+        aliasUser.setPhoneNumbers(originalEntity.getPhoneNumbers());
+
+        aliasUser.setActive(originalEntity.isActive());
+        aliasUser.setVerified(originalEntity.isVerified());
+
+        // idzId and alias properties will be set later
+        aliasUser.setZoneId(null);
```
strehle: you dont need to set objects to null, why these many calls with null

server/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandler.java
```diff
@@ -0,0 +1,158 @@
...
+        aliasUser.setPasswordLastModified(null);
+        aliasUser.setLastLogonTime(null);
+        aliasUser.setPreviousLogonTime(null);
+
+        /* password: empty string
+         *  - alias users are only allowed for IdPs that also have an alias
+         *  - IdPs can only have an alias if they are of type SAML, OIDC or OAuth 2.0
+         *  - users with such IdPs as their origin always have an empty password
+         */
+        aliasUser.setPassword(EMPTY_STRING);
```
strehle, line 110: empty is not what I would expect. is this a required field?

server/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandler.java
```diff
@@ -0,0 +1,158 @@
...
+        aliasUser.setZoneId(null);
+        aliasUser.setAliasId(null);
+        aliasUser.setAliasZid(null);
+
+        /* these timestamps will be overwritten:
+         *  - creation: with current timestamp during persistence (JdbcScimUserProvisioning)
+         *  - update: with values from existing alias entity */
+        aliasUser.setPasswordLastModified(null);
+        aliasUser.setLastLogonTime(null);
+        aliasUser.setPreviousLogonTime(null);
```
strehle: We do in other places Json object copy, e.g.

ScimUser scimCopy = JsonUtils.convertValue(originalEntity, ScimUser.class);

server/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandler.java
adrianhoelzl-sap: In the create user endpoint, we also set the password explicitly to an empty string for all users with a non-`uaa` origin: https://github.com/cloudfoundry/uaa/blob/6c30c8552fc9c9d4010d9b2c47d6b7771e42210a/server/src/main/java/org/cloudfoundry/identity/uaa/scim/endpoints/ScimUserEndpoints.java#L215

server/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandler.java
adrianhoelzl-sap: I wanted to make it explicit in this method which fields are set and which are `null`.

server/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandler.java
adrianhoelzl-sap: see the comment above

server/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandler.java
adrianhoelzl-sap: unit test added in commit e78e1fd

server/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandler.java
adrianhoelzl-sap: How should this bean look like? Just a wrapper for the boolean value?

peterhaochen47:
Hi, as this PR is a stepping stone toward the overall "alias feature," and as it looks like the changes are limited to the specific code paths related to the feature, I would defer the approval of this PR to @strehle, in terms of judging 1) whether this PR accomplishes its goal, and 2) wether this PR has enough test coverage and user docs (if applicable). 

I'll also approve as well as long as our prior understanding the "alias feature" are still true after this PR: That the functionality & performance of all the existing UAA features are not impacted (do not experience breaking changes / regressions) when the "alias feature" is not turned on (and the feature is off by default).

server/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandler.java
Tallicia: Any update on this question?

server/src/test/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandlerValidationTest.java
```diff
@@ -0,0 +1,286 @@
...
+        class AliasFeatureEnabled extends ExistingAlias_AliasFeatureEnabled {
+            // all tests defined in superclass
+        }
+
+        @Nested
+        class AliasFeatureDisabled extends ExistingAlias_AliasFeatureDisabled {
+            // all tests defined in superclass
+        }
+    }
+}
```
Tallicia: extra linefeed to remve.

Good tests, nice modular implementation.

Tallicia (approved):
Approved with request to answer questions before merging.

server/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandler.java
strehle: I would like to see a configuration either as single option or a more detailed one. You do not need a complete configuration class like ratelimit, e.g. https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/ratelimiting/config/InitialConfig.java , but a central configuration in uaa-release should be created and documented and then you can have a single bean/component in uaa.

server/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandler.java
strehle: sounds not clear to me, because in java the variables are null after initialization, do you not trust JVM ?

server/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandler.java
```diff
@@ -0,0 +1,158 @@
...
+        /* check if an IdP with the user's origin exists in both the current and the alias zone and that they are
+         * aliases of each other */
+        final String origin = requestBody.getOrigin();
+        final Optional<IdentityProvider<?>> idpInAliasZone = retrieveIdpByOrigin(origin, requestBody.getAliasZid());
+        if (idpInAliasZone.isEmpty()) {
+            return false;
+        }
+        final Optional<IdentityProvider<?>> idpInCurrentZone = retrieveIdpByOrigin(origin, identityZoneManager.getCurrentIdentityZoneId());
+        if (idpInCurrentZone.isEmpty()) {
+            return false;
```
strehle, line 50: This part is not covered, please add a Unit or IT

server/src/main/java/org/cloudfoundry/identity/uaa/alias/EntityAliasHandler.java
```diff
@@ -208,4 +216,23 @@ public final Optional<T> retrieveAliasEntity(final T originalEntity) {
...
+
+    protected static <T extends EntityWithAlias> boolean isValidAliasPair(
+            @NonNull final T entity1,
+            @NonNull final T entity2
+    ) {
+        // check if both entities have an alias
+        final boolean entity1HasAlias = hasText(entity1.getAliasId()) && hasText(entity1.getAliasZid());
+        final boolean entity2HasAlias = hasText(entity2.getAliasId()) && hasText(entity2.getAliasZid());
+        if (!entity1HasAlias || !entity2HasAlias) {
+            return false;
```
strehle, line 228: This part is not covered, please add a Unit or IT

strehle (commented):
I would like to see 100 coverage on lines at least. Coverage on all conditions might be hard, but we should have tests with a meaning on all code lines.

Tests are like a technical specification, they later help to understand the code

server/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandler.java
adrianhoelzl-sap: The setters are already named consistently to the underlying fields in the class as well as its JSON representation (`passwordLastModified` does not end with "time", `lastLogonTime` and `previousLogonTime` do), see for example here: https://github.com/cloudfoundry/uaa/blob/091fbf2416472d9278286ce1aed3b41b5e22f8a5/model/src/main/java/org/cloudfoundry/identity/uaa/scim/impl/ScimUserJsonDeserializer.java#L92

I would therefore suggest to leave them as they are. What do you think?

server/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandler.java
adrianhoelzl-sap: This method is used for managing the fields that should differ between the "original" user and its alias, i.e., the properties of an alias that should be independent from the original user. 

During updates, we build a copy of the original user and leave these three timestamps empty. After that, we call this method to overwrite the timestamps with the values from the version of the alias prior to the update.

server/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandler.java
adrianhoelzl-sap: fixed with commit 5699f43

server/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandler.java
adrianhoelzl-sap: added test with commit 97d13b3

server/src/main/java/org/cloudfoundry/identity/uaa/alias/EntityAliasHandler.java
adrianhoelzl-sap: added test with commit 4321693

server/src/test/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandlerValidationTest.java
adrianhoelzl-sap: fixed with commit 31ebf13

server/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandler.java
adrianhoelzl-sap: introduced bean for flag value with commit 823fc2f

strehle:
@adrianhoelzl-sap please cleanup PR , see 
https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&pullRequest=2769&id=cloudfoundry-identity-parent

Rest is ok

server/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandler.java
adrianhoelzl-sap: PR for specification of flag in uaa-release repo: https://github.com/cloudfoundry/uaa-release/pull/899

strehle (approved):

server/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandler.java
Tallicia: Seems the naming inconsistency is more widespread, maybe this would be a good clean up following this PR.

Tallicia: Seems the naming inconsistency is more widespread, maybe this would be a good clean up following this PR.

server/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandler.java
Tallicia: Which one is it, 

> "the properties of an alias that should be independent from the original user."

which I agree with
But this seems contrary to keeping them independent:

> "After that, we call this method to overwrite the timestamps with the values from the version of the alias prior to the update."

Tallicia:
> @adrianhoelzl-sap please cleanup PR , see https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&pullRequest=2769&id=cloudfoundry-identity-parent
> 
> Rest is ok

@strehle This link is not showing no results for me: 

Screenshot 2024-06-04 at 9 11 40 AM

strehle:
> > @adrianhoelzl-sap please cleanup PR , see https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&pullRequest=2769&id=cloudfoundry-identity-parent
> > Rest is ok
> 
> @strehle This link is not showing no results for me:
> 
> Screenshot 2024-06-04 at 9 11 40 AM

there where updates and now smells are solved
See overview https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2769

server/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandler.java
adrianhoelzl-sap: Sorry for the confusion in my explanation, I'll try again:

Let's say there is a user `U` in the "uaa" zone, which has an alias user `A` in the zone "custom". Whenever we perform an update on `U`, we persist the changes, which leads to a newer version `U'` with the changed properties.

Then, to propagate the changes also to the alias, we build a new clone of `U'`, i.e., `A'`. This is done here: https://github.com/cloudfoundry/uaa/blob/f451e8eac58919f28e07330ac04cce9687cb593b/server/src/main/java/org/cloudfoundry/identity/uaa/scim/ScimUserAliasHandler.java#L87

However, as you correctly addressed, `A'` would now have the same timestamp values (last logon, password last modified and previous logon) as `U'`, which is incorrect. Therefore, before persisting `A'`, we look up the version of `A'` before the update, i.e., `A`, and overwrite the timestamp values of `A'` with the timestamp values of `A`.

This is done in the method of this GitHub conversation. The parameter `newAliasEntity` corresponds to `A'`, while `existingAliasEntity` corresponds to `A` (and not to `U` or `U'`, as you might have thought when asking the question).

strehle:
@Tallicia do you have open issues here ?

torsten-sap:
I merge this PR as all questions seem to be addressed and two approvals exist.

PR #2771, opened 2024-03-07
Support own key and cert for jwtClientAuthentication

### LINKED ISSUE #2752: Support reuse of OIDC identity provider configuration jwtClientAuthentication on a global level

strehle:
Load key, cert, alg
Support reference loading

Spec [JwtClientAuthentication-PR2771.pdf](https://github.com/cloudfoundry/uaa/files/14637457/JwtClientAuthentication-PR2771.pdf)


strehle:
current doc
image

strehle:
[JwtClientAuthentication-PR2771.pdf](https://github.com/cloudfoundry/uaa/files/14637457/JwtClientAuthentication-PR2771.pdf)

strehle:
sonar:

https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2771

strehle:
@peterhaochen47 would be great if you could find time to see if this PR here is ok for you.

We will try it then in your SAP staging area if all assumptions are true

peterhaochen47:
> @peterhaochen47 would be great if you could find time to see if this PR here is ok for you.
> 
> We will try it then in our SAP staging area if all assumptions are true

@Tallicia is looking at this PR.

strehle:
> @Tallicia is looking at this PR.

ok thanks . If you have questions we can discuss either tomorrow or you ask in slack/here in PR

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/jwt/JwtClientAuthentication.java
```diff
@@ -158,4 +167,51 @@ private JWTClaimsSet validateClientJWToken(JWT jwtAssertion, JWKSet jwkSet, Stri
...
     }
   }
+
+  private static KeyInfo loadKeyInfo(KeyInfoService keyInfoService, HashMap<String, String> jwtClientConfiguration, String kid) {
+    KeyInfo keyInfo;
+    String signingKey = readJwtClientOption(jwtClientConfiguration.get("key"), null);
+    if (signingKey == null) {
+      keyInfo = Optional.ofNullable(keyInfoService.getKey(kid)).orElseThrow(() -> new BadCredentialsException("Missing requested signing key"));
+    } else {
+      String signingAlg = readJwtClientOption(jwtClientConfiguration.get("alg"), JWSAlgorithm.RS256.getName());
```
Tallicia, line 177: Is RS256 what we want as a default? Should this not default to a more secure algorithm?

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/jwt/JwtClientAuthentication.java
```diff
@@ -158,4 +167,51 @@ private JWTClaimsSet validateClientJWToken(JWT jwtAssertion, JWKSet jwkSet, Stri
...
+      value = defaultOption;
+    }
+    return value;
+  }
+
+  private static Matcher getDynamicValueMatcher(String value) {
+    return DYNAMIC_VALUE_PARAMETER_PATTERN.matcher(value);
+  }
+
+  @SuppressWarnings("java:S1874")
```
Tallicia: What warning is this?

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/jwt/JwtClientAuthenticationTest.java
```diff
@@ -179,10 +196,73 @@ void testGetClientAssertionCustomSingingKey() throws ParseException, JOSEExcepti
...
+    mockApplicationContext(Map.of("jwt.client.kid", "key-id-321"));
+    IdentityZone currentZone = IdentityZone.getUaa();
+    // modify to custom zone
+    currentZone.setId(new AlphanumericRandomValueStringGenerator().generate());
+    currentZone.setSubdomain(new AlphanumericRandomValueStringGenerator().generate());
+    IdentityZoneHolder.set(currentZone);
+    // Expect
+    Exception exception = assertThrows(BadCredentialsException.class, () ->
+        jwtClientAuthentication.getClientAuthenticationParameters(new LinkedMultiValueMap<>(), config));
+    assertEquals("Missing requested signing key", exception.getMessage());
```
Tallicia, line 263: Is the signing key the only thing required? or in other words should this also assert on any other missing items?

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/providers/IdentityProviderEndpointDocs.java
```diff
@@ -705,9 +705,12 @@ void createOidcIdentityProvider() throws Exception {
...
                         "<br>Please note that you can precise the created JWT for client authentication, e.g. if your IdP follows OAuth2 standard according to RFC 7523. For standard OIDC compliance, set true without any further sub-parameters. The supported sub-parameters are " +
-                        "<ul><li>    `kid` <small><mark>UAA 76.18.0</mark></small> Custom key from your defined keys, defaults to `activeKeyId` </li>" +
-                        "<li>        `iss` Custom issuer, see RFC 7523, defaults to `relyingPartyId` </li>" +
-                        "<li>        `aud` Custom audience, see RFC 7523, defaults to `tokenUrl` </li></ul>"),
+                        "<ul><li>    `kid`  <small><mark>UAA 76.18.0</mark></small> Custom key from your defined keys, defaults to `activeKeyId` </li>" +
+                        "<li>        `key`  <small><mark>UAA 77.3.0</mark></small> Optional custom private key, used to generate the client JWT signature </li>" +
```
Tallicia: If something is optional, that means it has a default. I suggest updating this to be consistent to either not use Optional, or anything that has a default to Indicate as Optional. Reading lines 708-713 all appear to be Optional or have defaults but is presented differently on 709-710 lines.

Tallicia: Following the all cap pattern at https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication for client_secret_jwt would be ideal.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/jwt/JwtClientAuthentication.java
Tallicia: Or more specifically, what is deprecated that this warning is about? Is there no alternative to address the deprecation with a replacement?

Tallicia (approved):
This is acceptable, though there are some questions and addressing the warning suppression and java doc consistency would be good followups.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/jwt/JwtClientAuthentication.java
strehle: RS256 is still the standard also for UAA standard JWT signing, see
https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/KeyInfo.java#L66

it is comparable with SHA-256 (saml2, where we now set SHA-1) . it needs a 2048 bit key, so current standard.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/jwt/JwtClientAuthentication.java
strehle: it ignores deprecation warnings in soar for IdentityZoneHolder...

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/jwt/JwtClientAuthenticationTest.java
strehle: signing key (private key) is the most important part. if this is missing, then then certificate is not there as well but the error error is about the most needed one.

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/providers/IdentityProviderEndpointDocs.java
strehle: client_secret_jwt is not wanted from SAP side because with this you have a symmentric/shared key - again. This is same level of security like a secret .
The private_key_jwt is an asymmetric mechanism, see https://datatracker.ietf.org/doc/draft-ietf-oauth-v2-1/ , section 2.4 ->

It is RECOMMENDED to use asymmetric (public-key based) methods for
   client authentication such as mTLS [RFC8705] or using signed JWTs
   ("Private Key JWT") in accordance with [RFC7521] and [RFC7523] (in
   [OpenID] defined as the client authentication method
   private_key_jwt).  When such methods for client authentication are
   used, authorization servers do not need to store sensitive symmetric
   keys, making these methods more robust against a number of attacks.

Tallicia: I'm referring to the doc.

In the doc I linked form you, they list each item as "OPTIONAL" or "REQUIRED". This is a clean way to identify configuration items. The docs here on 708-713 say Optional on some and each has a default telling me they are all optional.

strehle: ok will update documentation

strehle:
Doc

image

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/providers/IdentityProviderEndpointDocs.java
strehle: please see now doc below

PR #2790, opened 2024-03-18
fix: UAA delete user endpoint returns false error during upgrade canary deployment

### LINKED ISSUE #2789: During v76 -> v77 upgrade canary deployment, UAA delete user endpoint returns false error
### Description
During the upgrade to UAA v77 from UAA v76 or below with canary deployment (where briefly both the new and the old UAA servers could be running), UAA delete user endpoint might respond with an error even though the user deletion is successful. 

### Cause
During a canary deployment, briefly both the new and the old UAA servers could be running, resulting in the following:
* The new server (v77): Since the MFA feature is [removed in V77.0.0](https://github.com/cloudfoundry/uaa/pull/2717), a DB migration will be run to drop the DB tables related to the MFA feature. 
* The old server (v76): A user deletion would trigger explicit delete operations on the associated entries in other DB tables. In this case, a user deletion will trigger a delete operation on the MFA-related tables (even in the case where the MFA feature is not enabled). And since these tables have been dropped already by the new server, the operation fails. See relevant code below:
  * [SystemDeletable](https://github.com/cloudfoundry/uaa/blob/79f8ea08e9a41871dded18db5aed749773e7daa6/server/src/main/java/org/cloudfoundry/identity/uaa/audit/event/SystemDeletable.java#L57-L66)
  * [JdbcMfaProviderProvisioning](https://github.com/cloudfoundry/uaa/blob/9940b8d018da87b6e18cec0f08f9c07a65b780b0/server/src/main/java/org/cloudfoundry/identity/uaa/mfa/JdbcMfaProviderProvisioning.java#L19)

V76 -> V77

Any upgrade scenarios with canary deployment.

peterhaochen47:
- fixes https://github.com/cloudfoundry/uaa/issues/2789 (see bug root cause in the issue)
- by bringing back the MFA-related tables exactly as they were

[#187240345]

### DIFF
--- a/server/src/main/resources/org/cloudfoundry/identity/uaa/db/hsqldb/V4_108__Restore_MFA_Tables.sql
+++ b/server/src/main/resources/org/cloudfoundry/identity/uaa/db/hsqldb/V4_108__Restore_MFA_Tables.sql
@@ -1,0 +1,28 @@
+--
+-- These tables were previously dropped in https://github.com/cloudfoundry/uaa/pull/2717
+-- Restoring them here due to https://github.com/cloudfoundry/uaa/issues/2789
+-- Can consider dropping these again in the future (e.g. at UAA V78/79, when most users
+-- will no longer experience issue #2789)
+--
+
+CREATE TABLE mfa_providers (
+  id CHAR(36) NOT NULL PRIMARY KEY,
+  created TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL,
+  lastmodified TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL,
+  identity_zone_id varchar(36) NOT NULL,
+  name varchar(255) NOT NULL,
+  type varchar(255) NOT NULL,
+  config LONGVARCHAR
+);
+
+CREATE TABLE user_google_mfa_credentials (
+  user_id VARCHAR(36) NOT NULL,
+  secret_key VARCHAR(255) NOT NULL,
+  validation_code INTEGER,
+  scratch_codes VARCHAR(255) NOT NULL,
+  mfa_provider_id CHAR(36) NOT NULL,
+  zone_id CHAR(36) NOT NULL,
+  encryption_key_label VARCHAR(255),
+  encrypted_validation_code VARCHAR(255) NULL,
+  PRIMARY KEY (user_id,mfa_provider_id)
+);
--- a/server/src/main/resources/org/cloudfoundry/identity/uaa/db/postgresql/V4_108__Restore_MFA_Tables.sql
+++ b/server/src/main/resources/org/cloudfoundry/identity/uaa/db/postgresql/V4_108__Restore_MFA_Tables.sql
@@ -1,0 +1,29 @@
+--
+-- These tables were previously dropped in https://github.com/cloudfoundry/uaa/pull/2717
+-- Restoring them here due to https://github.com/cloudfoundry/uaa/issues/2789
+--
+-- Can consider dropping these again in the future (e.g. at UAA V78/79, when most users
+-- will no longer experience issue #2789)
+--
+
+CREATE TABLE IF NOT EXISTS mfa_providers (
+  id VARCHAR(36) NOT NULL PRIMARY KEY,
+  created TIMESTAMP default current_timestamp NOT NULL,
+  lastModified TIMESTAMP null,
+  identity_zone_id VARCHAR(36) NOT NULL,
+  name VARCHAR(255) NOT NULL,
+  type VARCHAR(255) NOT NULL,
+  config TEXT
+);
+
+CREATE TABLE IF NOT EXISTS user_google_mfa_credentials (
+  user_id VARCHAR(36) NOT NULL PRIMARY KEY,
+  secret_key VARCHAR(255) NOT NULL,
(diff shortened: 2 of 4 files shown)


Tallicia (approved):
Restore looks good

peterhaochen47:
Some tests are failing on uploading test results:
```
Run actions/upload-artifact@v4
/usr/bin/docker exec  d78f1e7ca0ca2ef54c3345e3b8c1e9465354a02daefa4d92294c23975801e2f3 sh -c "cat /etc/*release | grep ^ID"
With the provided path, there will be 73 files uploaded
Artifact name is valid!
Root directory input is valid!
Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run
```
Not sure what's wrong.

server/src/main/resources/org/cloudfoundry/identity/uaa/db/hsqldb/V4_108__Restore_MFA_Tables.sql
```diff
@@ -0,0 +1,30 @@
...
+  id CHAR(36) NOT NULL PRIMARY KEY,
+  created TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL,
+  lastmodified TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL,
+  identity_zone_id varchar(36) NOT NULL,
+  name varchar(255) NOT NULL,
+  type varchar(255) NOT NULL,
+  config LONGVARCHAR
+);
+
+CREATE UNIQUE INDEX idx_mfa_unique_name ON mfa_providers (identity_zone_id,name);
```
strehle: @peterhaochen47 dont know if we should create indexes .
If flyway migrates 76.31.0 -> 77.3.0 then these tables are deleted in a step before , so no content anymore. Then we create the tables again, but wont use it anymore in 77.x. version.
And if 76.31.0 run in parallel, then the empty DB should not be a performance issue

server/src/main/resources/org/cloudfoundry/identity/uaa/db/postgresql/V4_108__Restore_MFA_Tables.sql
```diff
@@ -0,0 +1,31 @@
...
+  id VARCHAR(36) NOT NULL PRIMARY KEY,
+  created TIMESTAMP default current_timestamp NOT NULL,
+  lastModified TIMESTAMP null,
+  identity_zone_id VARCHAR(36) NOT NULL,
+  name VARCHAR(255) NOT NULL,
+  type VARCHAR(255) NOT NULL,
+  config TEXT
+);
+
+CREATE UNIQUE INDEX idx_mfa_unique_name ON mfa_providers (identity_zone_id,LOWER(name));
```
strehle: See index comment before

strehle:
> Not sure what's wrong.

sql syntax for postgresl/mysql is "CREATE TABLE IF NOT EXISTS", corrected it

server/src/main/resources/org/cloudfoundry/identity/uaa/db/hsqldb/V4_108__Restore_MFA_Tables.sql
peterhaochen47: Yes indeed, I was just putting everything in bc I was afraid we would miss something. But no index is fine too.

server/src/main/resources/org/cloudfoundry/identity/uaa/db/postgresql/V4_108__Restore_MFA_Tables.sql
```diff
@@ -0,0 +1,29 @@
+--
```
strehle, line 1: 4_108 is now used from recent merge, so if you want this please incr. number

strehle (commented):
Since you have not merged it yesterday for the release I dont see a reason now to have it, correct ?

peterhaochen47:
> Since you have not merged it yesterday for the release I dont see a reason now to have it, correct ?

We have not merged it because this is still under discussion in slack, and this PR has not received enough approvals.

strehle:
> > Since you have not merged it yesterday for the release I dont see a reason now to have it, correct ?
> 
> We have not merged it because this is still under discussion in slack, and this PR has not received enough approvals.

ok, so you would get an approval, but my personal opinion is, now it is too late and I would not restore tables. UAA was not crashed,.... So too much effort for too less impact

strehle (approved):
please take care for correct versioning of flyway file ... 108 is used alredy.

my personal opinion is, now it is too late and I would not restore tables. UAA was not crashed,.... So too much effort for too less impact

However you can have an approval

peterhaochen47:
Let's keep this PR as a draft, and wait until next week, then I'll discuss this with @bruce-ricard, who worked on this issue with me.

peterhaochen47:
I have discussed with @bruce-ricard, and decided that if SAP, the party that reported this issue, does not care about this fix, and given that our side has not received any reports of this issue, we can close this PR for now. If there's any new complaints, we can always reopen this PR.

For now, we have documented this issue in release notes and this open issue: https://github.com/cloudfoundry/uaa/issues/2789 And we see that as sufficient for now.

PR #2792, opened 2024-03-19
fix: saml signatureAlgorithm in AuthnRequest

swalchemist:
* This fixes a regression introduced in commit https://github.com/cloudfoundry/uaa/commit/d10922a8f606b7d4ed8a7eb6bc585d9226e70f9e where we stopped reading the signatureAlgorithm from the properties file and only used the SHA1 default.

[#187250012]

### DIFF
--- a/uaa/src/main/webapp/WEB-INF/spring/saml-providers.xml
+++ b/uaa/src/main/webapp/WEB-INF/spring/saml-providers.xml
@@ -301,6 +301,10 @@
 
     <bean id="fixedHttpMetaDataProvider" class="org.cloudfoundry.identity.uaa.provider.saml.FixedHttpMetaDataProvider" />
 
+    <bean id="defaultSamlConfig" class="org.cloudfoundry.identity.uaa.provider.saml.SamlConfigurationBean">
+        <property name="signatureAlgorithm" value="${login.saml.signatureAlgorithm:SHA1}"/>
+    </bean>
+
     <beans profile="fileMetadata">
         <bean id="metaDataUrl" class="java.lang.String">
             <constructor-arg value="${login.idpMetadataFile:null}"/>
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/SamlConfigurationBean.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/SamlConfigurationBean.java
@@ -27,6 +27,10 @@ public class SamlConfigurationBean implements InitializingBean {
     signatureAlgorithm = s;
   }
 
+  public SignatureAlgorithm getSignatureAlgorithm() {
+    return signatureAlgorithm;
+  }
+
   @Override
   public void afterPropertiesSet() {
     BasicSecurityConfiguration config = (BasicSecurityConfiguration) Configuration.getGlobalSecurityConfiguration();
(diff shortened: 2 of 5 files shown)


hsinn0 (approved):

uaa/src/main/webapp/WEB-INF/spring/saml-providers.xml
```diff
@@ -301,6 +301,10 @@
 
     <bean id="fixedHttpMetaDataProvider" class="org.cloudfoundry.identity.uaa.provider.saml.FixedHttpMetaDataProvider" />
 
+    <bean id="defaultSamlConfig" class="org.cloudfoundry.identity.uaa.provider.saml.SamlConfigurationBean">
+        <property name="signatureAlgorithm" value="${login.saml.signatureAlgorithm:SHA1}"/>
```
Tallicia, line 305: Is this to be SHA256?

hsinn0: It is to be whatever you configure in TAS setting and what is in uaa.yml.

Tallicia (approved):
This looks approvable, though the SHA1 usage instead of SHA256 in the saml-providers.xml I'd like to understand why.

uaa/src/main/webapp/WEB-INF/spring/saml-providers.xml
torsten-sap: Same thought here :) I guess the state was SHA-1 before but would be good to have a plan to adapt a secure default.

hsinn0:
> the SHA1 usage instead of SHA256 in the saml-providers.xml I'd like to understand why.

The bean definition is the same as the old one we took out by mistake.

`value="${login.saml.signatureAlgorithm:SHA1}"` means that we set the value to `login.saml.signatureAlgorithm` property value if it exists, but set to "SHA1" if not. That is, it is the fallback default value.

uaa/src/main/webapp/WEB-INF/spring/saml-providers.xml
Tallicia: Should the default value be SHA256. the default being SHA1 is not secure. This is not terrible since it is overridden, but the default should be closer to SHAKE256 or SHA3-512

Tallicia:
> > the SHA1 usage instead of SHA256 in the saml-providers.xml I'd like to understand why.
> 
> The bean definition is the same as the old one we took out by mistake.
> 
> `value="${login.saml.signatureAlgorithm:SHA1}"` means that we set the value to `login.saml.signatureAlgorithm` property value if it exists, but set to "SHA1" if not. That is, it is the fallback default value.

Following this fix we should consider ensuring our defaults are as secure as possible: https://en.wikipedia.org/wiki/Secure_Hash_Algorithms

swalchemist:
We considered changing the default to SHA256, but decided to keep the same default as before. Note that TAS will always set it to SHA256 or SHA512, so we do not expect this default to have any effect in TAS. It is true that a default of SHA256 would have made the bug less severe and we might want to consider changing that soon.

uaa/src/main/webapp/WEB-INF/spring/saml-providers.xml
hsinn0: In production, SHA1 was never supposed to be used. In OpsManager, you can set the value to SHA256 or SHA512 only. SHA1 was exposed only because we removed the bean by mistake.

Tallicia: @torsten-sap Is this a PR you can approve as acceptable?

PR #2794, opened 2024-03-21
backfill tests: SAML SP metadata

peterhaochen47:
- in preparation for replacing the EOL spring saml extension lib with spring security core saml, adding more test coverage on the SAML SP metadata
- tests that SAML SP metadata matches the UAA configs (in the context of this test, the UAA configs are from the local uaa.yml used to start a local server)
- also explicitly declare some SAML-SP-related fields in the said local uaa.yml to make the inputs to the test clearer

[#186986697]

### DIFF
--- a/uaa/src/main/resources/uaa.yml
+++ b/uaa/src/main/resources/uaa.yml
@@ -405,9 +405,9 @@ login:
     #Local/SP metadata - requests signed
     signRequest: true
     #Local/SP metadata - want incoming assertions signed
-    #wantAssertionSigned: true
+    wantAssertionSigned: true
     #Algorithm for SAML signatures. Defaults to SHA1.  Accepts SHA1, SHA256, SHA512
-    #signatureAlgorithm: SHA256
+    signatureAlgorithm: SHA256
     socket:
       # URL metadata fetch - pool timeout
       connectionManagerTimeout: 10000
--- a/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/SamlLoginIT.java
+++ b/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/SamlLoginIT.java
@@ -12,9 +12,7 @@
  *******************************************************************************/
 package org.cloudfoundry.identity.uaa.integration.feature;
 
-import java.io.BufferedReader;
 import java.io.IOException;
-import java.io.InputStreamReader;
 import java.net.HttpURLConnection;
 import java.net.URL;
 import java.net.URLEncoder;
@@ -29,6 +27,7 @@ import org.springframework.beans.factory.annotation.Value;
 import org.springframework.http.HttpEntity;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpMethod;
+import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.security.oauth2.client.test.TestAccounts;
 import org.springframework.security.oauth2.common.util.RandomValueStringGenerator;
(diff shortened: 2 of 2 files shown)


strehle (approved):
wouldn't it make sense to use default SHA256 also here
https://github.com/cloudfoundry/uaa/blob/develop/uaa/src/main/webapp/WEB-INF/spring/saml-providers.xml#L305

hsinn0:
> wouldn't it make sense to use default SHA256 also here https://github.com/cloudfoundry/uaa/blob/develop/uaa/src/main/webapp/WEB-INF/spring/saml-providers.xml#L305

Right. It was mentioned and considered before in a different PR. In production though, for TAS at least, the property is always set to either 256 or 512, and we decided just to leave it like that as it is practically only used for dev build.

uaa/src/main/resources/uaa.yml
```diff
@@ -405,9 +405,9 @@ login:
     #Local/SP metadata - requests signed
     signRequest: true
     #Local/SP metadata - want incoming assertions signed
-    #wantAssertionSigned: true
+    wantAssertionSigned: true
     #Algorithm for SAML signatures. Defaults to SHA1.  Accepts SHA1, SHA256, SHA512
-    #signatureAlgorithm: SHA256
+    signatureAlgorithm: SHA256
```
strehle, line 410: this changes the default for SAML signatures

strehle:
> > wouldn't it make sense to use default SHA256 also here https://github.com/cloudfoundry/uaa/blob/develop/uaa/src/main/webapp/WEB-INF/spring/saml-providers.xml#L305
> 
> Right. It was mentioned and considered before in a different PR. In production though, for TAS at least, the property is always set to either 256 or 512, and we decided just to leave it like that as it is practically only used for dev build.

even in DEV it is no 256 as default... but therefore I would not let in xml sha1 with a settings where is looks like this is the default, e.g. https://github.com/cloudfoundry/uaa/blob/develop/uaa/src/main/webapp/WEB-INF/spring/saml-providers.xml#L305 because the value is no not null thus default in XML is never used

peterhaochen47:
> > wouldn't it make sense to use default SHA256 also here https://github.com/cloudfoundry/uaa/blob/develop/uaa/src/main/webapp/WEB-INF/spring/saml-providers.xml#L305
> 
> Right. It was mentioned and considered before in a different PR. In production though, for TAS at least, the property is always set to either 256 or 512, and we decided just to leave it like that as it is practically only used for dev build.

In [cf-deployment](https://github.com/cloudfoundry/cf-deployment/blob/7438ef8583de7cc4c9a637b90278f0fd469a17d7/cf-deployment.yml) (the OSS version of CF), this value is unset, hence UAA default will be used.

uaa/src/main/resources/uaa.yml
hsinn0: Yes, we acknowledge that. Do you think that we should not change this?

strehle:
> (the OSS version of CF), this value is unset, hence UAA default will be used.

I bet that even in cf-deployment now SHA256 is used... but I have no problem with it... only because of misleading XML default which is no default anymore..

See 
curl -k https://login.uaa-acceptance.cf-app.com/saml/metadata |grep sha

Then merge this PR and check again the signature of https://login.uaa-acceptance.cf-app.com

hsinn0:
> > > wouldn't it make sense to use default SHA256 also here https://github.com/cloudfoundry/uaa/blob/develop/uaa/src/main/webapp/WEB-INF/spring/saml-providers.xml#L305
> > 
> > 
> > Right. It was mentioned and considered before in a different PR. In production though, for TAS at least, the property is always set to either 256 or 512, and we decided just to leave it like that as it is practically only used for dev build.
> 
> In [cf-deployment](https://github.com/cloudfoundry/cf-deployment/blob/7438ef8583de7cc4c9a637b90278f0fd469a17d7/cf-deployment.yml) (the OSS version of CF), this value is unset, hence UAA default will be used.

Right, but the default has been always SHA1 there, as the property has been like that all the time. Do you think we should change the default value there for the OSS version of CF? Then we can make the change in another PR.

strehle:
> default value there for the OSS version of CF? Then we can make the change in another PR

ok

uaa/src/main/resources/uaa.yml
strehle: No fine for all... SHA1 would be a risk / issue nowadays

hsinn0: Ah, yes, I forgot about this. I was thinking about the other previous PR.

Issue #2738, opened 2024-02-21
LdapSkipCertificateMockMvcTests test failures in local runs, introduced by PR 2711


swalchemist:
Using the head of the develop branch - when I run all of the uaa unit tests on my machine like this: `./gradlew test`, both tests in the `LdapSkipCertificateMockMvcTests` class (in the `server` module) consistently fail in the setup:

```
LdapSkipCertificateMockMvcTests > testIgnoreExpiredServerCertificate() FAILED
    java.lang.AssertionError: Status expected:<200> but was:<401>
        at org.springframework.test.util.AssertionErrors.fail(AssertionErrors.java:59)
        at org.springframework.test.util.AssertionErrors.assertEquals(AssertionErrors.java:122)
        at org.springframework.test.web.servlet.result.StatusResultMatchers.lambda$matcher$9(StatusResultMatchers.java:627)
        at org.springframework.test.web.servlet.MockMvc$1.andExpect(MockMvc.java:214)
        at org.cloudfoundry.identity.uaa.mock.util.MockMvcUtils.getClientCredentialsOAuthAccessToken(MockMvcUtils.java:1213)
        at org.cloudfoundry.identity.uaa.mock.util.MockMvcUtils.getClientCredentialsOAuthAccessToken(MockMvcUtils.java:1188)
        at org.cloudfoundry.identity.uaa.mock.util.MockMvcUtils.createOtherIdentityZoneAndReturnResult(MockMvcUtils.java:555)
        at org.cloudfoundry.identity.uaa.mock.util.MockMvcUtils.createOtherIdentityZoneAndReturnResult(MockMvcUtils.java:627)
        at org.cloudfoundry.identity.uaa.mock.util.MockMvcUtils.createOtherIdentityZoneAndReturnResult(MockMvcUtils.java:636)
        at org.cloudfoundry.identity.uaa.mock.ldap.LdapSkipCertificateMockMvcTests.setUp(LdapSkipCertificateMockMvcTests.java:70)
```

I have not seen these test fail on Concourse, only locally. Similar result, though, when I run the tests the way we run them in CI:

```
./gradlew clean
./gradlew -Dspring.profiles.active=hsqldb,default -Djava.security.egd=file:/dev/./urandom assemble --max-workers=4 --no-daemon --stacktrace --console=plain --exclude-task :cloudfoundry-identity-samples:assemble
[...]

PR #2801, opened 2024-03-26
fix: gradle test might give false green

peterhaochen47:
- when failFast = true AND retry is on, there is a serious issue: gradle might stop the test run due to the failFast but still concludes with BUILD SUCCESSFUL (if the retry is successful)
  - we have observed this "false green" and this commit fixes this. 
- impact: this might cause builds with many failing tests to take longer to finish.

[#186986697]

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -86,7 +86,9 @@ subprojects {
 
     test {
         maxParallelForks = 1
-        failFast = true
+        // when failFast = true AND retry is on, there is a serious issue:
+        // gradle might stop the test run due to the failFast but still concludes with BUILD SUCCESSFUL (if the retry is successful)
+        failFast = false
         useJUnitPlatform()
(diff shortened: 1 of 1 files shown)


strehle:
> * when failFast = true AND retry is on, there is a serious issue: gradle might stop the test run due to the failFast but still concludes with BUILD SUCCESSFUL (if the retry is successful)
>   
>   * we have observed this "false green" and this commit fixes this.
> * impact: this might cause builds with many failing tests to take longer to finish.
> 
> [#186986697]

Ok, I run tests in a loop and yes there are some tests which are flaky and these tests will appear to be red then more often.

e.g. Ldap test, e.g. https://github.com/cloudfoundry/uaa/issues/2738 or https://github.com/cloudfoundry/uaa/issues/2802

strehle (approved):

swalchemist:
@strehle There seems to be a bad interaction between the retry feature and failFast, where a test fails once and then passes on retry - failFast sees that as a failure and stops the run, and retry properly counts it as a passing test and the overall test run is reported as passing. But if there are other tests that are failing but didn't run, then we have a false green, and the test run should have been reported as failing overall.

If this means we report failures more often, that's a good thing.

PR #2803, opened 2024-03-27
Prevent Update and Delete of Entities with Alias if Alias Feature is disabled

adrianhoelzl-sap:
see issue #2505 

Prevent update and deletion of identity providers that have an alias if the alias feature is disabled again (i.e., `aliasEntitiesEnabled` is set to `false`).

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/alias/EntityAliasHandler.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/alias/EntityAliasHandler.java
@@ -225,6 +225,9 @@ public abstract class EntityAliasHandler<T extends EntityWithAlias> {
     protected abstract T cloneEntity(final T originalEntity);
 
     public final Optional<T> retrieveAliasEntity(final T originalEntity) {
+        if (!hasText(originalEntity.getAliasId()) || !hasText(originalEntity.getAliasZid())) {
+            return Optional.empty();
+        }
         return retrieveEntity(originalEntity.getAliasId(), originalEntity.getAliasZid());
     }
 
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/alias/EntityAliasHandler.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/alias/EntityAliasHandler.java
@@ -37,8 +37,8 @@ public abstract class EntityAliasHandler<T extends EntityWithAlias> {
         final boolean entityAlreadyHasAlias = existingEntity != null && hasText(existingEntity.getAliasZid());
         if (entityAlreadyHasAlias) {
             if (!aliasEntitiesEnabled) {
-                // if the feature is disabled, we only allow setting both alias properties to null
-                return !hasText(requestBody.getAliasId()) && !hasText(requestBody.getAliasZid());
+                // reject ANY update of an entity with an existing alias if the feature is disabled
+                return false;
             }
 
             if (!hasText(existingEntity.getAliasId())) {
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/alias/EntityAliasHandler.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/alias/EntityAliasHandler.java
@@ -115,6 +115,10 @@ public abstract class EntityAliasHandler<T extends EntityWithAlias> {
      *                                    'aliasZid' does not exist
      * @throws EntityAliasFailedException if 'aliasId' and 'aliasZid' are set in the original entity, but the
      *                                    referenced alias entity could not be found
+     * @throws IllegalStateException      if {@code existingEntity} has an alias and 'aliasEntitiesEnabled' is set to
+     *                                    {@code false}
+     * @throws IllegalStateException      if a new alias is about to be created, i.e., {@code originalEntity} has a
+     *                                    non-empty 'aliasZid', and 'aliasEntitiesEnabled' is set to {@code false}
      */
     public final T ensureConsistencyOfAliasEntity(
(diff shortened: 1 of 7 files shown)


uaa/slateCustomizations/source/index.html.md.erb
```diff
@@ -1173,10 +1173,11 @@ _Request and Response Fields_
...
 
-| Error Code | Description                                                           |
-|------------|-----------------------------------------------------------------------|
-| 403        | Forbidden - Insufficient scope                                        |
-| 422        | Unprocessable Entity - Invalid config                                 |
+| Error Code | Description                                                                          |
+|------------|--------------------------------------------------------------------------------------|
+| 403        | Forbidden - Insufficient scope                                                       |
+| 422        | Unprocessable Entity - Invalid config                                                |
+| 422        | Unprocessable Entity - updating IdP with alias while `aliasEntitiesEnabled` is false |
```
adrianhoelzl-sap: Should we merge the two rows with the `422` status code or is it okay if we keep them separated?

server/src/main/java/org/cloudfoundry/identity/uaa/alias/EntityAliasHandler.java
```diff
@@ -121,53 +120,22 @@ public final T ensureConsistencyOfAliasEntity(
...
-                                existingEntity.getAliasDescription()
-                        ), HttpStatus.UNPROCESSABLE_ENTITY.value(), e
-                );
-            }
-
-            // no change required in the original entity since its aliasId and aliasZid were already set to null
-            return originalEntity;
+        if (entityHadAlias && !aliasEntitiesEnabled) {
+            // this should already be caught in the validation method
+            throw new IllegalStateException("Performing update on entity with alias while alias feature is disabled.");
```
hsinn0, line 130: I suggest to document this exception in this method's JavaDoc, i.e. add @throws IllegalStateExceptio with an explanation. Or, it appears that we can also throw EntityAliasFailedException instead for this case.

server/src/main/java/org/cloudfoundry/identity/uaa/alias/EntityAliasHandler.java
```diff
@@ -121,53 +120,22 @@ public final T ensureConsistencyOfAliasEntity(
...
         }
 
         if (!hasText(originalEntity.getAliasZid())) {
             // no alias handling is necessary
             return originalEntity;
         }
 
+        if (!aliasEntitiesEnabled) {
+            // this should already be caught in the validation method
+            throw new IllegalStateException("Trying to create a new alias while alias feature is disabled.");
```
hsinn0, line 140: The same as above.

server/src/test/java/org/cloudfoundry/identity/uaa/alias/EntityAliasHandlerValidationTest.java
```diff
@@ -124,16 +124,10 @@ final void shouldReturnFalse_NotBothAliasPropsEmptyInReqBody() {
...
-        final void shouldReturnTrue_BothAliasPropsEmptyInReqBody() {
-            final T existingEntity = buildEntityWithAliasProps(
-                    UUID.randomUUID().toString(),
-                    CUSTOM_ZONE_ID
-            );
-            final T requestBody = buildEntityWithAliasProps(null, null);
-            assertThat(aliasHandler.aliasPropertiesAreValid(requestBody, existingEntity)).isTrue();
+            // (9) alias ID removed, alias ZID removed
+            requestBody = buildEntityWithAliasProps(null, null);
+            assertThat(aliasHandler.aliasPropertiesAreValid(requestBody, existingEntity)).isFalse();
```
hsinn0, line 130: Not that it really matters to me, but just curious. Is there any specific reason that we use `org.assertj.core.api.Assertions` instead of `org.junit.jupiter.api.Assertions` in this class?

uaa/slateCustomizations/source/index.html.md.erb
hsinn0: My preference is merging them.

server/src/main/java/org/cloudfoundry/identity/uaa/alias/EntityAliasHandler.java
adrianhoelzl-sap: fixed with commit ead4176

server/src/main/java/org/cloudfoundry/identity/uaa/alias/EntityAliasHandler.java
adrianhoelzl-sap: fixed with commit ead4176

adrianhoelzl-sap: I decided to use `IllegalStateException`s for those cases that should not occur when the `aliasPropertiesAreValid` returned `true`.

uaa/slateCustomizations/source/index.html.md.erb
adrianhoelzl-sap: fixed with commit eca31fd

server/src/test/java/org/cloudfoundry/identity/uaa/alias/EntityAliasHandlerValidationTest.java
adrianhoelzl-sap: In general, I like AssertJ more as they offer many convenient assertion methods for commonly used types (`Collection`s, `Optional`, etc.). Also, they allow chaining of several assertions being performed on the same value.

hsinn0 (approved):

PR #2806, opened 2024-03-28
Move OAuth2 classes BaseClientDetails to UaaClientDetails

### LINKED ISSUE #2560: Spring Security Oauth2 End of Life

strehle:
Move the classes because finally we want remove https://github.com/cloudfoundry/uaa/blob/develop/dependencies.gradle#L14 

The classes here represent the persistence of UAA. Moved only the classes but not the interfaces yet. 
The interfaces used in many other library methods but finally the interfaces will be moved also next to the classes here

The BaseClientDetails class does not know jwtClientConfig field of UAA but now UaaClientDetails class know 
Added tests for the base classes and changed many imports


strehle:
Sonar
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2806

model/src/main/java/org/cloudfoundry/identity/uaa/client/Jackson2ArrayOrStringDeserializer.java
```diff
@@ -0,0 +1,40 @@
...
+import com.fasterxml.jackson.databind.type.SimpleType;
+import org.springframework.util.StringUtils;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+@SuppressWarnings("serial")
+public class Jackson2ArrayOrStringDeserializer extends StdDeserializer<Set<String>> {
```
hsinn0, line 25: I can see that this class and the above one, and probably many other classes in this PR are re-implementation of what is in `spring-security-oauth2`. However, I had to look into `spring-security-oauth2` to figure that out. I think that documenting about that in the source code a little bit will be helpful.

hsinn0: OK, after going through all the files in this PR, I realize that this PR is indeed all about replacing the classes in `spring-security-oauth2`. So a little bit of Javadoc in the new class sources will be definitely help people who read the codes later. That is all.

hsinn0 (changes requested):

model/src/main/java/org/cloudfoundry/identity/uaa/client/Jackson2ArrayOrStringDeserializer.java
strehle: This PR was independent so far from rest.
I tried to do similar things with Exceptions and OAuth2RestTemplate but UAA and spring-security-oauth2 are so deep connected that it does not work. This means the next https://github.com/cloudfoundry/uaa/pull/2813 will be then the final one - of course without dependency to spring-security-oauth2. But therefore I moved all classes into on namespace with UAA so that it gets visible what classes we need

hsinn0: Cool. It appears that documentations were added to the classes. Thanks.

hsinn0 (approved):

PR #2816, opened 2024-04-05
Sonar fix

strehle:
Found with https://github.com/cloudfoundry/uaa/pull/2813

Prefix it

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/TokenRevocationEndpoint.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/TokenRevocationEndpoint.java
@@ -2,6 +2,7 @@ package org.cloudfoundry.identity.uaa.oauth;
 
 import org.cloudfoundry.identity.uaa.audit.event.SystemDeletable;
 import org.cloudfoundry.identity.uaa.authentication.UaaPrincipal;
+import org.cloudfoundry.identity.uaa.logging.SanitizedLogFactory;
 import org.cloudfoundry.identity.uaa.oauth.client.ClientConstants;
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/cloudfoundry/identity/uaa/oauth/TokenRevocationEndpoint.java
```diff
@@ -29,26 +28,26 @@
...
 
 import static org.springframework.http.HttpStatus.OK;
-import static org.springframework.web.bind.annotation.RequestMethod.DELETE;
-import static org.springframework.web.bind.annotation.RequestMethod.GET;
 
 @Controller
 public class TokenRevocationEndpoint implements ApplicationEventPublisherAware {
 
-    protected final Logger logger = LoggerFactory.getLogger(getClass());
+    protected final SanitizedLogFactory.SanitizedLog logger = SanitizedLogFactory.getLog(getClass());
```
swalchemist, line 44: I'm curious what Sonar issue this fixes? Is there whitespace in the log output?

Should every uaa class use this?

strehle: missing input validation, can be solved with
https://github.com/cloudfoundry/uaa/blame/develop/server/src/main/java/org/cloudfoundry/identity/uaa/logging/LogSanitizerUtil.java#L16

Similar to other things, e.g. IdentityZoneHolder, yes we can do a replacement everywhere . I did it here because it was clear from sonar. Other exchanges might be without effect

swalchemist (approved):

PR #2828, opened 2024-04-15
fix: load static resources from default zone if zone not found

tack-sap:
This was already addressed with #979, but the fix there only works for a local setup. In a typical productive deployment, the path is not starting with /uaa/, so the check need to be adopted to work in both local environment and a deployment.


strehle:
Now
https://none.login.uaa-acceptance.cf-app.com/

image

After the goal is, show error 

image

strehle (approved):

strehle:
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2828

PR #2831, opened 2024-04-16
Show an example of how to produce SAML Metadata using an endpoint

fhanik:
Using a Spring Controller instead of a filter. Because Filters are complicated and annoying


Tallicia:
Closing as this has been incorporated locally, Thanks @fhanik!

PR #2843, opened 2024-04-19
 Move tests from spring-security-oauth2 into UAA

strehle:
Test coverage should help to identity if we need the coding or not.
Coding, which is not covered should be checked in Unit and Integration Tests
and should be removed if not really used.


bruce-ricard (approved):
If you're just adding a bunch of tests, and they are all passing I don't know what else to do but approve. Especially since you are not even merging into `develop`.

strehle:
Ok, reason for this PR is, that @swalchemist can check code and test in extra PRs, 

https://github.com/cloudfoundry/uaa/pull/2813
vs.
https://github.com/cloudfoundry/uaa/pull/2848

This PR here is the DIFF between both. Simply to have a bit smaller PR in 2813.

PR #1462, opened 2020-12-05
Bump version spring oauth 2.5.0.RELEASE

strehle:

PR #2845, opened 2024-04-21
test-refactoring: remove forked class in tests 

strehle:
Forked because of 

https://github.com/cloudfoundry/uaa/pull/1462 

We have now AlphanumericRandomValueStringGenerator and will get the other one.

For tests go with AlphanumericRandomValueStringGenerator

Prevent problems with https://github.com/cloudfoundry/uaa/pull/2813 where we now have the same class also in UAA context


swalchemist:
I'm having trouble understanding the context. The code changes are easy enough to review on the surface. @strehle could you explain the motivation?

Why were some RandomValueStringGenerator uses replaced but not all?

strehle:
> I'm having trouble understanding the context. The code changes are easy enough to review on the surface. @strehle could you explain the motivation?
> 
> Why were some RandomValueStringGenerator uses replaced but not all?

The productive usage of this class is in context of OAuth2 and this stays. 

This PR is only for test usages:
Motivation is to remove a mix of RandomValueStringGenerator and AlphanumericRandomValueStringGenerator usages.

History:
RandomValueStringGenerator comes from spring-security-oauth2 (https://github.com/spring-attic/spring-security-oauth/blob/main/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/RandomValueStringGenerator.java) and created random. In the past without extra characters, but with latest updates 2.5.1 , added -_

This new chars caused many test failures in UAA because we used the random for client / user creation.... and many test scenarios. Therefore I forked the original class from ext. project into UAA test scope, see
https://github.com/cloudfoundry/uaa/blob/develop/uaa/src/test/java/org/cloudfoundry/identity/uaa/login/util/RandomValueStringGenerator.java

Later there was the new class https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/util/AlphanumericRandomValueStringGenerator.java because others in UAA had similar issues. 

Now (back in present) we want remove the spring security oauth2 dependency but the  RandomValueStringGenerator  still is in test, therefore this PR is a cleanup to remove this fork and re-use the AlphanumericRandomValueStringGenerator in places where we need random without special chars.
In the big move https://github.com/cloudfoundry/uaa/pull/2813 the original class RandomValueStringGenerator  comes again to UAA but this time with its current behavior, means with extra characters which is needed for OAuth2 code creation.

You may see that all changes are only in TEST code, therefore this extra cleanup PR because I see it as a easy to approve PR and thought its better to have it in a extra PR and not within https://github.com/cloudfoundry/uaa/pull/2813

swalchemist (approved):
Thanks for splitting out this PR.

PR #2847, opened 2024-04-23
fix: only convert client auth method root claim

mikeroda:
Minor correction to apply the desired conversion from emtpy to none only to the client auth method root claim.


server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java
```diff
@@ -491,8 +491,11 @@ private CompositeToken createCompositeToken(String tokenId,
 
     private static Map<String, Object> addRootClaimEntry(Map<String, Object> additionalRootClaims, String entry, String value) {
         Map<String, Object> claims = additionalRootClaims != null ? additionalRootClaims : new HashMap<>();
-        // set externally none as client_auth_method if internally empty
-        claims.put(entry, CLIENT_AUTH_EMPTY.equals(value) ? CLIENT_AUTH_NONE : value);
+        if (CLIENT_AUTH_METHOD.equals(entry)) {
```
strehle, line 494: from this code I am not sure, what is the issue , so can you please create a test to see, what is before and after behaviour

Issue #2821, opened 2024-04-11
Fix performance issue with external identity provider lookup [OIDC]


strehle:
### What version of UAA are you running?

Develop, latest UAA

What output do you see from `curl <YOUR_UAA>/info -H'Accept: application/json'`?

### How are you deploying the UAA?

I am deploying the UAA

- locally only using gradlew
- using a bosh release I downloaded from bosh.io
- using cf-release
- using cf-deployment

### What did you do?

1. Add many external SAML or OIDC to an identity zone ( > 10.000)
2. Perform a SAML / OIDC login
3. Check login times / DB metrics

### What did you expect to see? What goal are you trying to achieve with the UAA?

Login < 1s , without memory and/or DB issues

### What did you see instead?

With SAML there are memory issues, with OIDC mainly DB issues.
Why:
* SAML delegates the lookup from entiyID (external key or the SAML assertion) to spring-security-saml and in UAA there is a cache but if there are many entries there is a memory problem, e.g. https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/LoginSamlAuthenticationProvider.java#L129 reads all saml providers from DB and resolves then the needed one from SAML message (entityID)

* OIDC similar readAll and filter in code, e.g. https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthProviderConfigurator.java#L153-L158

This readALL pattern should be prevented and identy_provider DB should have a field like external_key (type string) with an index on it. This should solve the lookup from external token to UAA IdP.
External_key should contain entityID in case of SAML and issuer in case of OIDC/OAUTH

strehle:
@hsinn0 see https://github.com/cloudfoundry/uaa/issues/2825#issuecomment-2160296620 
I will start on a fix soon, plan is
a) add externalKey as new column into identity_provider
b) add index for externalKey and zone_id to allow a lookup from a token to the IDP.

-> externalKey is issue in case of OIDC and should be entityID in case of SAML

PR #2850, opened 2024-04-24
Reject IdZ deletion if an IdP with alias exists in the zone

### LINKED ISSUE #2505: Missing user management from custom IdZ to CC

adrianhoelzl-sap:
see issue #2505 

Reject deletion if an IdP with alias exists in the zone to be deleted. With this restriction, we want to avoid that callers of the endpoint accidentally remove IdPs (and thereby the associated users) from the alias zone. Now the callers must actively delete the IdPs with alias beforehand (which will also remove all associated users, their alias users and the alias IdP in the "uaa" zone).


strehle:
Sonar issues should be solved @adrianhoelzl-sap 
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2663

bruce-ricard:
This appears to be a pretty major feature change. It looks like we are trying to merge this to develop, which would imply that it's going to be shipped in a new minor release in a few weeks.

It seems to me like this should be in a major release. I am uncertain of the ramifications of this change. The code was clearly set to do the opposite of what you are trying to do.

strehle:
I had not checked the logic before only code style and sonar issues, but now I see the impact because @bruce-ricard comments.

So I do not see a need for this PR. If you have the power to delete a zone , then we should not prevent it because some values in the zone are not as expected.

I do not understand the relation to mentioned issue https://github.com/cloudfoundry/uaa/issues/2505 .

uaa/slateCustomizations/source/index.html.md.erb
```diff
@@ -871,6 +871,7 @@ _Error Codes_
 | 401        | Unauthorized - Invalid token                                                |
 | 403        | Forbidden - Insufficient scope (zone admins can only delete their own zone) |
 | 404        | Not Found - Zone does not exist                                             |
+| 422        | Unprocessable Entity - at least one IdP with alias exists in the zone       |
```
Tallicia, line 874: good choice for response

server/src/test/java/org/cloudfoundry/identity/uaa/zone/IdentityZoneEndpointsTests.java
```diff
@@ -169,6 +186,53 @@ void reduce_zone_allowed_groups_on_update_should_fail() throws InvalidIdentityZo
...
+        final ResponseEntity<IdentityZone> response = endpoints.deleteIdentityZone(idzId);
+        assertNotNull(response);
+        assertEquals(HttpStatus.OK, response.getStatusCode());
+
+        final ArgumentCaptor<EntityDeletedEvent<IdentityZone>> eventArgument = ArgumentCaptor.forClass(EntityDeletedEvent.class);
+        verify(mockApplicationEventPublisher).publishEvent(eventArgument.capture());
+        final var capturedEvent = eventArgument.getValue();
+        assertEquals(idz, capturedEvent.getDeleted());
+    }
+
```
Tallicia, line 224: Both tests look good

Tallicia (approved):
Looks good

Tallicia (dismissed):
Un-approving this following review of the comments. The change appears to break existing behavior. I thought it was a fix, not a change.

server/src/main/java/org/cloudfoundry/identity/uaa/zone/IdentityZoneEndpoints.java
```diff
@@ -336,7 +336,18 @@ public ResponseEntity<IdentityZone> deleteIdentityZone(@PathVariable String id)
             IdentityZone zone = zoneDao.retrieveIgnoreActiveFlag(id);
             // ignore the id in the body, the id in the path is the only one that matters
             IdentityZoneHolder.set(zone);
-            if (publisher != null && zone != null) {
+
+            /* reject deletion if an IdP with alias exists in the zone - checking for users with alias is not required
+             * here, since they can only exist if their origin IdP has an alias as well */
+            final List<IdentityProvider> idps = idpDao.retrieveAll(false, zone.getId());
```
strehle: here is the breaking change.

beside this, please use a another approach to check all IdPs, see https://github.com/cloudfoundry/uaa/issues/2821 . 
If in a zone many idps installed, this can be slow.

strehle: Checked again and I think we should add here check for aliasEntitiesEnabled = true then you can do this extra check and return 422.

Btw. I search a while for aliasEntitiesEnabled in https://github.com/cloudfoundry/uaa-release , e.g. https://github.com/cloudfoundry/uaa-release/blob/develop/spec/input/all-properties-set.yml for this parameter and its documentation, but did not found it.
So we should add it to https://github.com/cloudfoundry/uaa-release/blob/develop/spec/input/all-properties-set.yml and with this parameter we can do some flows in UAA different ,
But for all CF usages without alias entry in IdP or User this SELECT is done

uaa/slateCustomizations/source/index.html.md.erb
strehle, line 874: Unprocessable Entity (e.g., deleting IdP with alias while aliasEntitiesEnabled is true)

Found similar fragments in https://docs.cloudfoundry.org/api/uaa/version/77.8.0/index.html#delete

server/src/main/java/org/cloudfoundry/identity/uaa/provider/JdbcIdentityProviderProvisioning.java
```diff
@@ -40,7 +40,7 @@ public class JdbcIdentityProviderProvisioning implements IdentityProviderProvisi
 
     public static final String DELETE_IDENTITY_PROVIDER_BY_ORIGIN_SQL = "delete from identity_provider where identity_zone_id=? and origin_key = ?";
 
-    public static final String DELETE_IDENTITY_PROVIDER_BY_ZONE_SQL = "delete from identity_provider where identity_zone_id=? or alias_zid=?";
+    public static final String DELETE_IDENTITY_PROVIDER_BY_ZONE_SQL = "delete from identity_provider where identity_zone_id=?";
```
strehle, line 46: see now that this is a fix but the issue was 
https://github.com/cloudfoundry/uaa/pull/2637/files#diff-8e7d64a1881ae78ed33d745ed1fd3a922312b89e740a3e1c75d1284b6a041b31R43

strehle (commented):
I know https://github.com/cloudfoundry/uaa/issues/2505 but during the review process we defined that this new feature should come with extra options, but for all others UAA should not change.

We have an option for this, e.g. aliasEntitiesEnabled  and therefore with this aliasEntitiesEnabled = false there should not be any differences to UAAs before

server/src/main/java/org/cloudfoundry/identity/uaa/zone/IdentityZoneEndpoints.java
adrianhoelzl-sap: We must also perform this check when `aliasEntitiesEnabled` is `false`, since it is possible to enable it, create some entities with alias and then disable it again. At this point, there would still be entities with alias left in the DB which should block the deletion of the zone.

uaa/slateCustomizations/source/index.html.md.erb
adrianhoelzl-sap: see my comment above - here, the flag is not relevant

server/src/main/java/org/cloudfoundry/identity/uaa/provider/JdbcIdentityProviderProvisioning.java
hsinn0: Was this concern from @strehle ever addressed?

server/src/main/java/org/cloudfoundry/identity/uaa/provider/JdbcIdentityProviderProvisioning.java
```diff
@@ -56,6 +60,19 @@ public JdbcIdentityProviderProvisioning(JdbcTemplate jdbcTemplate) {
...
+                new Object[]{zoneId},
+                new int[]{VARCHAR},
+                Integer.class
+        );
+        if (numberOfIdpsWithAlias == null) {
+            throw new IncorrectResultSizeDataAccessException(1);
+        }
+        return numberOfIdpsWithAlias > 0;
+    }
+
```
hsinn0, line 73: Why do you do count query instead of exists query here?

server/src/main/java/org/cloudfoundry/identity/uaa/zone/IdentityZoneEndpoints.java
hsinn0: > please use a another approach to check all IdPs, see https://github.com/cloudfoundry/uaa/issues/2821

Was this addressed?

hsinn0: It sounds like that both sides think that this should be considered as breaking change. So per our previous agreement, this should be part of major version bump. I see that it would only matter if the aliasId feature was being used. I guess that the aliasId feature was already released?

hsinn0: I got further context info regarding this breaking change from other members in the team. As I mentioned earlier, "it would only matter if the aliasId feature was being used."
So when considering the scope of this PR only, because I also heard that SAP is OK including this change in non-major version bump, I see no reason to reject that from our side anymore. I think that you can phrase this change whatever way you want as needed, e.g. experimental feature, etc., maybe in release notes.
So including this in non major UAA version update is fine, but address other concerns in my other review comments.

server/src/main/java/org/cloudfoundry/identity/uaa/provider/JdbcIdentityProviderProvisioning.java
adrianhoelzl-sap: We revert the SQL back to the state it was in before the alias feature for identity providers was integrated. This means that during the deletion of a zone, we no longer delete all identity providers in other zones that have an alias to the deleted zone.
Instead, we reject the deletion of the zone if identity providers with an alias exist in the zone to be deleted and thereby force the administrator to delete these IdPs beforehand (which will then also delete the users associated to the original and alias IdPs).

server/src/main/java/org/cloudfoundry/identity/uaa/provider/JdbcIdentityProviderProvisioning.java
adrianhoelzl-sap: I tried to use `exists` in two different ways:

1. `VALUES (EXISTS (SELECT ...))` (see https://github.com/cloudfoundry/uaa/pull/2850/commits/e19a904e6cfa922537175df2128ac2be491ee8db#diff-8e7d64a1881ae78ed33d745ed1fd3a922312b89e740a3e1c75d1284b6a041b31R39) &rarr; supported by PostgreSQL and HSQL, but not MySQL
2. `SELECT EXISTS (SELECT ...)` &rarr; supported by PostgreSQL and MySQL, but not HSQL

Therefore, I decided to use the `count` function.

adrianhoelzl-sap: Should I add a database migration that creates an index for the `alias_zid` column?

hsinn0: > I tried to use `exists` in two different ways:
> ...
> Therefore, I decided to use the `count` function.

I see. HSQL is the hindrance here. In that case, how about using `limit` clause as follows?
`select 1 from identity_provider idp where idp.identity_zone_id = ? and idp.alias_zid is not null and idp.alias_zid <> '' limit 1`
That would return 1 if the record exists or null otherwise. It should be better than `select count(*)`.

Then you can change your Java code to return as follows (no need for additional check for null):
`return 1 == idpWithAliasExists;`

adrianhoelzl-sap: Thanks for the hint, I changed the query in commit 7f52786. I also added indexes for the `alias_zid` column of the tables `identity_provider` and `users` in commit 780d9ff.

server/src/main/java/org/cloudfoundry/identity/uaa/provider/JdbcIdentityProviderProvisioning.java
```diff
@@ -56,6 +61,24 @@ public JdbcIdentityProviderProvisioning(JdbcTemplate jdbcTemplate) {
...
+                    new Object[]{zoneId},
+                    new int[]{VARCHAR},
+                    Integer.class
+            );
+        } catch (final EmptyResultDataAccessException e) {
+            return false;
+        }
+        if (idpWithAliasExists == null) {
+            return false;
+        }
```
hsinn0, line 78: This extra check for null is not needed. The next statement checks both.

hsinn0 (approved):
Approving, assuming other concerns from @strehle has been actually resolved (despite they are not shown as resolved in this PR).

strehle:
> Approving, assuming other concerns from @strehle has been actually resolved (despite they are not shown as resolved in this PR).

mainly resolved with the indexed search now @adrianhoelzl-sap , please ensure happy sonar
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2850

server/src/main/java/org/cloudfoundry/identity/uaa/provider/JdbcIdentityProviderProvisioning.java
```diff
@@ -56,6 +61,24 @@ public JdbcIdentityProviderProvisioning(JdbcTemplate jdbcTemplate) {
...
+        final Integer idpWithAliasExists;
+        try {
+            idpWithAliasExists = jdbcTemplate.queryForObject(
+                    IDP_WITH_ALIAS_EXISTS_QUERY,
+                    new Object[]{zoneId},
+                    new int[]{VARCHAR},
+                    Integer.class
+            );
+        } catch (final EmptyResultDataAccessException e) {
+            return false;
```
strehle: please test this part

server/src/main/java/org/cloudfoundry/identity/uaa/provider/JdbcIdentityProviderProvisioning.java
```diff
@@ -56,6 +61,24 @@ public JdbcIdentityProviderProvisioning(JdbcTemplate jdbcTemplate) {
         this.jdbcTemplate = jdbcTemplate;
     }
 
+    public boolean idpWithAliasExistsInZone(final String zoneId) {
+        final Integer idpWithAliasExists;
+        try {
+            idpWithAliasExists = jdbcTemplate.queryForObject(
```
strehle: please return from here, e.g. 

return jdbcTemplate.queryForObject( xxx ) == 1;

and omit the parts below, which are not covered by tests yet

server/src/main/java/org/cloudfoundry/identity/uaa/zone/IdentityZoneEndpoints.java
```diff
@@ -77,7 +77,7 @@ public class IdentityZoneEndpoints implements ApplicationEventPublisherAware {
     private ApplicationEventPublisher publisher;
 
     public IdentityZoneEndpoints(final IdentityZoneProvisioning zoneDao,
-                                 final @Qualifier("identityProviderProvisioning") IdentityProviderProvisioning idpDao,
+                                 final @Qualifier("identityProviderProvisioning") JdbcIdentityProviderProvisioning idpDao,
```
strehle: why not adding this to the interface ?

server/src/main/java/org/cloudfoundry/identity/uaa/provider/JdbcIdentityProviderProvisioning.java
```diff
@@ -56,6 +61,24 @@ public JdbcIdentityProviderProvisioning(JdbcTemplate jdbcTemplate) {
...
+                    IDP_WITH_ALIAS_EXISTS_QUERY,
+                    new Object[]{zoneId},
+                    new int[]{VARCHAR},
+                    Integer.class
+            );
```
hsinn0: Actually, I think it would be better, more efficient if you call `jdbcTemplate.queryForList()` instead and then just check whether the list is empty or not.

adrianhoelzl-sap: fixed with commit a09808b

server/src/main/java/org/cloudfoundry/identity/uaa/provider/JdbcIdentityProviderProvisioning.java
adrianhoelzl-sap: fixed with commit a09808b

server/src/main/java/org/cloudfoundry/identity/uaa/provider/JdbcIdentityProviderProvisioning.java
adrianhoelzl-sap: fixed with commit a09808b

server/src/main/java/org/cloudfoundry/identity/uaa/zone/IdentityZoneEndpoints.java
adrianhoelzl-sap: fixed with commit b77cecd

hsinn0 (approved):

PR #2857, opened 2024-04-29
Fix: performance issue in MySQL -- revert #2704

### LINKED ISSUE #2856: Excessive database load retrieving users when using MySQL with millions of users

The problem was found after upgrading from 77.1.0 to 77.3.0.

I am deploying the UAA as part of a commercial Cloud Foundry distribution.

### What did you do?

A customer has millions of users in a MySQL database in production. We're getting queries like this that are taking more than 10 minutes, and they're stacking up and bogging down the database, causing CF commands to time out. They originated from calls to cf's `/v3/users` endpoint.

> select count(*) from users u join identity_provider idp on u.origin = idp.origin_key and u.identity_zone_id = idp.identity_zone_id where idp.active is true and (((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((LOWER(u.id) = LOWER('...id here...') OR LOWER(u.id) [...]

This caused an outage, so they rolled UAA back to 77.1.0 and saw the database load return to normal. They don't have any issues in their testing environment that only has a few thousands of users.

We see that LOWER(u.id) is indexed for postgres, but we don't create an index for mysql.

bruce-ricard:
* Revert "Merge pull request #2704 from cloudfoundry/fix/optimize-ids-users-endpoint"

* This reverts commit 1324a81f8948250e993193f4be62dc5b9854965a, reversing changes made to b8826e1456b98ae179b85cab9e5261a64edc2d25.

* The original PR is suspected to cause the performance issues. We will try to bring this change back later.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/resources/jdbc/AbstractQueryable.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/resources/jdbc/AbstractQueryable.java
@@ -19,7 +19,7 @@ public abstract class AbstractQueryable<T> implements Queryable<T> {
 
     private NamedParameterJdbcTemplate namedParameterJdbcTemplate;
 
-    protected final JdbcPagingListFactory pagingListFactory;
+    private JdbcPagingListFactory pagingListFactory;
 
     protected RowMapper<T> rowMapper;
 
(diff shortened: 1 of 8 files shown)


Tallicia:
Merging to get through pipeline.

PR #2859, opened 2024-04-30
fix: MySQL Performance Issues in "/ids/Users" Endpoint

adrianhoelzl-sap:
see issue #2856

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
@@ -12,19 +12,32 @@
  *******************************************************************************/
 package org.cloudfoundry.identity.uaa.scim.jdbc;
 
-import org.cloudfoundry.identity.uaa.util.UaaStringUtils;
-import org.cloudfoundry.identity.uaa.zone.IdentityZone;
-import org.cloudfoundry.identity.uaa.zone.JdbcIdentityZoneProvisioning;
-import org.cloudfoundry.identity.uaa.zone.UserConfig;
-import org.cloudfoundry.identity.uaa.zone.ZoneDoesNotExistsException;
-import org.cloudfoundry.identity.uaa.zone.beans.IdentityZoneManager;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import static java.sql.Types.VARCHAR;
+import static java.util.stream.Collectors.joining;
+import static org.springframework.util.StringUtils.hasText;
+
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Timestamp;
+import java.sql.Types;
+import java.util.Arrays;
+import java.util.Calendar;
+import java.util.Date;
+import java.util.GregorianCalendar;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
+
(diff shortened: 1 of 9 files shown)


Tallicia:
Running through tests and this should be ready to merge once we get word back from Daniel and David.

uaa/src/main/webapp/WEB-INF/spring/scim-endpoints.xml
```diff
@@ -31,6 +31,7 @@
         <property name="deactivateOnDelete" value="${scim.delete.deactivate:false}"/>
         <property name="usernamePattern" value="${scim.username_pattern:[\p{L}+0-9+\-_.@'!]+}"/>
         <property name="timeService" ref="timeService"/>
+        <property name="useCaseInsensitiveQueries" ref="useCaseInsensitiveQueries"/>
```
Tallicia, line 34: Like seeing this.

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
```diff
@@ -172,6 +184,77 @@ public List<ScimUser> retrieveByUsernameAndZone(String username, String zoneId)
...
+            @Override
+            public String[] mapFromInternal(final String[] attr) {
+                return Stream.of(attr).map(this::mapFromInternal).toArray(String[]::new);
+            }
+        };
+        queryConverter.setAttributeNameMapper(attributeNameMapper);
+
+        // build WHERE clause
+        final ProcessedFilter where = queryConverter.convert(filter, sortBy, ascending, zoneId);
+        final String whereClauseScimFilter = where.getSql();
```
Tallicia, line 233: I believe this is where the many parens and ORs are generated, this should be looked at to refactor to use an IN clause.

server/src/test/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioningTests.java
```diff
@@ -1386,7 +1491,11 @@ private static void addMembership(final JdbcTemplate jdbcTemplate,
     }
 
     private static void addIdentityProvider(JdbcTemplate jdbcTemplate, String idzId, String originKey) {
-        jdbcTemplate.update("insert into identity_provider (id,identity_zone_id,name,origin_key,type) values (?,?,?,?,'UNKNOWN')", UUID.randomUUID().toString(), idzId, originKey, originKey);
+        addIdentityProvider(jdbcTemplate, idzId, originKey, true);
+    }
+
+    private static void addIdentityProvider(JdbcTemplate jdbcTemplate, String idzId, String originKey, boolean active) {
+        jdbcTemplate.update("insert into identity_provider (id,identity_zone_id,name,origin_key,type,active) values (?,?,?,?,'UNKNOWN',?)", UUID.randomUUID().toString(), idzId, originKey, originKey, active);
```
Tallicia, line 1498: good addition of active

Tallicia (approved):
Looks good, would like to see at 1-2 follow up PRs for SQL generated test, and where clause refactor to use `IN` instead of many nested ORs

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
swalchemist: I was surprised to see that the deep nesting has been there a long time, and wasn't the cause of the performance degradation. But still it would be much easier to read the SQL without all the nesting.

PR #2863, opened 2024-05-02
Refactoring: testable code in JdbcScimUserProvisioning

strehle:
* Create bean for extra SimpleSearchQueryConverter
* Need an extra SimpleSearchQueryConverter for the search with join but this object can be created once.

* Need NamedParameterJdbcTemplate for the search but this object can be created once

* Removed interface methods which are never used

Created JUnit tests for all cases in retrieveByScimFilterOnlyActive and verify the existence of LOWER

e.g.
https://www.baeldung.com/java-clean-code#characteristics_of_clean_code

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/resources/JoinAttributeNameMapper.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/resources/JoinAttributeNameMapper.java
@@ -1,0 +1,28 @@
+package org.cloudfoundry.identity.uaa.resources;
+
+public class JoinAttributeNameMapper implements AttributeNameMapper {
+
+  private final String customPrefix;
+  private final String joinPrefix;
+  private final int prefixLength;
+
+  public JoinAttributeNameMapper(String prefix) {
+    customPrefix = prefix;
+    joinPrefix = prefix + ".";
+    prefixLength = joinPrefix.length();
+  }
+
+  @Override
+  public String mapToInternal(String attr) {
+    return joinPrefix + attr;
+  }
+
+  @Override
+  public String mapFromInternal(String attr) {
+    return attr.substring(prefixLength);
+  }
+
+  public String getCustomPrefix() {
+    return customPrefix;
+  }
+}
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/resources/JoinAttributeNameMapper.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/resources/JoinAttributeNameMapper.java
@@ -1,14 +1,19 @@
 package org.cloudfoundry.identity.uaa.resources;
 
+/**
+ * Support table joins using a prefixed attribute mapping, e.g.
+ * select * from table1 joinName join table2 joinName2 on joinName.origin = joinName2.origin_key ...
+ * Used in SearchQueryConverter
+ */
 public class JoinAttributeNameMapper implements AttributeNameMapper {
 
-  private final String customPrefix;
+  private final String name;
   private final String joinPrefix;
   private final int prefixLength;
 
-  public JoinAttributeNameMapper(String prefix) {
-    customPrefix = prefix;
-    joinPrefix = prefix + ".";
+  public JoinAttributeNameMapper(String name) {
+    this.name = name;
+    joinPrefix = name + ".";
     prefixLength = joinPrefix.length();
   }
 
@@ -22,7 +27,7 @@ public class JoinAttributeNameMapper implements AttributeNameMapper {
     return attr.substring(prefixLength);
   }
 
-  public String getCustomPrefix() {
-    return customPrefix;
+  public String getName() {
+    return name;
   }
 }
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/resources/jdbc/AbstractQueryable.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/resources/jdbc/AbstractQueryable.java
@@ -14,10 +14,11 @@ import java.util.Set;
 import java.util.stream.Collectors;
 
 import static com.google.common.primitives.Ints.tryParse;
+import static org.cloudfoundry.identity.uaa.resources.jdbc.SearchQueryConverter.ProcessedFilter.ORDER_BY;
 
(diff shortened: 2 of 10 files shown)


server/src/main/java/org/cloudfoundry/identity/uaa/resources/jdbc/AbstractQueryable.java
```diff
@@ -43,6 +44,10 @@ public void setQueryConverter(SearchQueryConverter queryConverter) {
         this.queryConverter = queryConverter;
     }
 
+    public void setNamedParameterJdbcTemplate(NamedParameterJdbcTemplate namedParameterJdbcTemplate) {
```
strehle, line 47: this can be removed / refactored, if https://github.com/cloudfoundry/uaa/pull/2864 is merged before

adrianhoelzl-sap (approved):

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
```diff
@@ -191,45 +190,13 @@ public List<ScimUser> retrieveByScimFilterOnlyActive(
...
-                return attr.substring(2);
-            }
-
-            @Override
-            public String[] mapFromInternal(final String[] attr) {
-                return Stream.of(attr).map(this::mapFromInternal).toArray(String[]::new);
-            }
-        };
-        queryConverter.setAttributeNameMapper(attributeNameMapper);
+        String joinPrefix = ((SimpleSearchQueryConverter) joinConverter).getCustomPrefix();
```
swalchemist: The need to use a cast here is a pretty big code smell. Can we find a better way to make this work?

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
```diff
@@ -239,19 +206,21 @@ public String[] mapFromInternal(final String[] attr) {
         }
 
         final String userFieldsWithPrefix = Arrays.stream(USER_FIELDS.split(","))
-                .map(field -> "u." + field)
+                .map(field -> joinPrefix + "." + field)
                 .collect(joining(", "));
         final String sql = String.format(
-                "select %s from users u join identity_provider idp on u.origin = idp.origin_key and u.identity_zone_id = idp.identity_zone_id where %s",
+                "select %s from users %s join identity_provider idp on %s.origin = idp.origin_key and %s.identity_zone_id = idp.identity_zone_id where %s",
```
swalchemist: Not sure why this changed to parameterize the alias "u" for the users table, and also use a calculated value for it. Why the added complexity here?

server/src/main/java/org/cloudfoundry/identity/uaa/resources/jdbc/SimpleSearchQueryConverter.java
```diff
@@ -347,4 +349,8 @@ private Object getStringOrDate(String s) {
     public String map(String attribute) {
         return hasText(attribute) ? mapper.mapToInternal(attribute) : attribute;
     }
+
+    public String getCustomPrefix() {
+        return (mapper instanceof JoinAttributeNameMapper joinAttributeNameMapper) ? joinAttributeNameMapper.getCustomPrefix() : UaaStringUtils.EMPTY_STRING;
```
swalchemist: I'm not happy about the design requiring the use of "instanceof", though it's not as bad as the cast mentioned in the comment below.

server/src/main/java/org/cloudfoundry/identity/uaa/resources/JoinAttributeNameMapper.java
```diff
@@ -0,0 +1,28 @@
+package org.cloudfoundry.identity.uaa.resources;
+
+public class JoinAttributeNameMapper implements AttributeNameMapper {
```
swalchemist, line 8: From the name of this, I can't tell what the purpose of this class is. Maybe explain in a comment?

swalchemist (commented):
Some observations from my first reading of this PR.

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
strehle: changed

server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
strehle: changed

server/src/main/java/org/cloudfoundry/identity/uaa/resources/JoinAttributeNameMapper.java
strehle: Added comment

server/src/main/java/org/cloudfoundry/identity/uaa/resources/jdbc/SimpleSearchQueryConverter.java
strehle: Ok, I am surprised, because "instance of" is a typical UAA pattern because we have so many interfaces and mixing then the objects, but there is a good news with java 17. instance of has received an improvement so that you dont have to check and cast an object but now you can do it with one statement - like I did.

Sonar findings about instance of, e.g. (we have here 191 smells): 
https://sonarcloud.io/project/issues?resolved=false&rules=java%3AS6201&id=cloudfoundry-identity-parent

Additional ref. https://howtodoinjava.com/java/basics/pattern-matching-instanceof/

swalchemist: Ok, I hadn't noticed the use of pattern matching. I still think it's a design smell to need to use it, but won't insist on changing it.

(Its use in other places in UAA doesn't make me like it any better.)

swalchemist (approved):

PR #2864, opened 2024-05-02
Refactoring: use namedJdbcTemplate bean instead of internal new object

strehle:
Allow to mock the objects in future

e.g. if jdbcTemplate is a mock then sometimes such pattern new NamedParameterJdbcTemplate(jdbcTemplate) causes a not testable code block like in JdbcScimUserProvisioning

For jdbcTemplate we create one instance and re-use it, therefore if NamedParameterJdbcTemplate is used in several beans, then we should do the same

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/client/JdbcQueryableClientDetailsService.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/client/JdbcQueryableClientDetailsService.java
@@ -10,8 +10,8 @@ import org.cloudfoundry.identity.uaa.zone.MultitenantJdbcClientDetailsService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Qualifier;
-import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.jdbc.core.RowMapper;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.security.oauth2.provider.ClientDetails;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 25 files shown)


adrianhoelzl-sap (approved):

peterhaochen47:
> Allow to mock the objects in future
> e.g. if jdbcTemplate is a mock then sometimes such pattern new NamedParameterJdbcTemplate(jdbcTemplate) causes a not testable code block like in JdbcScimUserProvisioning

Hi! I don't think I understand this point. Could you elaborate? What code block is not testable under `JdbcTemplate` and how does `NamedParameterJdbcTemplate` solve that problem? 

It looks like the main advantage of `NamedParameterJdbcTemplate` is that you could perform param substitution on the SQL string in [a stricter manner](https://stackoverflow.com/questions/16359316/namedparameterjdbctemplate-vs-jdbctemplate) (but I don't see this feature being used in this PR?).

strehle:
> > Allow to mock the objects in future
> > e.g. if jdbcTemplate is a mock then sometimes such pattern new NamedParameterJdbcTemplate(jdbcTemplate) causes a not testable code block like in JdbcScimUserProvisioning
> 
> Hi! I don't think I understand this point. Could you elaborate? What code block is not testable under `JdbcTemplate` and how does `NamedParameterJdbcTemplate` solve that problem?
> 
> It looks like the main advantage of `NamedParameterJdbcTemplate` is that you could perform param substitution on the SQL string in [a stricter manner](https://stackoverflow.com/questions/16359316/namedparameterjdbctemplate-vs-jdbctemplate) (but I don't see this feature being used in this PR?).

The change is, that classes which use this can now be mocked, as done with https://github.com/cloudfoundry/uaa/pull/2863

Another reason why I wanted this refactoring is that I was wondering for a while, why we create a datasource and jdbcTemplate in a single place -> data-source.xml and re-use this reference in all needed places but the enhance jdbc class NamedParameterJdbcTemplate is always newly generated with each bean creating, see 
https://github.com/search?q=repo%3Acloudfoundry%2Fuaa+%22new+NamedParameterJdbcTemplate%22&type=code

I think we will re-use NamedParameterJdbcTemplate more often therefore good to use it in same way we do with underlying jdbcTemplate/datasource

peterhaochen47 (approved):

PR #2866, opened 2024-05-03
fix: replace regex usage with simple search

strehle:
Refactor use of regex to create a count statement
Issue found by Sonar


server/src/main/java/org/cloudfoundry/identity/uaa/resources/jdbc/JdbcPagingList.java
```diff
@@ -92,20 +94,23 @@ public E get(int index) {
...
             throw new IndexOutOfBoundsException("The indexes provided are outside the bounds of this list.");
         }
-        return new SafeIteratorList<E>(super.subList(fromIndex, toIndex));
+        return new SafeIteratorList<>(super.subList(fromIndex, toIndex));
     }
 
     private String getCountSql(String sql) {
-        String result = sql.replaceAll("(?i)select (.*?) from (.*)", "select count(*) from $2");
-        int orderByPos = result.toLowerCase().lastIndexOf("order by");
+        String result = UaaStringUtils.getValidatedString(sql).toLowerCase();
```
coolgang123: Converting it to lower case will not work as intended. Consider the sql query with a where clause with upper case criteria:

SELECT Foo.id, FOO.NAME from foo where name = 'Bar'

server/src/test/java/org/cloudfoundry/identity/uaa/resources/jdbc/JdbcPagingListTests.java
```diff
@@ -105,6 +106,23 @@ void jumpOverPages() {
         assertNotNull(map.get("name"));
     }
 
+    @Test
+    void selectColumnsFull() {
```
coolgang123, line 111: Could you add a test case with where and orderBy please

server/src/main/java/org/cloudfoundry/identity/uaa/resources/jdbc/JdbcPagingList.java
strehle: Thanks for the hint

server/src/main/java/org/cloudfoundry/identity/uaa/resources/jdbc/JdbcPagingList.java
```diff
@@ -73,7 +74,7 @@
         this.sql = sql;
         this.args = args;
         this.mapper = mapper;
-        this.size = parameterJdbcTemplate.queryForObject(getCountSql(sql), args, Integer.class);
+        this.size = Optional.ofNullable(parameterJdbcTemplate.queryForObject(getCountSql(sql), args, Integer.class)).orElse(0);
```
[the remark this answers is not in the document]
strehle: SQLs are not from users but internally

hsinn0 (approved):

PR #2868, opened 2024-05-03
fix: generate email if it is the empty string on external login

mikeroda:
If the email is the empty string, treat it as null and generate one.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
@@ -380,8 +380,8 @@ public class ExternalOAuthAuthenticationManager extends ExternalLoginAuthenticat
             Object verifiedObj = claims.get(emailVerifiedClaim == null ? "email_verified" : emailVerifiedClaim);
             boolean verified =  verifiedObj instanceof Boolean ? (Boolean)verifiedObj: false;
 
-            if (email == null) {
-                email = generateEmailIfNull(username);
+            if (!StringUtils.hasText(email)) {
+                email = generateEmailIfNullOrEmpty(username);
             }
 
             logger.debug(String.format("Returning user data for username:%s, email:%s", username, email));
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/manager/ExternalLoginAuthenticationManager.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/manager/ExternalLoginAuthenticationManager.java
@@ -245,8 +245,8 @@ public class ExternalLoginAuthenticationManager<ExternalAuthenticationDetails> i
             }
         }
 
-        if (email == null) {
-            email = generateEmailIfNull(name);
+        if (StringUtils.isEmpty(email)) {
+            email = generateEmailIfNullOrEmpty(name);
         }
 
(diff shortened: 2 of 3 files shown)


mikeroda:
Without the change, the test case will throw an exception:

java.lang.IllegalArgumentException: Email is required
    at org.springframework.util.Assert.hasText(Assert.java:289)
    at org.cloudfoundry.identity.uaa.user.UaaUser.<init>(UaaUser.java:152)

Because UaaUser has Assert.hasText(prototype.getEmail(), "Email is required") which is inconsistent with the null check in ExternalLoginAuthenticationManager.

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
```diff
@@ -380,8 +380,8 @@ protected UaaUser getUser(Authentication request, AuthenticationData authenticat
             Object verifiedObj = claims.get(emailVerifiedClaim == null ? "email_verified" : emailVerifiedClaim);
             boolean verified =  verifiedObj instanceof Boolean ? (Boolean)verifiedObj: false;
 
-            if (email == null) {
-                email = generateEmailIfNull(username);
+            if (!StringUtils.hasText(email)) {
```
peterhaochen47, line 383: Why use `!StringUtils.hasText(email)` here but `StringUtils.isEmpty(email)` in `ExternalLoginAuthenticationManager.java`? Why the different logics?

It looks like `!StringUtils.hasText` is a narrower/stricter validation? It also evaluates to `true` if the string contains only whitespace(s).

peterhaochen47:
Also, what is the issue in real life? When logging in with an external IDP, the email value we get from the IDP would be an empty string & the user will fail to log in due to "Email is required" error? Why would the IDP has an empty string as the email (instead of just not having it)? Is the real issue here the operator didn't set up the `attributeMappings` config correctly [for UAA API](https://docs.cloudfoundry.org/api/uaa/version/77.8.0/#oauth-oidc)?

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
mikeroda: Only because this one uses the Spring StringUtils and isEmpty is deprecated. ExternalLoginAuthenticationManager uses the Apache StringUtils. I didn't want to affect other lines of code by changing one or the other, and I didn't want to use a deprecated method.

mikeroda:
> Also, what is the issue in real life? When logging in with an external IDP, the email value we get from the IDP would be an empty string & the user will fail to log in due to "Email is required" error? Why would the IDP has an empty string as the email (instead of just not having it)? Is the real issue here the operator didn't set up the `attributeMappings` config correctly [for UAA API](https://docs.cloudfoundry.org/api/uaa/version/77.8.0/#oauth-oidc)?

The issue is that it's treated inconsistently. A null email will result in one being generated. An empty string will throw an exception that email is required. In reality an attributeMapping for email is not required, that why we can generate one. So we should be consistent about how we do that.

peterhaochen47 (approved):

peterhaochen47:
Hi @mikeroda, approved, please merge.

mikeroda:
> Hi @mikeroda, approved, please merge.

I don't have merge rights.

PR #2870, opened 2024-05-07
Fix: Zone creation fails when allow list does not contain all default groups

### LINKED ISSUE #2889: Missing Response Status Checks in IntegrationTestUtils

77.8.0 - older versions also affected

executing integration tests locally

### Summary

Some methods in `IntegrationTestUtils` do not check the response code of an HTTP call against the expected one. Instead, these methods just return the body of the response.

Furthermore, the used HTTP client (at least in the methods we checked) does not fail when calling `response.getBody()` on the response object of a failed request. Instead, it returns an object of the payload type where all properties are set to `null`.

This means that in the test cases that call these `IntegrationTestUtils` methods, failing requests are often either left unnoticed or cause other assertions to fail, making it harder to understand the cause for failing test cases.

#### Example

https://github.com/cloudfoundry/uaa/blob/ef939f7a965040794f6a396130391e3a0ae9d5f9/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/ScimGroupEndpointsIntegrationTests.java#L261

In this test case, we update an identity zone's configuration to include a certain group in its allowlist. Later in the test, we expect that it should now be possible to create the group in the zone.

However, we did not notice that the creation of the zone failed due to a bug in the zone endpoints because the error response was not checked in the corresponding `IntegrationTestUtils` method (here: `createClientAdminTokenInZone`).

Some lines below that, we then successfully create the group, which leads to the overall test case being successful.

BUT: If the zone creation would have been successful, the group would already be created then and subsequent creations of the group should now fail (`409 CONFLICT`). When we fixed the zone creation bug, the test now fails. However, it did not fail - as one would expect - during the `createGroup` call, but rather at the assertion in line 274: 

https://github.com/cloudfoundry/uaa/blob/ef939f7a965040794f6a396130391e3a0ae9d5f9/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/ScimGroupEndpointsIntegrationTests.java#L274

This is because for failing requests, the HTTP client returns an instance of the response class with all properties being set to `null`, which lead to the assertion failing.

### Solution

To make root cause analyses of failing integration tests easier, we suggest that all methods in `IntegrationTestUtils` should perform checks against the response status code. For some, this is already the case, for most of them however not.

Example: added assertions for the status code in this commit 884416db2b0ad88b56d8847893fd99b3cdff1c0b

adrianhoelzl-sap:
see issue #2505 

With PR #2606, we introduced an allow list for the groups in an identity zone. This PR fixes the issue that the creation of a zone fails whenever the allow list does not contain all system scopes ("scim.read/write", "sps.read/write", etc.).

Now, we only create those groups during zone creation that are part of the effectively allowed groups, i.e., all default groups and all groups in the allowlist.


uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/ScimGroupEndpointsIntegrationTests.java
adrianhoelzl-sap, line 1: We had to adjust the integration tests due to missing assertions in some `IntegrationTestUtils` methods, see issue #2889.

Before this PR, the tests were successful even though there were failures in the underlying requests. With the zone creation now working, the tests failed since the subsequent requests in the test cases now fail - as expected.

model/src/main/java/org/cloudfoundry/identity/uaa/zone/UserConfig.java
```diff
@@ -50,6 +52,7 @@ public boolean allGroupsAllowed() {
     }
 
     // return defaultGroups plus allowedGroups
+    @Nullable
```
strehle, line 55: what is the benefit of this ?

I see some with NonNullable, but Nullable..? so i would omit it

strehle (commented):
please rebase this , since some time has gone

model/src/main/java/org/cloudfoundry/identity/uaa/zone/UserConfig.java
adrianhoelzl-sap: When using this annotation, the IDE is able to show warnings when using the return value without a preceding null check. I tend to use this annotation in cases like this where the return type is a collection and the method returns null instead of an empty collection.

strehle (approved):
LGTM

PR #2879, opened 2024-05-09
refactor: avoid indirect dep from EOL lib spring-security-saml2-core

peterhaochen47:
- AuthorizationServerBeanDefinitionParser depends on spring-security-config, which before this commmit comes indirectly from spring-security-saml2-core (which is an EOL lib we will soon replace); hence instead, in this commit, import spring-security-config directly

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -21,6 +21,7 @@ dependencies {
     implementation(libraries.springWeb)
     implementation(libraries.springSecurityCore)
     implementation(libraries.springSecurityWeb)
+    implementation(libraries.springSecurityConfig)
     implementation(libraries.springBootStarterMail)
     implementation(libraries.owaspEsapi) {
         transitive = false


strehle:
before:
image


now:

image

strehle:
WAR still has same libs, but now cloudfoundry-identity-server uses it directly. Before it with indirect from cloudfoundry-identity-model

strehle (approved):

peterhaochen47:
For your screenshot, I think it's only showing the `uaa` sub-project. `AuthorizationServerBeanDefinitionParser` is in the `server` sub-project.

strehle:
> For your screenshot, I think it's only showing the `uaa` sub-project. `AuthorizationServerBeanDefinitionParser` is in the `server` sub-project.

yes, but if I would use server and its dependencies, then it would look like

before
image


after
image

peterhaochen47:
> WAR still has same libs, but now cloudfoundry-identity-server uses it directly. Before it with indirect from cloudfoundry-identity-model

I don't think I quite understand this haha. Inside of the `server` sub-project gradle file, if you comment out the EOL spring-security-saml2-core lib dep, `AuthorizationServerBeanDefinitionParser` will fail to compile, indicating that spring-security-saml2-core is the sole provider of spring-security-config - yeah not sure why the `model` can't supply it.

strehle:
> I don't think I quite understand this haha. Inside of the `server` sub-project gradle file, if you comment out the EOL spring-security-saml2-core lib dep, `AuthorizationServerBeanDefinitionParser` will fail to compile, indicating that spring-security-saml2-core is the sole provider of spring-security-config - yeah not sure why the `model` can't supply it.

but now it works ?

peterhaochen47:
> > I don't think I quite understand this haha. Inside of the `server` sub-project gradle file, if you comment out the EOL spring-security-saml2-core lib dep, `AuthorizationServerBeanDefinitionParser` will fail to compile, indicating that spring-security-saml2-core is the sole provider of spring-security-config - yeah not sure why the `model` can't supply it.
> 
> but now it works ?

Yes, right now, if you comment out  `spring-security-saml2-core`, it should still compile. This is the main motivation for this PR - on our SAML branch, we removed the ` spring-security-saml2-core` lib, so our rebase to develop failed but should work now.

PR #2882, opened 2024-05-10
 feature: filter IdP retrival

strehle:
- New: Filter IdP list by origin

Many calls to IdPs first retrieve all and then filter by origin on client side. The origin has an index in DB, therefore there should be an option to get the IdP by origin filter.

NEW 

Filter: uaac curl -b /identity-providers?origin=sap.proxy | jq -r .[0]

Allow to get IdP based on origin. Do not iterate over all.

Sonar
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2882


strehle:
generated documentation can be downloaded as ZIP
https://github.com/cloudfoundry/uaa/actions/runs/9034861916?pr=2882

strehle:
@torsten-sap 

NEW 

Filter: uaac curl -b /identity-providers?origin=sap.proxy | jq -r .[0]

Allow to get IdP based on origin. Do not iterate over all.

torsten-sap:
@strehle In comparison to https://docs.cloudfoundry.org/api/uaa/version/77.1.0/index.html#change-secret, does it make sense to also allow the update here (besides delete)? I would assume this makes things easier for some customers if only the secret shall be changed (although this is of course already possible via update of the IdP).

strehle:
> @strehle In comparison to https://docs.cloudfoundry.org/api/uaa/version/77.1.0/index.html#change-secret, does it make sense to also allow the update here (besides delete)? I would assume this makes things easier for some customers if only the secret shall be changed (although this is of course already possible via update of the IdP).

Ok, we can support the easy change of a secret, however I would stay with separat HTTP methods, e.g. DELETE and/or PATCH instead of a mix in PUT

See https://github.com/cloudfoundry/uaa/pull/2882/commits/cd32e6149374461b837a21daf02af151ab1361f5 which allows to patch a secret only

strehle:
> @strehle In comparison to https://docs.cloudfoundry.org/api/uaa/version/77.1.0/index.html#change-secret, does it make sense to also allow the update here (besides delete)? I would assume this makes things easier for some customers if only the secret shall be changed (although this is of course already possible via update of the IdP).

Created https://github.com/cloudfoundry/uaa/pull/2885 because from Prio the deletion should be added soon, all other changes in REST API of IdPs are an improvement but not a bug fixing

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/providers/IdentityProviderEndpointDocs.java
```diff
@@ -967,6 +967,47 @@ void getAllIdentityProviders() throws Exception {
...
+                    headerWithName("Authorization").description("Bearer token containing `zones.<zone id>.admin` or `zones.<zone id>.idps.read` or `uaa.admin` or `idps.read` (only in the same zone that you are a user of)"),
+                    headerWithName("X-Identity-Zone-Id").description("May include this header to administer another zone if using `zones.<zoneId>.admin` or `zones.<zone id>.idps.read` or `uaa.admin` scope against the default UAA zone.").optional(),
+                    IDENTITY_ZONE_SUBDOMAIN_HEADER
+                ),
+                requestParameters(
+                    parameterWithName("rawConfig").optional("false").type(BOOLEAN).description("Flag indicating whether the response should use raw, unescaped JSON for the `config` field of the IDP, rather than the default behavior of encoding the JSON as a string."),
+                    parameterWithName("active_only").optional("false").type(BOOLEAN).description("Flag indicating whether only active IdPs should be returned or all."),
+                    parameterWithName("origin").optional(null).type(STRING).description("<small><mark>UAA 77.10.0</mark></small> Return only IdPs with specific origin.")
```
peterhaochen47: Is this `origin` the same thing as the `originKey` when you [create the IDPs](https://docs.cloudfoundry.org/api/uaa/version/77.8.0/#saml)? If so, why not just call this field `originKey` to keep it consistent?

strehle: ok done

peterhaochen47 (approved):

Issue #2741, opened 2024-02-21
Avoid necessity to configure SAML SP in UAA


torsten-sap:
### What version of UAA are you running?

76.30

### How are you deploying the UAA?

- using cf-deployment

### What did you do?

Usage of UAA without the need of SAML.

### What did you expect to see? What goal are you trying to achieve with the UAA?

No need to configure SAML SP (including private key + certificate etc.) in uaa.yml.

### What did you see instead?

SAML SP configuration (private key + certificate etc.) is required in uaa.yml. Otherwise, UAA will not startup.

strehle:
Reproduce the issue:

1. remove https://github.com/cloudfoundry/uaa/blob/develop/scripts/cargo/uaa.yml#L58-L99
2. start uaa 

Open /login

The IdentityZone should be usable even without SAML keys, but there is execption:

.....a.lang.NullPointerException: Cannot invoke "org.springframework.security.saml.key.KeyManager.getDefaultCredentialName()" because the return value of "org.cloudfoundry.identity.uaa.zone.IdentityZoneHolder.getSamlSPKeyManager()" is null
        at org.cloudfoundry.identity.uaa.provider.saml.ZoneAwareKeyManager.getDefaultCredentialName(ZoneAwareKeyManager.java:41) ~[cloudfoundry-identity-server-0.0.0.jar:?]
        at org.springframework.security.saml.metadata.MetadataGenerator.getSigningKey(MetadataGenerator.java:802) ~[spring-security-saml2-core-1.0.10.RELEASE.jar:1.0.10.RELEASE]
        at org.springframework.security.saml.metadata.MetadataGenerator.buildSPSSODescriptor(MetadataGenerator.java:323) ~[spring-security-saml2-core-1.0.10.RELEASE.jar:1.0.10.RELEASE]
        at org.cloudfoundry.identity.uaa.provider.saml.ZoneAwareMetadataGenerator.buildSPSSODescriptor(ZoneAwareMetadataGenerator.java:101) ~[cloudfoundry-identity-server-0.0.0.jar:?]
        at org.springframework.security.saml.metadata.MetadataGenerator.generateMetadata(MetadataGenerator.java:189) ~[spring-security-saml2-core-1.0.10.RELEASE.jar:1.0.10.RELEASE]

hsinn0:
We will plan to look into it to prioritize in next iteration planning session.

aramprice:
Is there work planned to address this issue?

[...]

PR #2884, opened 2024-05-10
fix: /info docs test expectation

peterhaochen47:
- the value of "idpDefinitions.*" of /info response is in reality a String, not an array. Here is how it looks like:
```
 "idpDefinitions": {
    "testsaml-redirect-binding": "http://localhost:8080/uaa/saml/discovery?returnIDParam=idp&entityID=cloudfoundry-saml-login&idp=testsaml-redirect-binding&isPassive=true",
    "testsaml-post-binding": "http://localhost:8080/uaa/saml/discovery?returnIDParam=idp&entityID=cloudfoundry-saml-login&idp=testsaml-post-binding&isPassive=true"
  },
```
which is consistent with the description of "idpDefinitions" field: "A list of alias/url pairs of SAML IDP providers configured. Each url is the starting point to initiate the authentication process for the SAML identity provider."

- previously, this test is passing only because the test uaa.yml used in this test contains no IDP configs; but once you input some valid IDP configs, this test would fail since the value is actually a String.

- also clarify the description text

### DIFF
--- a/uaa/src/test/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpointDocs.java
+++ b/uaa/src/test/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpointDocs.java
@@ -38,6 +38,7 @@ import static org.springframework.restdocs.payload.JsonFieldType.ARRAY;
 import static org.springframework.restdocs.payload.JsonFieldType.BOOLEAN;
 import static org.springframework.restdocs.payload.JsonFieldType.OBJECT;
 import static org.springframework.restdocs.payload.JsonFieldType.STRING;
+import static org.springframework.restdocs.payload.JsonFieldType.VARIES;
 import static org.springframework.restdocs.payload.PayloadDocumentation.requestFields;
 import static org.springframework.restdocs.payload.PayloadDocumentation.responseFields;
 import static org.springframework.restdocs.request.RequestDocumentation.requestParameters;
@@ -59,7 +60,7 @@ class LoginInfoEndpointDocs extends EndpointDocs {
                 fieldWithPath("commit_id").type(STRING).description("The GIT sha for the UAA version"),
                 fieldWithPath("timestamp").type(STRING).description("JSON timestamp for the commit of the UAA version"),
                 fieldWithPath("idpDefinitions").optional().type(OBJECT).description("A list of alias/url pairs of SAML IDP providers configured. Each url is the starting point to initiate the authentication process for the SAML identity provider."),
-                fieldWithPath("idpDefinitions.*").optional().type(STRING).description("The URL to initiate the authentication process for the SAML identity provider."),
+                fieldWithPath("idpDefinitions.*").optional().type(VARIES).description("The URL to initiate the authentication process for the SAML identity provider."),
                 fieldWithPath("links").type(OBJECT).description("A list of alias/url pairs of configured action URLs for the UAA"),
                 fieldWithPath("links.login").type(STRING).description("The link to the login host alias of the UAA"),
                 fieldWithPath("links.uaa").type(STRING).description("The link to the uaa alias host of the UAA"),
--- a/uaa/src/test/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpointDocs.java
+++ b/uaa/src/test/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpointDocs.java
@@ -59,7 +59,7 @@ class LoginInfoEndpointDocs extends EndpointDocs {
                 fieldWithPath("commit_id").type(STRING).description("The GIT sha for the UAA version"),
                 fieldWithPath("timestamp").type(STRING).description("JSON timestamp for the commit of the UAA version"),
                 fieldWithPath("idpDefinitions").optional().type(OBJECT).description("A list of alias/url pairs of SAML IDP providers configured. Each url is the starting point to initiate the authentication process for the SAML identity provider."),
-                fieldWithPath("idpDefinitions.*").optional().type(ARRAY).description("A list of alias/url pairs of SAML IDP providers configured. Each url is the starting point to initiate the authentication process for the SAML identity provider."),
+                fieldWithPath("idpDefinitions.*").optional().type(STRING).description("The URL to initiate the authentication process for the SAML identity provider."),
                 fieldWithPath("links").type(OBJECT).description("A list of alias/url pairs of configured action URLs for the UAA"),
                 fieldWithPath("links.login").type(STRING).description("The link to the login host alias of the UAA"),
                 fieldWithPath("links.uaa").type(STRING).description("The link to the uaa alias host of the UAA"),


uaa/src/test/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpointDocs.java
```diff
@@ -59,7 +59,7 @@ void info_endpoint_for_json() throws Exception {
                 fieldWithPath("commit_id").type(STRING).description("The GIT sha for the UAA version"),
                 fieldWithPath("timestamp").type(STRING).description("JSON timestamp for the commit of the UAA version"),
                 fieldWithPath("idpDefinitions").optional().type(OBJECT).description("A list of alias/url pairs of SAML IDP providers configured. Each url is the starting point to initiate the authentication process for the SAML identity provider."),
-                fieldWithPath("idpDefinitions.*").optional().type(ARRAY).description("A list of alias/url pairs of SAML IDP providers configured. Each url is the starting point to initiate the authentication process for the SAML identity provider."),
+                fieldWithPath("idpDefinitions.*").optional().type(STRING).description("The URL to initiate the authentication process for the SAML identity provider."),
```
strehle: ok, it is not a JSON array but it is also not a pure String, but an JSON object simply...

So depends from which side you want solve it. Pure syntax, then object or human readable, then it is more an array

peterhaochen47: Hi, why is `idpDefinitions.*` a JSON object? 
Based on the sample /info endpoint output in the PR description, it currently looks like `idpDefinitions`'s value is an object (hence line 61 in the quoted code block is asserting so). But it looks like the value of `idpDefinitions.*` is an URL string. Also see [test here](https://github.com/cloudfoundry/uaa/blob/bef96ed0af5b7cd2cbe27dc495aa5b7cf1a68dec/server/src/test/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpointTests.java#L535-L540) showing that `idpDefinitions` is a `Map<String, String>` where the values are strings.

strehle: because these Maps are JSON objects. see https://www.w3schools.com/js/js_json_objects.asp

My point is, array maybe was not the best choice, but String is even worse, so either stay with this type or use the correct one, which is Object

strehle:
> * the value of "idpDefinitions.*" of /info response is in reality a String, not an array. Here is how it looks like:
> 
> ```
>  "idpDefinitions": {
>     "testsaml-redirect-binding": "http://localhost:8080/uaa/saml/discovery?returnIDParam=idp&entityID=cloudfoundry-saml-login&idp=testsaml-redirect-binding&isPassive=true",
>     "testsaml-post-binding": "http://localhost:8080/uaa/saml/discovery?returnIDParam=idp&entityID=cloudfoundry-saml-login&idp=testsaml-post-binding&isPassive=true"
>   },
> ```
> 
> which is consistent with the description of "idpDefinitions" field: "A list of alias/url pairs of SAML IDP providers configured. Each url is the starting point to initiate the authentication process for the SAML identity provider."
> 
> * previously, this test is passing only because the test uaa.yml used in this test contains no IDP configs; but once you input some valid IDP configs, this test would fail since the value is actually a String.
> * also clarify the description text

As always, if you have a deeper look into old UAA behavior. I find SAML links in "uaac info" [(JSON output) ](https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java#L472-L493) , but on OAUTH. 

After thinking, I would like to know if you have use /info in your products ? I think SAP never has used it really, but for us it was always a problem, therefore we introduced account chooser etc... (@torsten-sap , @tack-sap correct ?)

The point is, in /info I also see entityID, which is only for SAML. Why no OAUITH / OIDC issuer ?
In general I would omit such info, because we want in future a UAA without SAML SP , e.g. https://github.com/cloudfoundry/uaa/issues/2741

uaa/src/test/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpointDocs.java
peterhaochen47: > because these Maps are JSON objects.

In the `fieldWithPath` syntax here, `idpDefinitions` is the map (key-value pairs), but `idpDefinitions.*` is the key in the key-value pair. 

The line changed in this commit is a test (that will get turned into doc), and the test tries to validate if the type you put here matches the type in the real mockmvc api response. If we put:
```
fieldWithPath("idpDefinitions.*").optional().type(OBJECT).description("The URL to initiate the authentication process for the SAML identity provider.")
```
or 
```
fieldWithPath("idpDefinitions.*").optional().type(ARRAY).description("The URL to initiate the authentication process for the SAML identity provider.")
```
This test will fail, once you configure one or more SAML IDP in [its UAA.yml](https://github.com/cloudfoundry/uaa/blob/11b086c4017ba6b107f92ff42d17f66faf16080d/uaa/src/test/resources/integration_test_properties.yml#L95).

strehle: ok, understood, you had a test failure if you fill the yml...
Then I propose VARIES. It generates this map, which to me is no STRING

```
"idpDefinitions" : {
    "dummy" : "http://localhost:8080/uaa/saml/discovery?returnIDParam=idp&entityID=cloudfoundry-saml-login&idp=dummy&isPassive=true",
    "xs2security" : "http://localhost:8080/uaa/saml/discovery?returnIDParam=idp&entityID=cloudfoundry-saml-login&idp=xs2security&isPassive=true"
  },
```

peterhaochen47: I think we still have an misunderstanding of the `fieldWithPath` syntax. `idpDefinitions`'s value is 100% a map, NOT a string. That's why the test `fieldWithPath("idpDefinitions").optional().type(OBJECT).description("xxx")` is passing. 

But `fieldWithPath("idpDefinitions.*")` refers NOT to `idpDefinitions`'s value, but `idpDefinitions.dummy`'s value, where the `*` is matching `dummy`. So `idpDefinitions.*`'s value is really `"http://localhost:8080/uaa/saml/discovery?returnIDParam=idp&entityID=cloudfoundry-saml-login&idp=dummy&isPassive=true"`. That's why `fieldWithPath("idpDefinitions.*").optional().type(STRING).description("xxxx.")` is passing with some IDPs configured.

strehle: I know and I have run the tests with content you suggested and I could verify the test failures, however with VARIES the tests were succeed with OK

Another solution could be to have 2 extra entries, means instead idpDefinitions.*
* idpDefinitions.alias
* idpDefinitions.url 

both with type string .

strehle (approved):

PR #2885, opened 2024-05-13
fix: allow to delete a relyingPartySecret on IdP

### LINKED ISSUE #2880: Cannot remove a secret from IdP via REST

77.8.0 / develop

I am deploying the UAA

- locally only using gradlew

### What did you do?

### What did you see instead?

strehle:
Extract the delete change into an own PR, because this missing method is really a bug / issue in UAA. A UAA operator cannot remove a secret currently, but there are situation where this is needed, e.g. public flows or change to private_key_jwt.

Example to delete a secret, e.g. from DOC:

curl 'http://localhost/identity-providers/ba820e42-7bbf-445a-a241-bb49d61b8512/secret' -i -X DELETE \
    -H 'Authorization: Bearer 8b8a49f6186d4995af427ea59b40e8f8'

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/IdentityProviderEndpoints.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/IdentityProviderEndpoints.java
@@ -251,6 +251,11 @@ public class IdentityProviderEndpoints implements ApplicationEventPublisherAware
             body.setConfig(definition);
         }
 
+        return persistIdentityProviderChange(body, rawConfig, zoneId, existing);
+    }
+
+    private ResponseEntity<IdentityProvider> persistIdentityProviderChange(IdentityProvider body, boolean rawConfig, String zoneId,
+        IdentityProvider existing) {
         final IdentityProvider<?> updatedIdp;
         try {
             updatedIdp = transactionTemplate.execute(txStatus -> {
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/cloudfoundry/identity/uaa/provider/IdentityProviderEndpoints.java
```diff
@@ -308,6 +309,23 @@ public ResponseEntity<IdentityProviderStatus> updateIdentityProviderStatus(@Path
...
     }
 
+    @DeleteMapping(value = "{id}/secret")
+    public ResponseEntity<IdentityProvider> deleteSecret(@PathVariable String id) {
+        String zoneId = identityZoneManager.getCurrentIdentityZoneId();
+        IdentityProvider existing = identityProviderProvisioning.retrieve(id, zoneId);
+        if((OIDC10.equals(existing.getType()) || OAUTH20.equals(existing.getType()))
+            && existing.getConfig() instanceof AbstractExternalOAuthIdentityProviderDefinition<?> idpConfiguration) {
+            idpConfiguration.setRelyingPartySecret(null);
+            identityProviderProvisioning.update(existing, zoneId);
```
adrianhoelzl-sap: We also need to ensure that this update is performed on any potential alias of the IdP

strehle: ok

peterhaochen47:
Instead of offering a new `/identity-providers/{id}/secret` endpoint, why not just fix #2880 by fixing the existing [update IDP endpoint](https://docs.cloudfoundry.org/api/uaa/version/77.8.0/#update) to allow for a complete update of the entire IDP config (the removal of the secret, or migration to private_key_jwt, included)? That seems more straightforward? 
Are we concerned about breaking change (there should be a way to avoid it?)?

strehle:
> Instead of offering a new `/identity-providers/{id}/secret` endpoint, why not just fix #2880 by fixing the existing [update IDP endpoint](https://docs.cloudfoundry.org/api/uaa/version/77.8.0/#update) to allow for a complete update of the entire IDP config (the removal of the secret, or migration to private_key_jwt, included)? That seems more straightforward? Are we concerned about breaking change (there should be a way to avoid it?)?

ok, so I created another alternative PR, see https://github.com/cloudfoundry/uaa/pull/2896

This PR https://github.com/cloudfoundry/uaa/pull/2896 is introducing a new attribute "authMethod" and with this the change is non-breakable, but an enhancement. Without authMethod the secret is still retrieved from existing DB setting.

peterhaochen47:
Thanks! I prefer the alternative, so let's move our discussion there (#2896).

strehle:
close because we used https://github.com/cloudfoundry/uaa/pull/2896

PR #2887, opened 2024-05-14
alternative fix: allow to change or delete a relyingPartySecret on IdP

### LINKED ISSUE #2880: Cannot remove a secret from IdP via REST

strehle:
Alternative to PR: https://github.com/cloudfoundry/uaa/pull/2882
and PR: https://github.com/cloudfoundry/uaa/pull/2885

Only one call in backend, but more input on client side.


peterhaochen47:
Thanks! I prefer the alternative, so let's move our discussion there (#2896).

strehle:
close because we usee https://github.com/cloudfoundry/uaa/pull/2896

PR #2896, opened 2024-05-16
2nd alternative fix: allow to change or delete a relyingPartySecret on IdP

### LINKED ISSUE #2880: Cannot remove a secret from IdP via REST

strehle:
Alternative to PR: https://github.com/cloudfoundry/uaa/pull/2882, https://github.com/cloudfoundry/uaa/pull/2887
and PR: https://github.com/cloudfoundry/uaa/pull/2885

Only one call in backend, but more input on client side.

authMethod is used to make this change non-breakable.
Because before the update prevented the removal of the relyingPartySecret.
No if authMethod is not client_secret_basic or client_secret_post, then the relyingPartySecret can be overwritten with null.


strehle:
sonar fixes

https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2896

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/providers/IdentityProviderEndpointDocs.java
```diff
@@ -591,6 +590,7 @@ void createOAuthIdentityProvider() throws Exception {
                 fieldWithPath("config.userInfoUrl").optional(null).type(STRING).description("A URL for fetching user info attributes when queried with the obtained token authorization."),
                 fieldWithPath("config.showLinkText").optional(true).type(BOOLEAN).description("A flag controlling whether a link to this provider's login will be shown on the UAA login page"),
                 fieldWithPath("config.linkText").optional(null).type(STRING).description("Text to use for the login link to the provider"),
+                fieldWithPath("config.auth_method").optional(null).type(STRING).description("<small><mark>UAA 77.10.0</mark></small> Client authentication method. Possible strings are: client_secret_basic, client_secret_post, private_key_jwt, none."),
```
peterhaochen47: Why does this field exist in addition to `config.authMethod` (line 607, currently).

strehle: After diff of the whole code I noticed yes, this is a duplicate. We use authMethod not auth_method.

Removed auth_method

peterhaochen47 (approved):

PR #2903, opened 2024-05-29
Add kill_uaa script before run and integrationTest tasks

duanemay:
Add execution of the kill_uaa script  before the gradle tasks: run and integrationTest


Tallicia (approved):
Looks good and should be a nice time saver!

PR #2918, opened 2024-06-06
fix: Support authMethod=none in proxy password grant

strehle:
* Fix definition of authMethod in uaa.yml, which was not taken over
* Check for authMethod=none and allow to omit authentication with password grant

There are some OIDC IdPs which support password grant without secret, therefore offer this in proxy mode.

https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2918


server/src/main/java/org/cloudfoundry/identity/uaa/authentication/manager/PasswordGrantAuthenticationManager.java
```diff
@@ -132,9 +134,10 @@ Authentication oidcPasswordGrant(Authentication authentication, OIDCIdentityProv
         if (clientId == null) {
             throw new ProviderConfigurationException("External OpenID Connect provider configuration is missing relyingPartyId.");
         }
-        if (clientSecret == null && config.getJwtClientAuthentication() == null) {
-            throw new ProviderConfigurationException("External OpenID Connect provider configuration is missing relyingPartySecret or jwtclientAuthentication.");
+        if (clientSecret == null && config.getJwtClientAuthentication() == null && config.getAuthMethod() == null) {
```
hsinn0, line 137: Is this not a breaking change?

strehle: no because we have as default config.authMethod = null, means, if you do not provide any explicit method, and you dont have a secret or key configured, then if fails.

However, if you setup config.authMethod = none, then you should have the option to omit a secret (client authentication for the resource owner flow / password grant), because we have examples of OIDC providers, that allow this, e.g. https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth-ropc (Entra / Azure) ... see client_secret / client_assertion section. 
Btw. Okta and also SAP OIDC IDP and others provide this, therefore we should allow it.

server/src/test/java/org/cloudfoundry/identity/uaa/authentication/manager/PasswordGrantAuthenticationManagerTest.java
```diff
@@ -876,6 +878,31 @@ void oidcPasswordGrant_credentialsMustBeString() {
         assertThrows(BadCredentialsException.class, () -> instance.oidcPasswordGrant(authentication, config));
     }
 
+    @Test
+    void oidcPasswordGrant_credentialsMustBeStringButNoSecretNeeded() throws MalformedURLException {
+        RestTemplate restTemplate = mock(RestTemplate.class);
+        ResponseEntity responseEntity = mock(ResponseEntity.class);
+        UsernamePasswordAuthenticationToken authentication = new UsernamePasswordAuthenticationToken("user", "");
+        OIDCIdentityProviderDefinition config = new OIDCIdentityProviderDefinition()
+            .setRelyingPartyId("client-id").setTokenUrl(URI.create("http://localhost:8080/uaa/oauth/token").toURL());
```
strehle, line 887: See here the test where it is allowed

server/src/test/java/org/cloudfoundry/identity/uaa/authentication/manager/PasswordGrantAuthenticationManagerTest.java
```diff
@@ -876,6 +878,31 @@ void oidcPasswordGrant_credentialsMustBeString() {
...
+        assertNull(instance.oidcPasswordGrant(authentication, spyConfig));
+        verify(spyConfig, atLeast(2)).getAuthMethod();
+    }
+
+    @Test
+    void oidcPasswordGrant_requireAuthenticationStatement() {
+        UsernamePasswordAuthenticationToken authentication = new UsernamePasswordAuthenticationToken("user", new Object());
+        OIDCIdentityProviderDefinition config = new OIDCIdentityProviderDefinition()
+            .setRelyingPartyId("client-id");
+        assertThrows(ProviderConfigurationException.class, () -> instance.oidcPasswordGrant(authentication, config));
```
strehle: here it is not allowed.... will check for the exception test

hsinn0 (approved):

PR #2920, opened 2024-06-06
Identity provider key caching behavior configurable

strehle:
Issue: In UAA 74.5.x the JWKS is not cached. With newer UAAs there is a cache which is hard coded to 10 min.
If you have a fast rotation key, then this could be a problem.

Solution: deactivate the cache for such IdPs with cacheJwks=false options


strehle:
Alternative to https://github.com/cloudfoundry/uaa/pull/2915

model/src/main/java/org/cloudfoundry/identity/uaa/provider/AbstractExternalOAuthIdentityProviderDefinition.java
```diff
@@ -53,6 +53,7 @@ public enum OAuthGroupMappingMode {
     private boolean performRpInitiatedLogout = true;
     @JsonInclude(JsonInclude.Include.NON_NULL)
     private String authMethod;
+    private boolean cacheJwks = true;
```
strehle, line 56: @peterhaochen47 . The name of the attribute can be different, if you want or have a proposal for another name

peterhaochen47: This works

peterhaochen47:
Thanks! LGTM. I pushed a commit to edit the docs. Please review.

peterhaochen47 (approved):

PR #2923, opened 2024-06-07
 Fix issue #2917

### LINKED ISSUE #2917: Shadow Users are not getting updated after user logs out and login again

77.10.0

### Problem Statement
Today when 2 uaa are used federating with each other the shadow users that gets created is not getting updated with new set of permissions coming from access_token from the other uaa counterpart.

So what we see is shadow user , gets updated with new permissions when user authenticates but it does not remove the permissions that has been removed .

My user journey is like this 

1. User logins to the uaa (which is integerated with any Auth0 Powered IDP)
2. Application gets the access_token , id_token and refresh_token 
3. User has been removed with some previleges 
4. Now user logs out and login again 
5. He is not getting the access token reflecting the role assignments 
     Typically roles that are removed are not removed but roles that are added can be seen

strehle:
Perform shadow user creation (NewUserAuthenticatedEvent)
without authorities creation, but rely on event
ExternalGroupAuthorizationEvent later.

Includes: IT for testing a fix of issue https://github.com/cloudfoundry/uaa/issues/2917


strehle:
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2923

hsinn0 (approved):

PR #2941, opened 2024-06-21
fix: publish an external group auth event only with registered IDPs

mikeroda:
When authenticating a user with external OAuth, only publish the ExternalGroupAuthorizationEvent for registered IDPs, thereby skipping the event if authentication is through UAA itself, such as when using the JWT bearer token grant. Without this fix, the event is published every time you use the JWT bearer token grant.

### DIFF
--- a/server/src/test/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManagerIT.java
+++ b/server/src/test/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManagerIT.java
@@ -7,6 +7,7 @@ import org.apache.commons.codec.binary.Base64;
 import org.apache.commons.lang3.RandomStringUtils;
 import org.cloudfoundry.identity.uaa.authentication.AccountNotPreCreatedException;
 import org.cloudfoundry.identity.uaa.authentication.UaaAuthentication;
+import org.cloudfoundry.identity.uaa.authentication.event.IdentityProviderAuthenticationSuccessEvent;
 import org.cloudfoundry.identity.uaa.authentication.manager.ExternalGroupAuthorizationEvent;
 import org.cloudfoundry.identity.uaa.authentication.manager.InvitedUserAuthenticatedEvent;
 import org.cloudfoundry.identity.uaa.authentication.manager.NewUserAuthenticatedEvent;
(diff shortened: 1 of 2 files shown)


strehle:
clean

https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2941

so far valid

strehle (approved):

peterhaochen47:
> skipping the event if authentication is through UAA itself, such as when using the JWT bearer token grant

Hi @mikeroda, correct me if I'm wrong: After this PR, when the [JWT bearer grant](https://docs.cloudfoundry.org/api/uaa/version/77.11.0/index.html#jwt-bearer-token-grant) is invoked when the request param `assertion` is an external IDP token, the `ExternalGroupAuthorizationEvent` will still be published. This PR just skips the event when the JWT bearer grant's request `assertion` is a UAA internal token.

server/src/test/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManagerIT.java
```diff
@@ -908,6 +909,37 @@ void updateShadowUser_IfAlreadyExists() {
         assertEquals(OriginKeys.UAA, uaaUser.getZoneId());
     }
 
+    @Test
+    void publishExternalGroupAuthorizationEvent_skippedIf_isRegisteredIdpAuthentication() {
```
peterhaochen47: > publishExternalGroupAuthorizationEvent_skippedIf_isRegisteredIdpAuthentication

I thought this PR skips if it's not a "registered Idp Authentication"?

mikeroda:
> > skipping the event if authentication is through UAA itself, such as when using the JWT bearer token grant
> 
> Hi @mikeroda, correct me if I'm wrong: After this PR, when the [JWT bearer grant](https://docs.cloudfoundry.org/api/uaa/version/77.11.0/index.html#jwt-bearer-token-grant) is invoked when the request param `assertion` is an external IDP token, the `ExternalGroupAuthorizationEvent` will still be published. This PR just skips the event when the JWT bearer grant's request `assertion` is a UAA internal token.

@peterhaochen47, that's correct.

server/src/test/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManagerIT.java
mikeroda: Yes you're right. The method name is backwards. I'll fix it.

peterhaochen47 (approved):

PR #2942, opened 2024-06-25
Add 'identity_zone_id' column to indexes for 'alias_zid'

adrianhoelzl-sap:


strehle:
Refactorings
- shorten the index name. similar name pattern then existing idexes
- changed order of index. indentity_zone_id should be always first column
- use partial index for postgresql (because they support it)
- removed not null from SQL, since not needed, see https://stackoverflow.com/questions/8660203/how-to-check-for-is-not-null-and-is-not-empty-string-in-sql-server
or
https://stackoverflow.com/questions/1106258/mysql-null-vs

See indexes now
image

Verification done with some new zones with aliases, then

explain analyze select 1 from identity_provider idp where idp.identity_zone_id = 'uaa' and idp.alias_zid <> '' limit 1

Result on MYSQL (more info but similar to postgres)
-> Limit: 1 row(s)  (cost=0.886 rows=1) (actual time=0.0163..0.0164 rows=1 loops=1)
    -> Filter: ((idp.identity_zone_id = 'uaa') and (idp.alias_zid <> ''))  (cost=0.886 rows=3) (actual time=0.0156..0.0156 rows=1 loops=1)
        -> Covering index range scan on idp using alias_in_zone over (identity_zone_id = 'uaa' AND NULL < alias_zid < '') OR (identity_zone_id = 'uaa' AND '' < alias_zid)  (cost=0.886 rows=3) (actual time=0.014..0.014 rows=1 loops=1)


Both use alias_in_zone . In postgresql I found out, that the order of columns is important

strehle (approved):

PR #2948, opened 2024-07-05
feature: Zone-aware SAML SP metadata

hsinn0:
- Implemented to the same level as the default IdenityZone's SP metadata generation.
- Minus `NameIDFormat` value populaition and registration-ID specific implementation.

[#187846376]


duanemay:
LGTM

PR #2908, opened 2024-05-30
build(deps): Replace SAML Library

Tallicia:
Replacing the other feature branch https://github.com/cloudfoundry/uaa/pull/2862 for new SAML library replacement effort.

Sonar
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2908

PR #2959, opened 2024-07-16
fix: Integration test bootstrap fixed

strehle:
These tests need a bootstraping which is not in the test itself but externally. This means, these tests cannot be executed standalone, which is bad. If another test runs before, then these tests can run as well, therefore it has not been found before. Normally we run all tests in a suite.

Found the issue, when running LoginServerSecurityIntegrationTests standalone .... because this is the failing test from SAML refactoring


strehle:
created the PR. after I checked https://github.com/cloudfoundry/uaa/pull/2908

peterhaochen47 (approved):

PR #2960, opened 2024-07-16
Show error in IT

strehle:
Created this change in order to better display the problem,
See https://github.com/cloudfoundry/uaa/pull/2908


strehle:
@duanemay  This PR simply show the error in saml branch

peterhaochen47 (approved):

PR #2962, opened 2024-07-17
Fix regression in identity-provider endpoint

### LINKED ISSUE #2963: Regression in Identity-Provider Endpoint

UAA 77.13.0 with existing OIDC Idps. Call them via REST

uaac curl -X PUT /identity-providers/3ad529ed-3ef0-498d-b79e-9a6b0512ce8c

issuer is null

Reason: 
If no external id is set, then issuer is overwritten with null
https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/provider/JdbcIdentityProviderProvisioning.java#L215

- locally only using gradlew
- using cf-release
- using cf-deployment

### What did you do?

uaac curl -X PUT /identity-providers/3ad529ed-3ef0-498d-b79e-9a6b0512ce8c

issuer of OIDC IdP

### What did you see instead?

issuer null

strehle:
Issue:
If existing entries in identity-provider with new external_key the field is null, which is expected. If external_key is null, this must not overwrite the issuer in rest endpoint, but it does

For SAML there is no issue, because here the entityId is really new in REST output and in DB. For OIDC and OAuth2 the issuer was used in REST already and there was no check before overwrite it from external_key.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/JdbcIdentityProviderProvisioning.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/JdbcIdentityProviderProvisioning.java
@@ -1,6 +1,7 @@
 package org.cloudfoundry.identity.uaa.provider;
 
 import static java.sql.Types.VARCHAR;
+import static org.cloudfoundry.identity.uaa.util.UaaStringUtils.isNotEmpty;
 
 import org.cloudfoundry.identity.uaa.audit.event.SystemDeletable;
 import org.cloudfoundry.identity.uaa.constants.OriginKeys;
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/cloudfoundry/identity/uaa/provider/JdbcIdentityProviderProvisioning.java
```diff
@@ -218,15 +219,21 @@ public IdentityProvider mapRow(ResultSet rs, int rowNum) throws SQLException {
                 switch (identityProvider.getType()) {
                     case OriginKeys.SAML:
                         definition = JsonUtils.readValue(config, SamlIdentityProviderDefinition.class);
-                        Optional.ofNullable(definition).map(SamlIdentityProviderDefinition.class::cast).ifPresent(e -> e.setIdpEntityId(externId));
+                        if (isNotEmpty(externId)) {
```
adrianhoelzl-sap: Can we rename `externId` to `externalKey` to be consistent with the column name?

server/src/main/java/org/cloudfoundry/identity/uaa/provider/JdbcIdentityProviderProvisioning.java
```diff
@@ -218,15 +219,21 @@ public IdentityProvider mapRow(ResultSet rs, int rowNum) throws SQLException {
...
+                            Optional.ofNullable(definition).map(SamlIdentityProviderDefinition.class::cast).ifPresent(e -> e.setIdpEntityId(externId));
```
adrianhoelzl-sap: Instead of the surrounding `if` statement, we could also use `.filter(UaaStringUtils::isNotEmpty)` before the call to `ifPresent`.

For me, both is fine.

server/src/main/java/org/cloudfoundry/identity/uaa/provider/JdbcIdentityProviderProvisioning.java
strehle: done

server/src/main/java/org/cloudfoundry/identity/uaa/provider/JdbcIdentityProviderProvisioning.java
strehle: no this is not working since the externalKey is not part of the config JSON....

strehle:
Sonar

https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=2962

adrianhoelzl-sap (approved):

PR #2967, opened 2024-07-18
Fix Alias Validation when no Identity Zone ID is set in the Request Body

adrianhoelzl-sap:
During the validation of the alias properties, we perform the check whether an alias is created to the same zone as the original entity (in this case, the request is rejected).

Since the SCIM user update endpoint does not require the identity zone ID property to be set in the request body, we got a NullPointerException there.

This PR adds an additional validation step and ensures that the identity zone ID property is set in the user update call.


Tallicia (approved):
Additional tests and implementation looks good.

PR #2978, opened 2024-07-25
fix: Couple of failing test cases due to `500 INTERNAL_SERVER_ERROR` …

hsinn0:
…from `/oauth/token` endpoint

- Stepping through the server code revealed that an exception was thrown as follows:
```
org.cloudfoundry.identity.uaa.util.JsonUtils$JsonUtilException: com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "relyingPartyRegistrationId" (class org.cloudfoundry.identity.uaa.authentication.UaaPrincipal), not marked as ignorable (6 known properties: "origin", "zoneId", "id", "email", "externalId", "name"])
at [Source: REDACTED (StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION disabled); line: 1, column: 205] (through reference chain: org.cloudfoundry.identity.uaa.authentication.UaaPrincipal["relyingPartyRegistrationId"])
```
- Added a `jackson` annotation to ignore the 3 properties in UaaSamlPrincipal that were causing the `UnrecognizedPropertyException`.
- Added back a line that sets zoneId in a test case, which apparently had been removed by mistake.

[#187986233]
[#187986220]

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/UaaSamlPrincipal.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/UaaSamlPrincipal.java
@@ -14,6 +14,7 @@
 package org.cloudfoundry.identity.uaa.authentication;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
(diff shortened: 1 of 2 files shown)


hsinn0:
I did verify that all unit & integration tests were passing.

duanemay:
👍

PR #2980, opened 2024-07-27
feat: Return a sorted IdP list

strehle:
Sort the IdP list returned to thymeleaf
Sort always by login link text and ignore case of link text
Do not sort the JSON list but only HTML page

see
https://cloudfoundry.slack.com/archives/C03FXANBV/p1721916411577139

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java
@@ -451,7 +451,7 @@ public class LoginInfoEndpoint {
             boolean fieldUsernameShow,
             boolean linkCreateAccountShow
     ) {
-        Comparator sortingByLinkText = Comparator.comparing(SamlIdentityProviderDefinition::getLinkText);
+        Comparator<SamlIdentityProviderDefinition> sortingByLinkText = Comparator.comparing(SamlIdentityProviderDefinition::getLinkText);
         model.addAttribute(LINK_CREATE_ACCOUNT_SHOW, linkCreateAccountShow);
         model.addAttribute(FIELD_USERNAME_SHOW, fieldUsernameShow);
         model.addAttribute(IDP_DEFINITIONS, samlIdentityProviders.values().stream().sorted(sortingByLinkText).toList());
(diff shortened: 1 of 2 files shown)


markdboyd:
Wow, thanks so much @strehle ! This looks great!

server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java
```diff
@@ -450,12 +451,13 @@ private void updateLoginPageModel(
             boolean fieldUsernameShow,
             boolean linkCreateAccountShow
     ) {
+        Comparator<SamlIdentityProviderDefinition> sortingByLinkText = Comparator.comparing(SamlIdentityProviderDefinition::getLinkText);
```
bengerman13: This looks like it's a case-sensitive sort, right? Could we consider a case-insensitive sort here?

I'm not fluent in Java, but the docs read like it might be this simple:

```suggestion
        Comparator<SamlIdentityProviderDefinition> sortingByLinkText = Comparator.comparing(SamlIdentityProviderDefinition::getLinkText, String.CASE_INSENSITIVE_ORDER);
```

bengerman13: (also thanks! this is a great improvement for our users!)

strehle: ok

server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java
```diff
@@ -450,12 +451,13 @@ private void updateLoginPageModel(
...
     ) {
+        Comparator<SamlIdentityProviderDefinition> sortingByLinkText = Comparator.comparing(SamlIdentityProviderDefinition::getLinkText, String.CASE_INSENSITIVE_ORDER);
         model.addAttribute(LINK_CREATE_ACCOUNT_SHOW, linkCreateAccountShow);
         model.addAttribute(FIELD_USERNAME_SHOW, fieldUsernameShow);
-        model.addAttribute(IDP_DEFINITIONS, samlIdentityProviders.values());
+        model.addAttribute(IDP_DEFINITIONS, samlIdentityProviders.values().stream().sorted(sortingByLinkText).toList());
         Map<String, String> oauthLinks = new HashMap<>();
         ofNullable(oauthIdentityProviders).orElse(emptyMap()).entrySet().stream()
-                .filter(e -> e.getValue().isShowLinkText())
+                .filter(e -> e.getValue().isShowLinkText() && e.getValue() != null && e.getKey() != null)
```
coolgang123: The condition needs to be rewritten as follows. The null check should happen before using the value.

e.getValue() != null && e.getValue().isShowLinkText() && e.getKey() != null

strehle: Thanks

coolgang123 (approved):

PR #2983, opened 2024-07-29
clean-up: IdentityZone samlConfig properties for UAA-as-SAML-IdP feature

hsinn0:
- As they are now obsolete because the UAA-as-SAML-IDP feature had been removed.
- Removed the properties from samlConfig.

[#187986019]


strehle (approved):
Lgtm sure

PR #2988, opened 2024-07-31
feature: Handle icorrect SAML response

hsinn0:
- Set the `Saml2WebSsoAuthenticationFilter`'s `AuthenticationFailureHandler` to the custom failure handler.
- Updated the test case's page source validation condition to check for the string that is based on the new exception message.

[#187986112]


hsinn0:
I verified that all unit & integration tests passed locally.

PR #3006, opened 2024-08-16
fix: use up-to-date jwt sigining key config fields in local UAA.yml

peterhaochen47:
- replace legacy config fields in favor of more up-to-date jwt sigining key config in UAA.yml
used in local server run
- new fields as documeted in: https://github.com/cloudfoundry/uaa/blob/fe95b5c71a55b264e3d1ab7899904c096a7224d8/docs/Sysadmin-Guide.rst#L152


coolgang123 (approved):

peterhaochen47:
@strehle, thoughts on the `PrivateKeyJwtClientAuthIT` failures? Shouldn't the new fields be equivalent to the legacy ones? And why would private key jwt client auth flow be dependent on what UAA's own jwt signing key is (I thought the client auth flow is using a different signing key specific to that client)?

strehle:
Thanks for the PR

I will check failure next week, but it's good to use new structure

strehle (approved):

PR #3014, opened 2024-08-23
Fix Update of Users with Alias during Login

adrianhoelzl-sap:
Whenever a user logs in, their properties, e.g., `externalId`, are updated with the corresponding values from the IdP (e.g., the ID token in case of OIDC).

This update operation does not work for users with an alias. Such users are synchronized with an identical user in a different identity zone and reference each other through the properties `aliasId` and `aliasZid`. 

Since the update operation performed during the login does not handle these alias properties correctly, the user is updated without the changes being propagated to the alias. Furthermore, the alias properties are set to `null`, which breaks the alias reference from this side.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/scim/bootstrap/ScimUserBootstrap.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/scim/bootstrap/ScimUserBootstrap.java
@@ -179,8 +179,13 @@ public class ScimUserBootstrap implements
             newScimUser.setAliasZid(existingUser.getAliasZid());
         }
 
-        // this will also handle the update of the alias user, if necessary
-        scimUserService.updateUser(id, newScimUser);
+        if (aliasEntitiesEnabled) {
+            // update the user and propagate the changes to the alias, if present
+            scimUserService.updateUser(id, newScimUser);
+        } else {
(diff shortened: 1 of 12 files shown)


hsinn0 (approved):
I did not find any outstanding issues.

server/src/main/java/org/cloudfoundry/identity/uaa/scim/bootstrap/ScimUserBootstrap.java
```diff
@@ -160,7 +164,18 @@ private void updateUser(ScimUser existingUser, UaaUser updatedUser, boolean upda
...
+
+        /* the user in the event won't have the alias properties set, we must therefore propagate them from the existing
+         * user, if present */
+        if (hasText(existingUser.getAliasId()) && hasText(existingUser.getAliasZid())) {
+            newScimUser.setAliasId(existingUser.getAliasId());
+            newScimUser.setAliasZid(existingUser.getAliasZid());
+        }
+
+        // this will also handle the update of the alias user, if necessary
+        scimUserService.updateUser(id, newScimUser);
```
adrianhoelzl-sap: How is the AliasPropertiesInvalidException handled here? -> ScimException

klaus-sap (approved):
lgtm

PR #3025, opened 2024-09-04
Add Response Status Code Checks to IntegrationTestUtils

### LINKED ISSUE #2889: Missing Response Status Checks in IntegrationTestUtils

adrianhoelzl-sap:
closes #2889

### DIFF
--- a/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/LdapIntegrationTests.java
+++ b/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/LdapIntegrationTests.java
@@ -58,7 +58,7 @@ public class LdapIntegrationTests {
         String token = IntegrationTestUtils.getClientCredentialsToken(serverRunning.getBaseUrl(), "admin", "adminsecret");
 
         ScimGroup group = new ScimGroup(null, "zones.testzone1.admin", null);
-        IntegrationTestUtils.createGroup(token, "", serverRunning.getBaseUrl(), group);
+        IntegrationTestUtils.createGroupAndIgnoreStatusCode(token, "", serverRunning.getBaseUrl(), group);
     }
 
     @After
--- a/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/LdapIntegrationTests.java
+++ b/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/LdapIntegrationTests.java
@@ -58,7 +58,7 @@ public class LdapIntegrationTests {
         String token = IntegrationTestUtils.getClientCredentialsToken(serverRunning.getBaseUrl(), "admin", "adminsecret");
 
         ScimGroup group = new ScimGroup(null, "zones.testzone1.admin", null);
-        IntegrationTestUtils.createGroupAndIgnoreStatusCode(token, "", serverRunning.getBaseUrl(), group);
+        IntegrationTestUtils.createGroupAndIgnoreConflict(token, "", serverRunning.getBaseUrl(), group);
     }
 
     @After
--- a/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/LdapIntegrationTests.java
+++ b/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/LdapIntegrationTests.java
@@ -22,7 +22,6 @@ import org.cloudfoundry.identity.uaa.integration.util.IntegrationTestUtils;
 import org.cloudfoundry.identity.uaa.oauth.jwt.Jwt;
 import org.cloudfoundry.identity.uaa.provider.LdapIdentityProviderDefinition;
 import org.cloudfoundry.identity.uaa.oauth.token.ClaimConstants;
-import org.cloudfoundry.identity.uaa.scim.ScimGroup;
 import org.cloudfoundry.identity.uaa.scim.ScimUser;
 import org.cloudfoundry.identity.uaa.test.UaaTestAccounts;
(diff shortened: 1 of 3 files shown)


peterhaochen47:
@adrianhoelzl-sap Makes sense overall, thanks!
I tried and saw that for lines like this one: https://github.com/cloudfoundry/uaa/blob/bdec670938373f688e1c9ab7f680c74290436828/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/SamlLoginIT.java#L162

If you narrow it to just [IntegrationTestUtils.createGroup](https://github.com/cloudfoundry/uaa/blob/bdec670938373f688e1c9ab7f680c74290436828/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/util/IntegrationTestUtils.java#L571) (aka do not ignore conflicts but still expect `HttpStatus.CREATED`), the tests will still be passing. Why ignore? Maybe we should use the more strict check (expect `HttpStatus.CREATED` only)?

adrianhoelzl-sap:
> @adrianhoelzl-sap Makes sense overall, thanks! I tried and saw that for lines like this one:
> 
> https://github.com/cloudfoundry/uaa/blob/bdec670938373f688e1c9ab7f680c74290436828/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/SamlLoginIT.java#L162
> 
> If you narrow it to just [IntegrationTestUtils.createGroup](https://github.com/cloudfoundry/uaa/blob/bdec670938373f688e1c9ab7f680c74290436828/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/util/IntegrationTestUtils.java#L571) (aka do not ignore conflicts but still expect `HttpStatus.CREATED`), the tests will still be passing. Why ignore? Maybe we should use the more strict check (expect `HttpStatus.CREATED` only)?

For me, the tests fail because the groups already exist, maybe they are created already by previous tests. You can see the error messages in the voters for commit [c51ed9f](https://github.com/cloudfoundry/uaa/pull/3025/commits/c51ed9fb524b66a99db5b9f9cddf52261c6f5994).
IMO, it is fine to accept the conflict status code here, since we just want to ensure that the groups are present. If they already are, we can ignore the error.

adrianhoelzl-sap:
> @adrianhoelzl-sap Makes sense overall, thanks! I tried and saw that for lines like this one:
> 
> https://github.com/cloudfoundry/uaa/blob/bdec670938373f688e1c9ab7f680c74290436828/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/SamlLoginIT.java#L162
> 
> If you narrow it to just [IntegrationTestUtils.createGroup](https://github.com/cloudfoundry/uaa/blob/bdec670938373f688e1c9ab7f680c74290436828/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/util/IntegrationTestUtils.java#L571) (aka do not ignore conflicts but still expect `HttpStatus.CREATED`), the tests will still be passing. Why ignore? Maybe we should use the more strict check (expect `HttpStatus.CREATED` only)?

@peterhaochen47 I made the behavior of the `createGroup` calls in `SamlLoginIT` more explicit by adding a method `ensureGroupExists` and removing the `createGroupAndIgnoreConflict` method (commit [7fc64b2](https://github.com/cloudfoundry/uaa/pull/3025/commits/7fc64b222e7939fd32c9fdce874ea10a370f86ab)).

peterhaochen47 (approved):

PR #3026, opened 2024-09-04
Refactor ScimUserProvisioning Dependency Injection

adrianhoelzl-sap:


server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java
```diff
@@ -131,8 +130,6 @@ public Logger getLogger() {
 
     private static final RowMapper<ScimUser> mapper = new ScimUserRowMapper();
 
-    private Pattern usernamePattern = Pattern.compile("[\\p{L}+0-9+\\-_.@'!]+");
```
peterhaochen47, line 134: Could you provide some context on how this is now an obsolete property. Is this property obsolete globally in UAA, or just obsolete in `JdbcScimUserProvisioning`?

If the former (globally), this property still exist in these places and need to be cleaned up:
- https://github.com/cloudfoundry/uaa/blob/65952b1b53b8d01cf93e68493a3f6ac85ad8a825/uaa/src/test/resources/test/config/uaa.yml#L23
- https://github.com/cloudfoundry/uaa/blob/4fba1e4c0142691ab25cb313d60dd26b53774686/server/src/main/java/org/cloudfoundry/identity/uaa/impl/config/UaaConfiguration.java#L233

adrianhoelzl-sap: Thanks for the hint, I removed the variable in commit c7c8bdc.

peterhaochen47 (approved):

PR #3062, opened 2024-09-23
Test-Fix of chromdriver129 issue

strehle:
With Chrome 128 we had an issue already, solved with https://github.com/cloudfoundry/uaa/pull/3020
Now again , but another issue.

This option
--headless=old

uses old behavrior, therefore revert PR https://github.com/cloudfoundry/uaa/pull/3020 and set window-position.

With newer Chrome that might be refactored... but for now use these settings because of

- https://github.com/SeleniumHQ/selenium/issues/14514
- https://issues.chromium.org/issues/367755364
- https://stackoverflow.com/questions/78996364/chrome-129-headless-shows-blank-window

### DIFF
--- a/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/pageObjects/Page.java
+++ b/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/pageObjects/Page.java
@@ -1,6 +1,7 @@
 package org.cloudfoundry.identity.uaa.integration.pageObjects;
 
 import java.time.Duration;
+import java.util.concurrent.TimeUnit;
 
 import org.hamcrest.Matcher;
 import org.openqa.selenium.By;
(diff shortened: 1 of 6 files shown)


uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/pageObjects/Page.java
```diff
@@ -49,4 +50,11 @@ private void clickLogout() {
     public void clearCookies() {
         driver.manage().deleteAllCookies();
     }
+
+    public static void validateUrlStartsWithWait(WebDriver driver, String currentUrlStart) throws InterruptedException {
+        if (!driver.getCurrentUrl().startsWith("https://www.google.com")) {
```
peterhaochen47: Is this line supposed to be `if (!driver.getCurrentUrl().startsWith(currentUrlStart)) {`?

strehle: Yes because if the SAML login takes too long then the current URL is not the expected one.

Therefore ,wait for 5 sec. And then, if still not expected one then fail

peterhaochen47: I see. The current code in this PR is hardcoded to expect `https://www.google.com` though. I was just asking if the code should expect the `currentUrlStart` from the function param.

strehle: ups... sorry, now I see it. Yes I tested many times and behavior in github action and concourse pipeline was always different so.... these waits should help , but should be only done the described situations

Thanks for the hint

peterhaochen47 (approved):

Issue #2561, opened 2023-10-19
Spring Security SAML2 End of Life 


strehle:
UAA uses https://github.com/spring-attic/spring-security-saml
This was no issues in the past but with the Spring Boot 3 update the library needs to be changed or removed

Similar issue is with opensaml2 from shibboleth

Issue #2115, opened 2022-11-25
Spring Boot 3x time line 


strehle:
### What version of UAA are you running?

dev

### How are you deploying the UAA?

I am deploying the UAA

- locally only using gradlew
- using a bosh release I downloaded from bosh.io
- using cf-release
- using cf-deployment
- as part of a commercial Cloud Foundry distribution
- other (please explain)

### What did you do?

spring boot 2.7x used

### What did you expect to see? What goal are you trying to achieve with the UAA?

See 
https://github.com/cloudfoundry/uaa/pull/2113

### What did you see instead?

Please include UAA logs if available.

DanSalt:
Hi @strehle -- any updates on the status/timeline of this?

strehle:
> Hi @strehle -- any updates on the status/timeline of this?

Not concrete but since 27x will be deprecated end of year we plan to move to new version then. FYI @Tallicia

strehle:
Update.

UAA team is working on Java 17 runtime in all pipelines. FYI @Tallicia 

Then we can start working on it:
https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0-Migration-Guide

Next step : running and validation UAA on tomcat 10.x 

Plan is to have it in Q4 this year, maybe with some weeks of overlap of ending OSS support https://endoflife.date/spring-boot but only a few

strehle:
additional information for upgrade

removed dependency https://github.com/cloudfoundry/uaa/pull/2611

Blog
https://spring.io/blog/2022/05/24/preparing-for-spring-boot-3-0

Migration Tool
https://github.com/spring-projects-experimental/spring-boot-migrator

jan872:
Hi - any news or updates on this issue?

Tallicia:
@jan872 This is progressing, currently with significant SAML library replacement pre requisite work underway.

strehle:
merged https://github.com/cloudfoundry/uaa/pull/3517, thus close it because next release build will contain spring boot 3

PR #3076, opened 2024-10-07
Upgrade to boot 3.0

rpau:
Hi, 

This is an experiment running an automatic upgrade to adopt Spring Boot 3.0 using Tanzu Spring Application Advisor. It is still work in progress. Some of the learnings we are accumulating for this example are:

* This project can't be upgraded to boot 3.0 until spring-security-saml-extensions are not migrated to Spring security. https://spring.io/blog/2020/09/22/spring-security-saml-extensions-1-x-eol-on-october-6-2021 .

* On the other hand, looks like you are using spring-security-saml 1.0.x instead of 2.0.x, which uses Spring security 5.2.x. This means, that in our end, the application requires the following upgrade plan for upgrading until boot 3.0.

Step 1:
    *  Upgrade spring-security-saml 1.0.x to 2.0.x
Step 2:
    * Replace spring-security-saml 2.0.x for spring-security 5.8.x
Step 3:
    * Upgrade spring-framework from 5.3.x to 6.0.x
    * Upgrade spring-security from 5.8.x to 6.0.x
    * Upgrade spring-ldap from 2.4.x to 3.0.x
    * Upgrade spring-boot from 2.7.x to 3.0.x
    * Upgrade spring-retry from 1.3.x to 2.0.x

* The dependency versions can't be upgraded in the way they are defined, but looks like the main reason is to overwrite the dependency resolution. This means that dependency versions should be changed manually.


strehle:
FYI: https://github.com/cloudfoundry/uaa/issues/2561 , https://github.com/cloudfoundry/uaa/issues/2115

strehle:
@rpau we will do the upgrade soon, now all legacy parts are removed

rpau:
I am sending a PR at this moment.

rpau:
https://github.com/cloudfoundry/uaa/pull/3171

PR #3078, opened 2024-10-07
Allow Dynamic Value Lookup in Custom Zone for Private Key JWT when IdP has Alias

### LINKED ISSUE #3077: Login via Alias IdP in custom IdZ not possible when using Private Key JWT with dynamic Key/Cert Lookup
UAA supports private key JWT as a client authentication method for the calls to the external IdP during OpenID Connect
flows with user involvement (i.e., password grant and authorization code flow).
The identity provider can be configured to use a key and certificate injected into the UAA environment for building the
client assertion.
This is done by referencing the path of the key and/or certificate (defined in `uaa.yml`) in the
`jwtClientAuthentication` property of its config.
For example, the IdP config field `jwtClientAuthentication.key` can be set to `${jwt.client.key}` to use the private key
defined in the `uaa.yml` in the property `jwt.client.key`.

Such dynamic lookups are only allowed **in the "uaa" zone**.

Using the dynamic value lookup feature for an IdP in the "uaa" zone leads to failing OIDC flows of its alias in a custom zone.
Since an IdP and its alias have identical configurations, the alias IdP will also reference the dynamic value.

Whenever OIDC flows are performed with the alias IdP in the custom zone, looking up the dynamic value will be rejected and the overall flow fails as no key was found.

adrianhoelzl-sap:
closes issue #3077

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/manager/PasswordGrantAuthenticationManager.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/manager/PasswordGrantAuthenticationManager.java
@@ -39,7 +39,6 @@ import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.util.Base64Utils;
 import org.springframework.util.LinkedMultiValueMap;
 import org.springframework.util.MultiValueMap;
-import org.springframework.util.StringUtils;
 import org.springframework.web.client.HttpClientErrorException;
 import org.springframework.web.client.RestTemplate;
 
@@ -49,6 +48,7 @@ import java.util.function.Supplier;
 
 import static org.cloudfoundry.identity.uaa.oauth.token.TokenConstants.GRANT_TYPE_PASSWORD;
 import static org.springframework.http.MediaType.APPLICATION_JSON;
+import static org.springframework.util.StringUtils.hasText;
 
 public class PasswordGrantAuthenticationManager implements AuthenticationManager, ApplicationEventPublisherAware {
 
@@ -105,7 +105,7 @@ public class PasswordGrantAuthenticationManager implements AuthenticationManager
         if (identityProvider == null || loginHintToUse == null || loginHintToUse.getOrigin() == null || loginHintToUse.getOrigin().equals(OriginKeys.UAA) || loginHintToUse.getOrigin().equals(OriginKeys.LDAP)) {
             return zoneAwareAuthzAuthenticationManager.authenticate(authentication);
         } else {
-            return oidcPasswordGrant(authentication, identityProvider.getConfig());
+            return oidcPasswordGrant(authentication, identityProvider);
         }
     }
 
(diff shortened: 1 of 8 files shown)


server/src/main/java/org/cloudfoundry/identity/uaa/authentication/manager/PasswordGrantAuthenticationManager.java
```diff
@@ -175,8 +177,11 @@ Authentication oidcPasswordGrant(Authentication authentication, OIDCIdentityProv
         MultiValueMap<String, String> params = new LinkedMultiValueMap<>();
 
         if (ClientAuthentication.PRIVATE_KEY_JWT.equals(calcAuthMethod)) {
+            /* ensure that the dynamic lookup of the cert and/or key for private key JWT works for an alias IdP in a
+             * custom IdZ */
+            final boolean allowDynamicValueLookupInCustomZone = hasText(identityProvider.getAliasZid()) && hasText(identityProvider.getAliasId());
```
strehle, line 182: please add aliasEntitiesEnabled check and best use it from https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/alias/EntityAliasHandler.java#L22

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/jwt/JwtClientAuthentication.java
```diff
@@ -74,33 +75,53 @@ public JwtClientAuthentication(KeyInfoService keyInfoService, OidcMetadataFetche
     this.oidcMetadataFetcher = oidcMetadataFetcher;
   }
 
-  public String getClientAssertion(OIDCIdentityProviderDefinition config) {
+  @VisibleForTesting
```
strehle: why such annotations... method is public

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/jwt/JwtClientAuthentication.java
```diff
@@ -74,33 +75,53 @@ public JwtClientAuthentication(KeyInfoService keyInfoService, OidcMetadataFetche
...
+  public String getClientAssertion(final OIDCIdentityProviderDefinition config) {
+    return getClientAssertion(config, false);
+  }
+
+  @VisibleForTesting
```
strehle: make it public

adrianhoelzl-sap: fixed with commit 68d3fdaf8389acf203480d331ad01fd93aa73e87

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/jwt/JwtClientAuthentication.java
adrianhoelzl-sap: fixed with commit 68d3fdaf8389acf203480d331ad01fd93aa73e87

server/src/main/java/org/cloudfoundry/identity/uaa/authentication/manager/PasswordGrantAuthenticationManager.java
adrianhoelzl-sap: I intentionally did not add the `aliasEntitiesEnabled` check here. Whenever an IdP with alias (that also references a dynamic value) exists and `aliasEntitiesEnabled` is deactivated afterwards, the check would lead to failing logins.

In general, the `aliasEntitiesEnabled` flag should only prevent new aliases to be created and existing ones to be updated. It should not block entities with alias from working in, e.g., login flows.

hsinn0 (approved):

strehle (approved):

PR #3084, opened 2024-10-12
fix issue 3083: check user_name claim type

### LINKED ISSUE #3083: Again error from IdP if mapped claim is no string

develop

I am deploying the UAA

- locally only using gradlew
- using cf-release
- using cf-deployment
- other (please explain)

### What did you do?

Configure external IdP where an array is part of id_token, e.g.

email = [ "one", "two" ]

And email is the user_name. Then we ran into same issue then https://github.com/cloudfoundry/uaa/issues/1831

Either a better error, similar to https://github.com/cloudfoundry/uaa/issues/1831 or a fix like https://github.com/cloudfoundry/uaa/pull/1925

### What did you see instead?

strehle:
solve https://github.com/cloudfoundry/uaa/issues/3083

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
@@ -238,11 +238,11 @@ public class ExternalOAuthAuthenticationManager extends ExternalLoginAuthenticat
 
             String userNameAttributePrefix = (String) attributeMappings.get(USER_NAME_ATTRIBUTE_NAME);
             String username;
-            if (StringUtils.hasText(userNameAttributePrefix)) {
-                username = (String) claims.get(userNameAttributePrefix);
+            if (hasText(userNameAttributePrefix)) {
+                username = getMappedClaim(userNameAttributePrefix, USER_NAME_ATTRIBUTE_NAME, claims);
                 logger.debug(String.format("Extracted username for claim: %s and username is: %s", userNameAttributePrefix, username));
             } else {
-                username = (String) claims.get(SUB);
+                username = getMappedClaim(null, SUB, claims);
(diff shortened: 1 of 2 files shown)


strehle:
after first comming

ClassCastException: class java.util.ArrayList cannot be cast to class java.lang.String (java.util.ArrayList and java.lang.String are in module java.base of loader 'bootstrap')

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
```diff
@@ -424,7 +424,7 @@ private String getMappedClaim(String externalName, String internalName, Map<Stri
             return (String) claimObject;
         }
         if (claimObject instanceof Collection) {
-            Set<String> entry = ((Collection<?>) claimObject).stream().map(String.class::cast).collect(Collectors.toSet());
+            Set<String> entry = ((Collection<?>) claimObject).stream().filter(String.class::isInstance).map(String.class::cast).collect(Collectors.toSet());
```
coolgang123, line 427: Will collection items always going to be a string? Should we consider handle it in more generic way:

Set<String> entry = ((Collection<?>) claimObject).stream()
        .map(claim -> String.valueOf(claim))
        .collect(Collectors.toSet());

strehle: I think it is not wanted to have such a generic approach, but simple check if the entries inside of the token are strings. With your approach we allow all but then we are incompatible with now. I mean if claims are type of Boolean or int we would allow them now. But that is not wanted

strehle:
Wanted to solve is that there is a human readable error and not a technical class cast error. The error should show what mapping is wrong, eg. email , first name , user_name etc.

coolgang123 (approved):

adrianhoelzl-sap (approved):

PR #3127, opened 2024-11-12
Add acr value into User Authentication

strehle:
re-establish IT

see former retrieval
https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/LoginSamlAuthenticationProvider.java#L292-L298


duanemay (approved):
Thank You!

PR #3129, opened 2024-11-12
Cleanup libraries not needed anymore

strehle:
bound to old opensaml

here is the old deps. list
https://mvnrepository.com/artifact/org.springframework.security.extensions/spring-security-saml2-core/1.0.10.RELEASE


duanemay (approved):
👍

PR #3133, opened 2024-11-14
feature: ingtegration test coverage

hsinn0:
- Added a task to get the report.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -124,6 +124,44 @@ subprojects {
         }
     }
 
+    // Integration teset coverage
+    // - To get integration test only coverage, do
+    //   `./gradlew clean integrationTest` first. This will create
+    //   `integrationTest.exec` files in subproject `build/jacoco/` folders.
+    //   Here, `clean` is for removing any `unit-test` coverage data and reports
+    //   generated from previous runs of `./gradlew test` if you have any,
+    //   so they are not included in the aggregated report.
+    // - To get integration test coverage from the generated
+    //   `integrationTest.exec` files, do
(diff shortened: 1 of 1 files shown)


hsinn0:
When running the new Gradle task as described in the `build.gradle`, I get:
```
\> Task :cloudfoundry-identity-uaa:jacocoLogIntegrationTestCoverage
Test Coverage:
    - Class Coverage: 13%
    - Method Coverage: 14.4%
    - Branch Coverage: 3.7%
    - Line Coverage: 8.6%
    - Instruction Coverage: 6.9%
    - Complexity Coverage: 8.4%
```
```
\> Task :jacocoLogAggregatedCoverage
Test Coverage:
    - Class Coverage: 14%
    - Method Coverage: 15.2%
    - Branch Coverage: 4%
    - Line Coverage: 9.3%
    - Instruction Coverage: 7.5%
    - Complexity Coverage: 9%
```
~You can also get the coverage number from IntelliJ Idea by running `integrationTest` gradle task there with Run with Coverage option there. In my case, the numbers from Idea was slightly different from the gradle task result above, but close enough.~ There is an option to run `integrationTest` task with Coverage in IntelliJ Idea. However, the coverage report from it was not what we expected.  It was not showing coverage for the app server code path that was being executed.

hsinn0:
This is not working.  The JaCoCo coverage data is collected not from app server but from the integration test process itself. Looking at our `integrationTest` definition, it should have been somewhat obvious but I missed that. Will have to change to instrument cargo.local process instead.

PR #3132, opened 2024-11-13
Enhancements for SAML2 bearer flow

strehle:

PR #3136, opened 2024-11-15
Enhancements for SAML2 bearer and IdP initiated SSO

strehle:
FYI create also a full PR because sonar runs and helps to cleanup code, e.g.
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=3135

Includes https://github.com/cloudfoundry/uaa/pull/3132

In addition now created DefaultRelyingPartyRegistrationResolver which retrieves the correct registry / idp trust for the issuer.

We dont need the default return.

Open:
 a) remove dummy -> in preparation
 b) remove classpath: from idpMetadata , not need if a) is solved
 c) refactorings, e.g. move the XML parsing from SAML2 bearer class into Saml2Utils


server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/DefaultRelyingPartyRegistrationResolver.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.Map;
+import java.util.function.Function;
+
+/**
+ * This was copied from Spring Security, and modified to work with Open SAML 4.0.x
+ * <p/>
+ * We may keep this implementation, because it resolves the correct SamlIdp with relyingPartyRegistrationId==null
+ */
+@Slf4j
+public final class DefaultRelyingPartyRegistrationResolver implements Converter<HttpServletRequest, RelyingPartyRegistration>, RelyingPartyRegistrationResolver {
```
duanemay: How about renaming this to UaaRelyingPartyRegistrationResolver to not confuse it with the spring-sec one

server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/SamlAuthenticationFilterConfig.java
```diff
@@ -225,8 +226,9 @@ Saml2LogoutResponseFilter saml2LogoutResponseFilter(RelyingPartyRegistrationReso
     @Bean
     Saml2LogoutRequestFilter saml2LogoutRequestFilter(RelyingPartyRegistrationRepository relyingPartyRegistrationRepository,
                                                       UaaAuthenticationFailureHandler authenticationFailureHandler,
-                                                      CookieBasedCsrfTokenRepository loginCookieCsrfRepository) {
-        RelyingPartyRegistrationResolver relyingPartyRegistrationResolver = new DefaultRelyingPartyRegistrationResolver(relyingPartyRegistrationRepository);
```
duanemay, line 229: should be able to autowire a DefaultRelyingPartyRegistrationResolver instead of creating a new one, in a few places in this config class

duanemay (approved):

PR #3138, opened 2024-11-15
Replace dummy-saml-idp-metadata

strehle:
and create the data based on real key data

Until now we do not deliver any keys in uaa.war.


duanemay (approved):
Awesome!

PR #3140, opened 2024-11-16
omit hard coded example name

strehle:


server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/SamlMetadataEndpoint.java
```diff
@@ -12,17 +12,17 @@
...
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
+import java.util.UUID;
 
 @RestController
 public class SamlMetadataEndpoint implements ZoneAware {
-    public static final String DEFAULT_REGISTRATION_ID = "example";
+    protected static final String DEFAULT_REGISTRATION_ID = UUID.randomUUID().toString();
```
strehle, line 25: with this there are no clashes in case you define IdPs in uaa.yml

duanemay (approved):

PR #3141, opened 2024-11-18
Improve Performance of LoginInfoEndpoint

adrianhoelzl-sap:
The method `SamlIdentityProviderConfigurator.getIdentityProviderDefinitionsForZone` (called during the evaluation of a request to `/login`) returns the configurations of all SAML IdPs in an identity zone. Instead of specifying a condition on the `type` column of the `identity_provider` table, it reads all (active) IdPs of the identity zone and filters them in the Java code.

This PR fixes the lookup of SAML IdPs by moving the filtering for the SAML type to the query and introducing an index on the columns `identity_zone_id`, `active` and `type` of the `identity_provider` table.


duanemay (approved):
👍

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthProviderConfigurator.java
```diff
@@ -179,6 +181,30 @@ public List<IdentityProvider> retrieveActive(String zoneId) {
...
     }
 
+    @Override
+    public List<IdentityProvider> retrieveActiveByType(final String type, final String zoneId) {
+        if (!OAUTH20.equals(type) && !OIDC10.equals(type)) {
+            return emptyList();
+        }
+        return providerProvisioning.retrieveActiveByType(type, zoneId).stream()
+                .map(idp -> {
+                    if (idp.getType().equals(OIDC10)) {
```
strehle: very similar to retrievAll, can we combine it. ?
maybe with type = 1 or type = 2.
The index should cover it

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthProviderConfigurator.java
```diff
@@ -179,6 +181,30 @@ public List<IdentityProvider> retrieveActive(String zoneId) {
...
+                            );
+                            idp.setConfig(overlayedDefinition);
+                        } catch (final Exception e) {
+                            LOGGER.error("Identity provider excluded from login page due to a problem.", e);
+                            return null;
+                        }
+                    }
+                    return idp;
+                })
+                .filter(not(Objects::isNull))
```
strehle: .filter(Objects::nonNull)

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthProviderConfigurator.java
```diff
@@ -179,6 +181,30 @@ public List<IdentityProvider> retrieveActive(String zoneId) {
         return retrieveAll(true, zoneId);
     }
 
+    @Override
+    public List<IdentityProvider> retrieveActiveByType(final String type, final String zoneId) {
```
strehle: this method is never call in /login but only retrieveAll

adrianhoelzl-sap: I added this method to `ExternalOAuthProviderConfigurator` because it needs to correctly implement the interface (where the new method was added).

`retrieveActiveByType` has two differences to the method `retrieveAll` in this class:

- `retrieveAll` allows passing a parameter `activeOnly`; `retrieveActiveByType` does not allow querying for inactive IdPs
- `retrieveAll` returns the IdPs of type `oidc1.0` **and** `oauth2.0`; `retrieveActiveByType` only allows filtering for one type

I could address these two points by ...

1. changing the method to `retrieveByType` and adding a similar `activeOnly` parameter and
2. allowing multiple types to be passed to the method.

What do you think?

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthProviderConfigurator.java
adrianhoelzl-sap: fixed with commit ff96103

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthProviderConfigurator.java
adrianhoelzl-sap: fixed with commit 8b29079

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthProviderConfigurator.java
strehle: This sounds reasonable, 

- retrieveAll returns the IdPs of type oidc1.0 and oauth2.0; retrieveActiveByType only allows filtering for one type

because the retrieveAll is used in realality and retrieveActiveByType is not used but the filter might be ok. Maybe if type is null return simply 
`oidc1.0 and oauth2.0`

adrianhoelzl-sap: fixed with commit 7eefd26

strehle (approved):

duanemay (approved):

PR #3153, opened 2024-11-22
fix: default values of custom zone's saml entityID and saml alias (wh…

peterhaochen47:
…en the configured entityID is a URL)

- maintain the existing behavior where a custom identity zone's saml entityID is defaulted to either 1) `zoneSubdomain.uaaWideSamlEntityID` if `uaaWideSamlEntityID` is not a URL, or 2) if `uaaWideSamlEntityID` is a URL, integration the zoneSubdomain into the URL (see tests for example).

- similar logic for saml entity alias (which is used in various saml sp urls, such as `AssertionConsumerService`) except that the alias should not include url scheme (aka without `https://`), so that the resulting saml sp urls are valid urls (e.g.: `https://zone1.uaa.com/saml/SSO/alias/[saml entity alias]`, see tests for examples).

- reference on develop branch (old saml code):
  - doc: https://github.com/cloudfoundry/uaa/blob/65952b1b53b8d01cf93e68493a3f6ac85ad8a825/docs/login/Okta-README.md?plain=1#L73-L75
  - code: https://github.com/cloudfoundry/uaa/blob/cc5f76fba495e5d1b3fd755ac3a6ff137fc91878/server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/ZoneAwareMetadataGenerator.java#L53-L54

# problem statement
without this commit, when
* a custom zone is created without a `zone.config.samlConfig.entityID` specified
* the default zone's `login.entityID` is configured to a URL, such as `https://uaa.com`
* the default zone's `login.saml.entityIDAlias` is not set, aka default to `login.entityID` Then the resulting custom zone sp metadata has some discrepancies with the old saml code's metadata:

## For `AssertionConsumerService`:
- old (correct) value is: https://test-zone-before.uaa.com/saml/SSO/alias/test-zone-before.uaa.com
- new value is: https://test-zone.uaa.com/saml/SSO/alias/test-zone.http:/uaa.com 
## For `entityID`:
- old (correct) value is: http://test-zone-before.uaa.com
- new value is: test-zone.http://uaa.com

This results in the external SAML login for this zone not working.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/BaseUaaRelyingPartyRegistrationRepository.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/BaseUaaRelyingPartyRegistrationRepository.java
@@ -1,6 +1,7 @@
 package org.cloudfoundry.identity.uaa.provider.saml;
 
 import lombok.extern.slf4j.Slf4j;
+import org.cloudfoundry.identity.uaa.util.UaaUrlUtils;
 import org.cloudfoundry.identity.uaa.zone.IdentityZone;
 import org.cloudfoundry.identity.uaa.zone.IdentityZoneConfiguration;
 import org.cloudfoundry.identity.uaa.zone.SamlConfig;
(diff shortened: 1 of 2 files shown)


duanemay (approved):
👍

PR #3161, opened 2024-11-27
Improve Performance of IdP Lookup during Password Grant

adrianhoelzl-sap:
Optimize the lookup of IdPs during password grant flow: whenever "uaa" or "ldap" is used as a login hint, we no longer read all active IdPs in the identity zone but rather only the single one.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/manager/PasswordGrantAuthenticationManager.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/manager/PasswordGrantAuthenticationManager.java
@@ -97,7 +97,7 @@ public class PasswordGrantAuthenticationManager implements AuthenticationManager
 
         // determine the IdP to use from the list of possible ones
         UaaLoginHint loginHintToUse;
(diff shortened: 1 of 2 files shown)


strehle:
please have a look on Refactoring from https://github.com/cloudfoundry/uaa/pull/3165

this PR here increases the code complexity but I tried in https://github.com/cloudfoundry/uaa/pull/3165 to reduce complexity and removed the not needed interface: 
ExternalOAuthProviderConfigurator externalOAuthProviderProvisioning

because we have the generic one with "IdentityProviderProvisioning identityProviderProvisioning"

In 2nd PR I added your tests, see https://github.com/cloudfoundry/uaa/pull/3168

because for retrieveByOrigin it does not matter if externalOAuthProviderProvisioning or identityProviderProvisioning.
Both call providerProvisioning.retrieveByOrigin, therefore removed externalOAuthProviderProvisioning and tried to be more generic with refactoring from https://github.com/cloudfoundry/uaa/pull/3165

adrianhoelzl-sap:
closing in favor of #3165

PR #3165, opened 2024-11-28
Refactor password grant

strehle:
* if origin is available, retrieve IdP and check if password can be used
* use always identityProviderProvisioning and remove externalOAuthProviderConfigurator


server/src/main/java/org/cloudfoundry/identity/uaa/authentication/manager/PasswordGrantAuthenticationManager.java
```diff
@@ -145,8 +143,11 @@ private UaaLoginHint getUaaLoginHintForChainedAuth(List<String> allowedProviders
         return loginHintToUse;
     }
 
-    Authentication oidcPasswordGrant(Authentication authentication, final IdentityProvider<OIDCIdentityProviderDefinition> identityProvider) {
-        final OIDCIdentityProviderDefinition config = identityProvider.getConfig();
+    Authentication oidcPasswordGrant(Authentication authentication, final IdentityProvider<?> identityProvider) {
+        final OIDCIdentityProviderDefinition config = Stream.of(identityProvider).
```
adrianhoelzl-sap: I would prefer doing this cast/check already at the caller's site (should be line 108). The method should not allow passing anything other than an IdP with an OIDC configuration, therefore I like the old signature more.

server/src/main/java/org/cloudfoundry/identity/uaa/authentication/manager/PasswordGrantAuthenticationManager.java
```diff
@@ -109,12 +107,12 @@ public Authentication authenticate(Authentication authentication) throws Authent
         }
     }
 
-    private IdentityProvider<OIDCIdentityProviderDefinition> retrieveOidcPasswordIdp(UaaLoginHint loginHint, String defaultOrigin, List<String> allowedProviders) {
-        IdentityProvider<OIDCIdentityProviderDefinition> idp = null;
+    private IdentityProvider<?> retrievePasswordIdp(UaaLoginHint loginHint, String defaultOrigin, List<String> allowedProviders) {
+        IdentityProvider<?> idp = null;
```
adrianhoelzl-sap: Can we get rid of this variable and directly return either `null` or `retrievedByOrigin`?

adrianhoelzl-sap (approved):

mikeroda:
@strehle I believe this caused a regression with the password grant. If the IdP config only has the discoveryUrl set, the identityProviderProvisioning lookup does not fill in the tokenUrl like the externalOAuthProviderConfigurator did, and this results in a NullPointerException here:
https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/manager/PasswordGrantAuthenticationManager.java#L231

strehle:
@mikeroda ok, we can check ( FYI @adrianhoelzl-sap ) because this PR was created as optimization for IDP look.

Can you either create an issue ticket with information about type of IdP (oauth or oidc) and then simply discovery only and you do password grant correct ?

mikeroda:
@strehle, OK I created #3271.

PR #3172, opened 2024-11-29
Support SAML keypair without a passphrase

strehle:
found during testings with various IdPs... we need a dummy entry, but this should not be the case


duanemay (approved):
good catch

PR #3173, opened 2024-11-29
Add logs in case of SAML issues

strehle:
These issues should be ignored, because they were ignored before. Issue type:
1. Invalid IdP setup, e.g. invalid metadata. Write an error, but start the UAA
2. Invalid resolution of SAMl metadata during runtime. Add warning for such problems.


duanemay (approved):
LGTM

PR #3174, opened 2024-12-02
Code cleanup

duanemay:
Mostly automated with Openrewrite rules, and some manual updates.

Rules Run: CodeStyle, Common Static Analysis Issue remediation, Java 17


model/src/main/java/org/cloudfoundry/identity/uaa/codestore/ExpiringCode.java
```diff
@@ -76,28 +76,30 @@ public void setIntent(String intent) {
...
+        if (!Objects.equals(code, that.code)) {
             return false;
-        if (!Objects.equals(expiresAt, that.expiresAt))
+        }
+        if (!Objects.equals(data, that.data)) {
             return false;
-
-        return true;
+        }
+        return !!Objects.equals(expiresAt, that.expiresAt);
```
strehle: indendet to have !! ?

strehle (commented):
can you please check these 7 findings about
!!

https://sonarcloud.io/project/issues?severities=BLOCKER%2CCRITICAL%2CMAJOR%2CMINOR&sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&types=BUG&pullRequest=3174&id=cloudfoundry-identity-parent

model/src/main/java/org/cloudfoundry/identity/uaa/codestore/ExpiringCode.java
duanemay: Part of using automated tools... Fixed

model/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/idp/SamlServiceProvider.java
```diff
@@ -149,58 +149,71 @@ public boolean configIsValid() {
...
                 return false;
-        } else if (!entityId.equals(other.entityId))
-            return false;
-        if (version != other.version)
+            }
+        } else if (!entityId.equals(other.entityId)) {
             return false;
-        return true;
+        }
+        return !(version != other.version);
```
strehle: See here logic from your tool or was there before, but can be simplified

strehle (commented):
Please resolve with develop

model/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/idp/SamlServiceProvider.java
duanemay: previous was
``` 
  if (version != other.version)
            return false;
        return true;
``` 

I rebased develop, and added a few more fixes, now:
```
        return version == other.version;
```

strehle:
@duanemay So far I am through with, thanks for the cleanup. I have 2 open clarifications

1. The test coverage is low because some objects from IdP removal ( see https://github.com/cloudfoundry/uaa/pull/2638 , reason of major upgrade to 77.x ) still there. Suggest to remove these objects with an extra PR, WDYT ?

2. Guava. Should we remove Guava from usage at all ? Because right now you removed the direct dependency only but then we get Guava from the transient dependency from opensaml 4.x . We can exclude it in general but then also some other classes still use Util functions from Guava.  Problem now: transient dependency adds Guava 28.x which has some CVEs. With that package we run into a problem with our validation

strehle:
> @duanemay So far I am through with, thanks for the cleanup. I have 2 open clarifications
> 
> 1. The test coverage is low because some objects from IdP removal ( see [Remove SAML IDP feature #2638](https://github.com/cloudfoundry/uaa/pull/2638) , reason of major upgrade to 77.x ) still there. Suggest to remove these objects with an extra PR, WDYT ?

Tested without guava and yes, 
`net.shibboleth.utilities.java.support.xml.BasicParserPool.setBuilderFeatures`
needs it, so we cannot omit but then it is better we define the version of used guava and use the lastest. Otherwise we get an outdated version if we rely on transitive deps.

strehle (approved):

mikeroda:
@duanemay, can you give the specific rewrite recipes that were used for this code cleanup?

PR #3178, opened 2024-12-04
Fix retrieval of login.saml.providers

strehle:
The map can be incomplete if a dot is in key

e.g.
https://stackoverflow.com/questions/44014910/preventing-spring-boot-from-creating-nested-map-from-dot-separated-key-in-applic


server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/SamlConfigProps.java
```diff
@@ -106,18 +116,21 @@ public class SamlConfigProps {
 
     /**
      * Deprecated but sill working: login.serviceProviderKey
+     * WILL be REMOVED with UAA 78.x
```
duanemay: Should these be annotated @Deprecated(since = "77.20.0", forRemoval = true)

server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/SamlConfigProps.java
```diff
@@ -130,4 +143,23 @@ public class SamlConfigProps {
     public SamlKey getActiveSamlKey() {
         return keys != null ? keys.get(activeKeyId) : null;
     }
+
+    public Map<String, Map<String, Object>> getEnvironmentProviders() {
```
duanemay: Class is marked with @Data, so this is auto generated, and can be removed

server/src/test/java/org/cloudfoundry/identity/uaa/provider/saml/SamlConfigurationTest.java
```diff
@@ -0,0 +1,73 @@
...
+                        .getPropertySources()
+                        .get("servletConfigYaml")
+                        .getProperty("login")
+        )
+                .thenReturn(Map.of("saml", Map.of("providers", Map.of("idp.1", Map.of("idpMetadata", "<xml/>")))));
+    }
+
+    @Test
+    void bindFromEnvironment() {
+        //@formatter:off
```
duanemay: At least with Idea, text blocks do not get reformatted

strehle:
Sonar clean, now ready for review

adrianhoelzl-sap (approved):

duanemay (approved):

PR #3186, opened 2024-12-10
Fix DB unit tests

Kehrlann:
## Context

@fhanik and I have noticed that some unit tests did not respect the profile that was passed in, and always ran against an HSQL database, or were ignored

When you ran:

```
./gradlew '-Dspring.profiles.active=postgresql,default' test
```

without a postgres database running, the tests would pass, which should not be the case when a database is involved.

This PR only addresses tests, by fixing this issue, fixing failing tests and adding tooling.

### Consequences of the PR

- Higher confidence when making db-related changes
- Running against `postgresql` or `mysql` is now twice as slow. On my m1 mac, it goes from "just under 5 minutes" to "in the 10-12 minute range"
- The tests now run against the actual DB and you may have a problem with database names - they are `uaa_1`, `uaa_2`, etc ; all the way up to 24. Sometimes your tests may complain that `uaa_33` or something similar does not exist. In order to avoid most problems, please run your gradle tests with `--no-daemon`. If you want more details, head over to `docs/testing.md`.

### Fixes

The tests in question:
- Any test with `@WithDatabaseContext` ; because that annotation had `@ActiveProfiles("default")`. We removed it.
- `TableAndColumnNormalizationTest`, again annotated with `@ActiveProfiles("default")`
- And possibly some, but not all, the tests annotated with `@DefaultTestContext`

When we re-enabled the multi-profile support, we had issues in some tests, both with Postgres and MySQL. We have fixed those issues. The main issues were:

- The column name `groups` was improperly quoted in MySQL
- Database connection leaks, both in prod and in the tests
- Some tests cannot work on certain databases
- Some tests have casing and sorting issues
- There are timezone issues specifically with MySQL when the host running your tests is ahead of UTC (e.g. `Europe/Paris`) ; this is addressed by coercing the MySQL container to the same TZ as your host.

### Addition: `scripts/docker-compose.yml`

We added a docker compose file to have MySQL and Postgres running locally. Using `run-unit-tests.sh` was not working well on our ARM-based macs. This is not expected to be the final setup, but a first step to help us run tests fast.

We will converge startup scripts later on.

### Addition: `@EnabledWithProfile` / `@DisabledWithProfile`

To selectively turn off some tests based on profiles ; instead of relying on custom use of `Assume.assumeXXX` and `Assumptions.assumeXXX` which are harder to locate in the code.

See the javadoc for their usage.

### Addition: `docs/testing.md`

Some findings on testing have been recorded in `docs/testing.md`.


scripts/docker-compose.yml
```diff
@@ -0,0 +1,40 @@
+name: uaa
+
+services:
+  postgres:
+    image: "postgres:17.2"
```
strehle: please check if https://github.com/cloudfoundry/uaa/blob/develop/scripts/start_db_helper.sh can be re-used

fhanik: PR still in Work in Progress mode

We wrote this on MacOS, I see failures on Linux. Working on those today

fhanik:
***New findings***

1. Tests have database connection leaks. Many tests use `JdbcTemplate.getDataSource().getConnection()` and never call `connection.close()`

2. The  `@WithTimeZone` cause some tests to fail. Need to be investigated. For now, this annotation is limited to `mysql`

3. Columns have mismatched data types causing failure in later versions of MySQL (`users.id` is a padded `char` while `authz_approvals.user_id` is a varchar. 

4. Because tests didn't run against MySQL or PostgreSQL, we found many places where the word `groups` was used in queries and cause BadSqlGrammarException

5. All the MockMvc tests also had hardcoded to use in memory HSQLDB. This was fixed.

6. MySQL has in later versions deprecated `PAD_CHAR_TO_FULL_LENGTH` so we have converted `users.id` and `oauth_client_details.created_by` to VARCHAR

***Items Left to Do***

1. We can add maildev to docker-compose.yml (not to conflict with the in memory SMTP server `SimpleSmtpServer` that the integration tests run
2. We should enhance the LDAP image in docker-compose.yml to properly configure and seed the LDAP server - see `scripts/integration-tests.sh`
3. RateLimiter for `cargo` is causing 429 error on my Macbook M4
4. /etc/hosts requires the following entrie

```
127.0.0.1 testzone1.localhost
127.0.0.1 testzone2.localhost
127.0.0.1 testzone3.localhost
127.0.0.1 testzone4.localhost
127.0.0.1 testzonedoesnotexist.localhost
127.0.0.1 oidcloginit.localhost
127.0.0.1 testzoneinactive.localhost
```

5. Verify that `hsqldb` isn't unintentionally used in any tests

7. Fix remaining integration tests - current status has IT failures (assuming rate limits have been adjusted) - and one of them relates to LDAP preloading data (bullet 2 above) `LdapIntegrationTests.test_LDAP_Custom_User_Attributes_In_ID_Token`
8. is CHAR to VARCHAR a breaking change?



***Notes***
```
./gradlew \
  -Dspring.profiles.active=mysql,default \
  :cloudfoundry-identity-uaa:test \
  --tests org.cloudfoundry.identity.uaa.login.LoginMockMvcTests
```
This previously passed even if I shut down MySQL database. proving that the DB was not tested.

Kehrlann:
Update:

> The @WithTimeZone cause some tests to fail. Need to be investigated. For now, this annotation is limited to mysql

This was fixed with a new composed annotation, `@DatabaseTestUTC`

> Verify that hsqldb isn't unintentionally used in any tests

Verified that it's not used in unit tests, by removing the `hsqldb` dependency entirely.

Kehrlann:
Side note, there was a flake earlier on:

`DeprecatedUaaTokenServicesTests#testCreateAccessTokenRefreshGrantSomeScopesAutoApproved`, for both variants:
- https://github.com/cloudfoundry/uaa/actions/runs/12318536514/job/34383499756?pr=3186#step:5:8381
- https://github.com/cloudfoundry/uaa/actions/runs/12318536514/job/34383499756?pr=3186#step:5:10149

Errors were:
```
    java.lang.AssertionError: 
    Expected: should be valid for <is <43200>>
         but:  but was <43199>
        at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
        at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
        at org.cloudfoundry.identity.uaa.oauth.DeprecatedUaaTokenServicesTests.testCreateAccessTokenRefreshGrantSomeScopesAutoApproved(DeprecatedUaaTokenServicesTests.java:925)
```

And:
```
    java.lang.AssertionError: 
    Expected: Refresh token should be valid for null
         but:  was <2591999>
        at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
        at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
        at org.cloudfoundry.identity.uaa.oauth.DeprecatedUaaTokenServicesTests.testCreateAccessTokenRefreshGrantSomeScopesAutoApproved(DeprecatedUaaTokenServicesTests.java:930)
```

server/src/main/resources/org/cloudfoundry/identity/uaa/db/mysql/V4_112__Users_id_to_Varchar.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE users MODIFY id VARCHAR(36) NOT NULL;
```
strehle, line 1: @fhanik can you help @Kehrlann to create dummy migrations, e.g. https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/resources/org/cloudfoundry/identity/uaa/db/postgresql/V4_99_1561608282__NOOP.sql for postgresql and hsql to have same last migration number available

docs/testing.md
```diff
@@ -0,0 +1,91 @@
...
+
+    $ docker compose --file scripts/docker-compose.yaml up
+
+If you wish to launch only one of the DBs, select the appropriate service name:
+
+    $ docker compose --file scripts/docker-compose.yaml up postgresql
+
+To run tests against either Postgres or MySQL, use the `postgresql` or `mysql` profile, to select the DB. Be sure
+to add the `default` profile which will trigger seeding the database with some admin users, clients, etc. For example:
+
```
strehle, line 40: in main doc we state to test with 
https://github.com/cloudfoundry/uaa/blob/develop/run-integration-tests.sh
and
https://github.com/cloudfoundry/uaa/blob/develop/run-unit-tests.sh

docs/testing.md
```diff
@@ -0,0 +1,91 @@
...
+## Timezone issues
+
+The UAA and its DB server _MUST_ have the same timezone, because dates are not uniformly stored in UTC and timezones
+do matter. Specifically for MySQL, there are issues when your local host is ahead of UTC, because:
+
+1. The default containers runs in UTC
+2. When calling `current_timestamp` the value is in UTC
+3. But when calling a prepared statement from JDBC with a Date/Timestamp/time-based the timezone is sent to the server
+
+So, when running e.g. in France (CET):
```
strehle: So, when running e.g. in Europe (CET):

docs/testing.md
```diff
@@ -0,0 +1,91 @@
+# Testing considerations
```
strehle, line 1: if you want have an own testing for documentation, can you please add a section for developer tests about
- howto enable annotations to ensure to run on a certain DB or howto disable this.

server/src/main/java/org/cloudfoundry/identity/uaa/db/beans/FlywayConfiguration.java
```diff
@@ -63,6 +63,9 @@ public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata)
         public Flyway flyway(Flyway baseFlyway) {
             baseFlyway.repair();
             baseFlyway.migrate();
+            org.apache.tomcat.jdbc.pool.DataSource ds =
```
strehle, line 66: do you see the leakage for all UAA starts or only in case a migration takes place / is needed ?

Kehrlann: Only in migrations so far ; both here and in [mysql/V4_9_2__AddPrimaryKeysIfMissing.java](https://github.com/cloudfoundry/uaa/pull/3186/commits/1d58d2a50157c8675595e5b82494d419d2cf6254#diff-fec9a569831747f5830030ab9802a3ee8e65b07cdd0a8ae8a7ed5fa808d1b19c)

strehle:
maybe was not good enough documented, but if I had to use a DB in past I setup the docker for it via

`docker pull cfidentity/uaa-mysql-8 && pushd ~/work/uaa && docker run -p 3306:3306 --name db --rm --env MYSQL_ROOT_HOST=% --env DB=mysql --env RUN_TESTS=false --volume $PWD:/root/uaa --privileged --tty --interactive --shm-size=1G cfidentity/uaa-mysql-8 /root/uaa/scripts/integration-tests.sh mysql,default`

because this uses the same images which run in all of our tests

I guess for your Mac/Processor this is not possible... thus a plain Linux is always the best choice for development

server/src/main/resources/org/cloudfoundry/identity/uaa/db/mysql/V4_112__Users_id_to_Varchar.sql
Kehrlann: Done, thanks!

docs/testing.md
Kehrlann: Amended to reflect this

Kehrlann:
Thanks for the trick running the tests.

It _is_ possible to run on my Mac, it's just slower than a plain x86 proc.

The reason for the docker-compose is mostly to be able to debug DB tests inside my IDE :)

strehle:
> Thanks for the trick running the tests.
> 
> It _is_ possible to run on my Mac, it's just slower than a plain x86 proc.
> 
> The reason for the docker-compose is mostly to be able to debug DB tests inside my IDE :)

yeah the docker-compose helps to better understand the setup. for ldap we need similar thing. https://github.com/cloudfoundry/uaa?tab=readme-ov-file#connecting-uaa-to-local-ldap-server

I only start DBs, when we have new columns with an index on it, then we check with explain and SQL if the index is really used or not...  for all other cases... only a new column and no index I dont need the real DB

docs/testing.md
```diff
@@ -0,0 +1,136 @@
...
+## Timezone issues
+
+The UAA and its DB server _MUST_ have the same timezone, because dates are not uniformly stored in UTC and timezones
+do matter. Specifically for MySQL, there are issues when your local host is ahead of UTC, because:
+
+1. The default containers runs in UTC
+2. When calling `current_timestamp` the value is in UTC
+3. But when calling a prepared statement from JDBC with a Date/Timestamp/time-based the timezone is sent to the server
+
+So, when running e.g. in `Europe/Paris` (CET):
```
strehle, line 118: :-) its ok, but what about Europe/Berlin or Europe/Madrid ? ... I try to omit such country specifics...but CET is good enough to call it Europe

strehle (approved):
Ok for me

server/src/main/java/org/cloudfoundry/identity/uaa/db/beans/FlywayConfiguration.java
fhanik: @strehle So I'm not convinced it is a leak as much as "parallelism" causes too many idle connections and overwhelms a developer postgresql instance (max=100). When I'm investigating, I don't see an actual leak in this method.

the `purge` simply instructs the pool to disconnect existing connections that aren't being used, and let the DB connection pool build up again. That way the test suite stays under the limit of max connections (which were mostly idle connections)

strehle:
> debug DB tests inside my IDE :)

if I started a DB on my machine and map ports , e.g. -p 3306:3306

I can debug then everything and can access my tables with a GUI like DBeaver, but docker-compose is ok

duanemay (approved):
👍

server/src/test/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioningTests.java
```diff
@@ -443,7 +447,11 @@ void retrieveByScimFilter_IncludeInactive() {
             );
             Assertions.assertThat(result).isNotNull();
             final List<String> usernames = result.stream().map(ScimUser::getUserName).toList();
-            Assertions.assertThat(usernames).isSorted();
```
coolgang123, line 444: Isn't it a difference of functionality with different databases. Is it something that needs to be fixed?

coolgang123 (approved):

docs/testing.md
Kehrlann: Yes, but also:
- `Europe/Lisbon` (Western European Time, WEST) is UTC / UTC+1
- `Europe/Bucharest` (Eastern European Time, EEST) is UTC+2 / UTC+3

And if you truly want to be nitpicky:
-  There's even a `Europe/Moscow` (+3/+5) and `Europe/Samara` (+4) 😱 

Promise I'll do `Europe/Berlin` next time!

server/src/test/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioningTests.java
Kehrlann: I looked into it. The test calls:

```java
jdbcScimUserProvisioning.query(
                    ...,
                    "userName",
                    ...,
                    ...
            )
```

So it sorts by username alphabetically, whatever "alphabetical" means for the given database. In production code it's only ever sorted by `created`.

I updated the test to reflect this.

fhanik: hi @coolgang123 

Yes, the databases indeed are different, but it also depends on locale for the sorting. So the test itself is not all that meaningful, because the sort order can change when the database configuration changes.

If we wanted to test all DBs we could use the following fix, but I would still vote to remove the check for MySQL 

```
diff --git a/server/src/test/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioningTests.java b/server/src/test/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioningTests.java
index 864f1af2c..9c1d0b7b6 100644
--- a/server/src/test/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioningTests.java
+++ b/server/src/test/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioningTests.java
@@ -448,9 +448,11 @@ class JdbcScimUserProvisioningTests {
             Assertions.assertThat(result).isNotNull();
             final List<String> usernames = result.stream().map(ScimUser::getUserName).toList();
             if (Arrays.stream(enviroment.getActiveProfiles()).noneMatch("mysql"::equalsIgnoreCase)) {
-                // MySQL has a different ordering from HSQL and Postgres. In MySQL the two values
-                // we inserted are sorted in reverse order, so we skip the assert entirely
                 Assertions.assertThat(usernames).isSorted();
+            } else {
+                //MySQL has a different ordering from HSQL and Postgres. In MySQL the two values
+                //we inserted are sorted in reverse order,
+                Assertions.assertThat(usernames).isSortedAccordingTo((o1, o2) -> o2.compareTo(o1));
             }
             return usernames;
         };
```

coolgang123: sounds good

docs/testing.md
strehle: ok, maybe I explained my intension for "So, when running e.g. in CET" not good enough.

CET is enough, we dont need further information like the country or city.
You use UTC also without context like Earth (UTC)

Kehrlann: Ah yes makes sense, I get it. My bad.

To clarify, when I say "[e.g.](https://simple.wikipedia.org/wiki/Exempli_gratia) in" it means "for example in"

strehle: I understood for example, but TMI

PR #3187, opened 2024-12-11
Update start_db_helper.sh

strehle:


Kehrlann (changes requested):
I think there are two issues with this:

**A.** If you run `./gradlew test`, several gradle modules may be tested in parallel, and so there WILL be multiple workers.
**B.** As the comment above states, "We make extra dbs because a gradle worker ID can exceed the max number of workers." In the test I just ran, I got workers 7 and 8

This change is only acceptable if:
**1.** You run without a gradle daemon, e.g. in a brand new container, in CI, with `--no-daemon`, or after `./gradlew --stop`
**2.** AND you run tests for a single project, e.g. `./gradlew cloudfoundry-identity-server:test`

So this would hinder our ability to run tests locally because of we want to use gradle daemons locally (so, not **1.**); and would very likely break CI whenever we merge #3186 because in CI we just run "test" on all projects (not **2**) .

I think this will require changes in how the tests run in the first place ; either make projects depend on each other, or changes to how databases are provisioned.

Edit:

It is actually already failing in CI, probably because of **A** / not **2**.

I will think about it over the week-end, but I _think_ the correct move is to have to have three databases:

1. `uaa` (which is there by default) for connecting a local UAA
2. `cloudfoundry-identity-server-test` for tests 
3. `cloudfoundry-identity-server-uaa` for tests

@strehle happy to take care of this if you want.

strehle:
> I think there are two issues with this:
> 
> **A.** If you run `./gradlew test`, several gradle modules may be tested in parallel, and so there WILL be multiple workers. **B.** As the comment above states, "We make extra dbs because a gradle worker ID can exceed the max number of workers." In the test I just ran, I got workers 7 and 8
> 
> This change is only acceptable if: **1.** You run without a gradle daemon, e.g. in a brand new container, in CI, with `--no-daemon`, or after `./gradlew --stop` **2.** AND you run tests for a single project, e.g. `./gradlew cloudfoundry-identity-server:test`
> 
> So this would hinder our ability to run tests locally because of we want to use gradle daemons locally (so, not **1.**); and would very likely break CI whenever we merge #3186 because in CI we just run "test" on all projects (not **2**) .
> 
> I think this will require changes in how the tests run in the first place ; either make projects depend on each other, or changes to how databases are provisioned.
> 
> Edit:
> 
> It is actually already failing in CI, probably because of **A** / not **2**.
> 
> I will think about it over the week-end, but I _think_ the correct move is to have to have three databases:
> 
> 1. `uaa` (which is there by default) for connecting a local UAA
> 2. `cloudfoundry-identity-server-test` for tests
> 3. `cloudfoundry-identity-server-uaa` for tests
> 
> @strehle happy to take care of this if you want.

@Kehrlann Thanks for your effort. It was not my intension to do a change here. Did it to see if multiple DBs are really needed

PR #3195, opened 2024-12-13
Testing Updates

duanemay:
This PR attempts to provide more consistency in our test suites, standardizing on Junit5 and AssetJ. 

- Migrate to Junit 5
- Migrate to AssertJ
- Migrate Hamcrest Assertions
- Use SpringJUnitConfig

We have had a mix of Junit 4 and 5 annotations, and a mix of Junit, Hamcrest, and AssertJ asserts, and various hamcrest matchers.
The Junit Annotations have all been moved to v5. Junit Rules have been migrated to Extensions. There remain places where Junit 4 support code is used notably in Extensions, e.g. OAuth2ContextExtension.
Hamcrest Matchers are still used in the production code and a few places in the tests.


server/src/test/java/org/cloudfoundry/identity/uaa/ServerRunningExtension.java
```diff
@@ -50,44 +47,42 @@
...
+ *      ResponseEntity<Void> response = serverRunning.postForResponse(serverRunning.getAuthorizationUri(), headers, params);
+ * }
  * </pre>
+ *
  * <p>
- * The rule can be declared as static so that it only has to check once for all
- * tests in the enclosing test case, but there isn't a lot of overhead in making
- * it non-static.
+ * The Extension can be declared as static so that it only has to check once for all
+ * tests in the enclosing test case.
```
Kehrlann: **[question]** I'm not sure this comment makes sense anymore, maybe we should remove it? As I understand it, the check is performed in the `BeforeAll` callback and will run once in any case, whether your register the extension as a static `@RegisterExtension` field or with `@ExtendWith`.

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/util/ScreenshotOnFailExtension.java
```diff
@@ -0,0 +1,73 @@
...
+
+    private File getDestinationFile(String className, String description, String extension) {
+        String home = System.getProperty("user.home");
+        String absoluteFileName = "%s/build/cloudfoundry/uaa/uaa/build/reports/tests/%s/%s.%s".formatted(home, className, description, extension);
+        return new File(absoluteFileName);
+    }
+
+    public void setWebDriver(WebDriver webDriver) {
+        this.driver = webDriver;
+    }
```
Kehrlann: **[suggestion, non-blocking]** This is an extension, and the web driver is a bean, it should be possible to fetch the webdriver it from the `ExtensionContext` without programatically setting it:

```java
SpringExtension.getApplicationContext(extensionContext).getBean(WebDriver.class);
```

Kehrlann (approved):
I have skimmed some of the changes at random, nothing jumped out.

I have taken a closer look at the extensions ; they are still used as "old-style" JUnit rules, instead of newer `@ExtendWith`. I think it's fine, might be worth changing in the future if we do more work on the integration tests themselves. Of all the extensions, the `ScreenshotOnFailExtension` one can be more easily changed to be `@ExtendWith(ScreenshotOnFailExtension.class)`.

There is a comment that seems off in the `ServerRunningExtension` though.

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/util/ScreenshotOnFailExtension.java
duanemay: I'll make a smaller follow up later, this to take care of these items

Kehrlann: If you think it's valuable sure, but feel free to also not do it, it's very minor.

strehle (approved):

PR #3198, opened 2024-12-16
Add legacy SAML configuration for uaa-ci

strehle:
Should replace
https://github.com/cloudfoundry/uaa-ci/blob/main/concourse/uaa-acceptance-gcp/tasks/push-oidc10/manifest-oidc10.yml#L18

This example can be used to show SAML setup but should serve for CI

### DIFF
--- a/scripts/saml/saml-legacy-uaa.yml
+++ b/scripts/saml/saml-legacy-uaa.yml
@@ -1,0 +1,139 @@
+# SAML Configuration with legacy key configuration
+# Added 2 Example IdPs
+login:
+  saml:
+    providers:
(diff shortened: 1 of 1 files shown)


strehle:
@duanemay can you check if Ok for you ?

This PR should 
a) ensure that we have SAML setup with dots in origin and invalid metadata, but it should start
b) remove dependency in our CI , means move away from https://raw.githubusercontent.com/cf-identity/test-oidc-1.0/master/saml-providers.yml towards a yaml inside uaa github

duanemay (approved):
looks fine

PR #3203, opened 2024-12-18
tests: enable parallel testing to speed up database-backed tests

Kehrlann:
## Context

Now that we have re-enabled DB testing with gh-3186, Postgres and MySQL tests are slower, and it is useful to enable some level of parallelism.

Please pay attention to `docs/testing.md`, to ensure the explanations are clear.

### Results

On my mac M1, this speeds up the tests, from around 13 minutes to under 5 minutes for MySQL.

HSQL-based unit tests are only marginally faster.

### Blockers

Blocked by gh-3202 ; as some LDAP tests fail when running in parallel.


server/src/main/java/org/cloudfoundry/identity/uaa/db/UaaDatabaseName.java
```diff
@@ -14,6 +14,7 @@ public String getName() {
             return UAA_DB_NAME;
         }
 
+        System.out.println("⚙️ daniel --- " + gradleWorkerId);
```
mamachanko: **[question]:** is this sawdust?

Kehrlann: 💯 

thank you!

duanemay (approved):
👍

strehle (approved):

PR #3208, opened 2024-12-20
Move `login-ui` Service classes to Javaconfig

Kehrlann:
## Context
  
  We are moving out of XML config and into java config.
  
  This moves `*Service` used by `login-ui` out of XML config, and can be used as a blueprint to move some config out of XML.
  
  Below are a few notable changes to the configuration.
  
  ### Configuration properties
  
  Example: `SmtpConfigurationProperties`.
  
Avoid `@Value`, prefer `@ConfigurationProperties` classes which allow for easily following where they are actually used. Default values can be provided. Validation can be applied.  Avoid SpEL expressions where possible, prefer Java code, either in constructors or in bean methodes.

For configuration properties, when possible, prefer immutable `record`s.
  
  ### Code-independent beans
  
  Examples: `EmailChangeEmailService`, `EmailAccountCreationService`, etc.
  
  Any bean that:
  1. Has a unique implementation
  1. Does not have conditions to decide whether it should be present or not
  1. Does not have configuration variants ("use this or that param, based on some  condition")
  
  can be provided with a stereotype annotation, such as `@Component` or `@Service`. When they are not used anywhere in XML config, they do not need to be named.
  
  ### Conditional and code-dependent beans
  
  Examples: `JavaMailSender`, `MessageService`.
  
  Some beans are configuration-dependent. They require some code to be correctly created, and may be conditional on some application properties ("only create a `NotificationsService` when there is a `notifications.url` property"). They are typically interfaces, that have multiple implementations.
  
  Those beans need to be provided in `@Bean` functions in `@Configuration` classes.
  
  Typically, one implementation is the default, and other is what happens when certain conditions are met. A way to achieve this in javaconfig is to annotation the configuration-dependent bean with `@Primary`, and it will override the default bean. Spring Framework 6.2 / Boot 3.4 will provide an [@Fallback](https://docs.spring.io/spring-framework/reference/core/beans/annotation-config/autowired-primary.html#page-title) annotation to achieve the same result.

Keep the code minimum in `@Bean` methods when possible, and rely on declarative configuration instead.
  
  ### Reference data
  
  With beans moved out of XML, reference data, such as `util:map id="notifications"`, needs to be moved to java (in this case, in a helper static method, `notifications()`). 

In this particular case we chose functions to avoid a nested `Map<String, Map<String, String>>`: it makes in XML or other text-based data representation, but less in Java.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/account/EmailAccountCreationService.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/account/EmailAccountCreationService.java
@@ -1,38 +1,45 @@
 package org.cloudfoundry.identity.uaa.account;
 
 import com.fasterxml.jackson.core.type.TypeReference;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.cloudfoundry.identity.uaa.codestore.ExpiringCode;
 import org.cloudfoundry.identity.uaa.codestore.ExpiringCodeStore;
 import org.cloudfoundry.identity.uaa.constants.OriginKeys;
 import org.cloudfoundry.identity.uaa.error.UaaException;
 import org.cloudfoundry.identity.uaa.message.MessageService;
 import org.cloudfoundry.identity.uaa.message.MessageType;
+import org.cloudfoundry.identity.uaa.oauth.provider.ClientDetails;
+import org.cloudfoundry.identity.uaa.provider.NoSuchClientException;
 import org.cloudfoundry.identity.uaa.scim.ScimUser;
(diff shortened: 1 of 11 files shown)


coolgang123 (approved):

strehle (approved):

strehle:
@Kehrlann I would merge this. You want stay with all commits, correct ?

Only ask because I dont see a value in 4 commits with 3 having same info
image

But you want all, ?

Kehrlann:
@strehle yes we'd like to keep the commits. The first one is more involved that the last three, but those are valid, self-described commits, so we prefer keeping them.

PR #3209, opened 2024-12-20
tests: address flakyness in ScimUserEndpointsAliasMockMvcTests

Kehrlann:
Trying to address test flakyness in `ScimUserEndpointsAliasMockMvcTests` as raised by @fhanik and sometimes seen in Github actions.

Unsure whether that fixes everything but it seems to stabilize some tests on my machine.

### DIFF
--- a/uaa/src/test/java/org/cloudfoundry/identity/uaa/scim/endpoints/ScimUserEndpointsAliasMockMvcTests.java
+++ b/uaa/src/test/java/org/cloudfoundry/identity/uaa/scim/endpoints/ScimUserEndpointsAliasMockMvcTests.java
@@ -1,7 +1,6 @@
 package org.cloudfoundry.identity.uaa.scim.endpoints;
 
 import com.fasterxml.jackson.core.type.TypeReference;
-
 import org.cloudfoundry.identity.uaa.DefaultTestContext;
 import org.cloudfoundry.identity.uaa.alias.AliasMockMvcTestBase;
 import org.cloudfoundry.identity.uaa.mock.util.MockMvcUtils;
@@ -14,6 +13,7 @@ import org.cloudfoundry.identity.uaa.scim.ScimUser;
 import org.cloudfoundry.identity.uaa.scim.ScimUserAliasHandler;
 import org.cloudfoundry.identity.uaa.scim.jdbc.JdbcScimUserProvisioning;
 import org.cloudfoundry.identity.uaa.scim.services.ScimUserService;
+import org.cloudfoundry.identity.uaa.util.AlphanumericRandomValueStringGenerator;
 import org.cloudfoundry.identity.uaa.util.JsonUtils;
 import org.cloudfoundry.identity.uaa.util.UaaStringUtils;
 import org.cloudfoundry.identity.uaa.zone.IdentityZone;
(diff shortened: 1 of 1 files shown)


duanemay (approved):

uaa/src/test/java/org/cloudfoundry/identity/uaa/scim/endpoints/ScimUserEndpointsAliasMockMvcTests.java
```diff
@@ -1600,7 +1609,7 @@ private static ScimUser buildScimUser(
         scimUser.setAliasZid(aliasZid);
         scimUser.setZoneId(zoneId);
 
-        scimUser.setUserName("john.doe");
```
fhanik, line 1603: Maybe generate a username each test run, that way this issue can't come back ?

Kehrlann: Good idea, addressed 👍

uaa/src/test/java/org/cloudfoundry/identity/uaa/scim/endpoints/ScimUserEndpointsAliasMockMvcTests.java
```diff
@@ -69,6 +80,11 @@ void setUp() throws Exception {
         scimUserService = requireNonNull(webApplicationContext.getBean(ScimUserService.class));
     }
 
+    @AfterEach
+    void tearDown() {
+        jdbcTemplate.execute("DELETE FROM users WHERE username = '" + TEST_USERNAME + "'");
```
strehle, line 87: do we need the deletion if we use random user names ?

uaa/src/test/java/org/cloudfoundry/identity/uaa/scim/endpoints/ScimUserEndpointsAliasMockMvcTests.java
```diff
@@ -1600,7 +1615,7 @@ private static ScimUser buildScimUser(
         scimUser.setAliasZid(aliasZid);
         scimUser.setZoneId(zoneId);
 
-        scimUser.setUserName("john.doe");
+        scimUser.setUserName(TEST_USERNAME);
         scimUser.setName(new ScimUser.Name("John", "Doe"));
         scimUser.setPrimaryEmail("john.doe@example.com");
```
strehle: not 100% sure if email should be TEST_USERNAME + "@example.com" to stay on safe side

strehle (commented):
only minor comments

strehle (approved):

uaa/src/test/java/org/cloudfoundry/identity/uaa/scim/endpoints/ScimUserEndpointsAliasMockMvcTests.java
Kehrlann: Very good point. It is not required. 

Since there are many tests, it's useful when debugging to have less users with random names in the table. I'm going to add a comment.

uaa/src/test/java/org/cloudfoundry/identity/uaa/scim/endpoints/ScimUserEndpointsAliasMockMvcTests.java
Kehrlann: Good point, changing.

PR #3219, opened 2025-01-06
JWT federated client authentication

### LINKED ISSUE #2838: Allow creating a trust relationship with Github Actions OIDC provider to retrieve an uaa access token using a Github id_token

* `76.26.0`
* `77.1.0`

I am deploying the UAA

#### `JWT Bearer Token Grant` by adding the Github OIDC Provider:

I added the Github OIDC provider using non existing credentials and used the `repository_owner` claim as the `user_name`:

```
uaa curl /identity-providers -X POST -H "Content-Type: application/json" -d '{"type": "oidc1.0", "name": "Github", "originKey": "github", "config": {"discoveryUrl": "https://token.actions.githubusercontent.com/.well-known/openid-configuration", "scopes": ["read:user", "user:email"], "linkText": "Login with Github", "showLinkText": false, "addShadowUserOnLogin": true, "clientAuthInBody": true, "relyingPartyId": "uaa", "relyingPartySecret": "uaa", "addShadowUserOnLogin": true, "attributeMappings" : {"given_name": "repository_owner", "family_name": "repository_owner_id", "user_name": "repository_owner"}}}'
```

The `sub` can't be used for the `user_name`, as it includes unsupported characters like `/`  and `:`. I used the `repository_owner` claim for the  `user_name`.

UAA client required for authentication:
```
uaa curl /oauth/clients -X POST -H "Content-Type: application/json" -d '{"client_id" : "jwt-bearer-client", "client_secret" : "secret", "access_token_validity": 1800,  "authorities" : [ "uaa.resource" ], "authorized_grant_types" : [ "urn:ietf:params:oauth:grant-type:jwt-bearer" ], "scope": ["openid", "cloud_controller.read"], "allowedproviders" : [ "github" ], "name" : "JWT Bearer Client"}'
```

Finally the request to get an `access_token` (`uaa` has to be used for requesting the `id_token`)
```
curl -u 'jwt-bearer-client:secret' -d 'grant_type=urn:ietf:params:oauth:grant-type:jwt-bearer&assertion=<github id token>' https://uaa.sys.domain.com/oauth/token
```

The resulting `access_token`:
```
{
  "jti": "a6bfcbc1e851444196ce76779567cbed",
  "sub": "c8068c6a-8e21-408f-9532-0f3a413f988e",
  "scope": [
    "openid",
    "cloud_controller.read"
  ],
  "client_id": "jwt-bearer-client",
  "cid": "jwt-bearer-client",
  "azp": "jwt-bearer-client",
  "grant_type": "urn:ietf:params:oauth:grant-type:jwt-bearer",
  "user_id": "c8068c6a-8e21-408f-9532-0f3a413f988e",
  "origin": "github",
  "user_name": "vchrisb",
  "email": "vchrisb@user.from.github.cf",
  "rev_sig": "7ec0c4f2",
  "iat": 1713450185,
  "exp": 1713493385,
  "iss": "https://uaa.sys.domain.com/oauth/token",
  "zid": "uaa",
  "aud": [
    "cloud_controller",
    "openid",
    "jwt-bearer-client"
  ]
}
```

Example Github workflow to retrieve the `id_token` using a custom github action [vchrisb/setup-cf](https://github.com/vchrisb/setup-cf):
```
name: CI
permissions:
  id-token: write # This is required for requesting the JWT
  contents: read  # This is required for actions/checkout

on:
  push:

jobs:
  deploy:
    name: password
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v4
    - name: setup cf cli
      uses: vchrisb/setup-cf@v0.0.3
      with:
        api: ${{ secrets.CF_API }}
        client_id: ${{ secrets.CF_CLIENT_ID }}
        client_secret: ${{ secrets.CF_CLIENT_SECRET }}
        grant_type: jwt-bearer
        org: test
        space: dev
    - name: access cloud foundry api
      run: cf push --strategy rolling
```

It is possible to get an uaa ` access_token` using the `JWT Bearer Token Grant`, but this has _limitations_:

* adding the identity provider requires fake credentials
* the `sub`, or combination of claims, can't be used for the `user_name`
* a `client_secret` is required

#### `client_credentials Grant`:

I also tried using the `client_credentials Grant` with `private_key_jwt` to retrieve an UAA access_token by providing a Github `id_token`.
For that I created a UAA client, which name matched the `sub` claim of the `id_token`. I configured the client's jwks_uri with the one of Github. An authentication does fail, because the `iss` claim does not include the `client_id`, which is required by the [specs ](https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication)for this flow.

Conversation on that in [UAA Slack](https://cloudfoundry.slack.com/archives/C03FXANBV/p1712759392190589).

_I would like to be able to access the cloud foundry api from a Github workflow using temporary credentials._

#### Background

From [Github Security hardening with OpenID Connect](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#example-subject-claims):

> GitHub Actions workflows are often designed to access a cloud provider (such as AWS, Azure, GCP, or HashiCorp Vault) in order to deploy software or use the cloud's services. Before the workflow can access these resources, it will supply credentials, such as a password or token, to the cloud provider. These credentials are usually stored as a secret in GitHub, and the workflow presents this secret to the cloud provider every time it runs.
> 
> However, using hardcoded secrets requires you to create credentials in the cloud provider and then duplicate them in GitHub as a secret.
> 
> With OpenID Connect (OIDC), you can take a different approach by configuring your workflow to request a short-lived access token directly from the cloud provider. Your cloud provider also needs to support OIDC on their end, and you must configure a trust relationship that controls which workflows are able to request the access tokens. Providers that currently support OIDC include Amazon Web Services, Azure, Google Cloud Platform, and HashiCorp Vault, among others.

Github Actions does provide an `id_token` to workflows with following relevant claims from [Understanding the OIDC token](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#understanding-the-oidc-token):
* `sub`, e.g. `repo:octo-org/octo-repo:environment:prod`, the template used to generate the `sub` value can be [configured](https://docs.github.com/en/rest/actions/oidc?apiVersion=2022-11-28#set-the-customization-template-for-an-oidc-subject-claim-for-an-organization)
* `iss` is always `https://token.actions.githubusercontent.com`
* `aud` can be configured when requesting the `id_token` in the workflow

Example Job to retrieve the `id_token`:
```
  job:
    environment: Production
    runs-on: ubuntu-latest
    steps:
    - name: Install OIDC Client from Core Package
      run: npm install @actions/core@1.10.1 @actions/http-client
    - name: Get Id Token
      uses: actions/github-script@v7
      id: idtoken
      with:
        script: |
          const coredemo = require('@actions/core')
          let id_token = await coredemo.getIDToken('https://uaa.sys.domain.com/oauth/token')
          coredemo.setOutput('id_token', id_token)
```

The Github OIDC provider is solely exposing the JSON web key (JWK) endpoint: `https://token.actions.githubusercontent.com/.well-known/jwks` to be used to validate the Github `id_token`.

[Github OIDC openid-configuration](https://token.actions.githubusercontent.com/.well-known/openid-configuration)

The Github documentation does have examples how to create a trust relation with [AWS](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services), [Azure](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-azure), [GCP ](https://docs.github.com/en/enterprise-server@3.10/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-google-cloud-platform) and [Vault](https://docs.github.com/en/enterprise-server@3.10/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-hashicorp-vault).

(The Github actions OIDC provider is not the same as the Oauth2 provider for [Github Oauth2 Apps](https://docs.github.com/en/apps/oauth-apps).)

#### Ask

I want to use a short live token to interact with the cloud foundry api.
For that, I would like to be able to create a trust relation from UAA to Github, (and other CI systems like [GitLab](https://docs.gitlab.com/ee/ci/cloud_services/)) to use the `id_token` provided in the workflow to retrieve an UAA client `access_token`.
The UAA client might be created using the `sub` claim of the `id_token` or a pre created client might be pattern matched against the `sub`.
The client will be used to assign a cloud foundry space role to it.

strehle:
Support integration of so called workload identities

Allow client authentication based on RFC 7523, see https://oauth.net/private-key-jwt/

Scenario: Integrate JWT from another systems to authentication within UAA / CF landscape.
e.g. github action token like https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/about-security-hardening-with-openid-connect#understanding-the-oidc-token


model/src/main/java/org/cloudfoundry/identity/uaa/oauth/client/ClientJwtCredential.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public ClientJwtCredential(String subject, String issuer, String audience) {
+        this.subject = subject;
+        this.issuer = issuer;
+        this.audience = audience;
+    }
+
+    @JsonIgnore
+    public boolean isValid() {
+        return subject != null && issuer != null && !subject.isEmpty() && !issuer.isEmpty();
```
Kehrlann: **[question, blocking]** Should it even be possible to register an invalid configuration? If possible, I'd prefer the constructor (or a factory method) to reject invalid values, rather than accepting them and then failing much later on, when trying to obtain a token.

**[nitpick]**

```suggestion
        return StringUtils.hasText(subject) && StringUtils.hasText(issuer);
```

you can also use `hasLength` if you want to allow blanks.

model/src/test/java/org/cloudfoundry/identity/uaa/oauth/client/ClientJwtCredentialTest.java
```diff
@@ -0,0 +1,43 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+
+class ClientJwtCredentialTest {
+
+    @Test
+    void parse() {
+        assertDoesNotThrow(() -> ClientJwtCredential.parse("[{\"iss\":\"http://localhost:8080/uaa\",\"sub\":\"client_with_jwks_trust\"}]"));
+        List<ClientJwtCredential> federationList = ClientJwtCredential.parse("[{\"iss\":\"http://localhost:8080/uaa\",\"sub\":\"client_with_jwks_trust\"},{\"iss\":\"http://localhost:8080/uaa\"}]");
+        assertEquals(2, federationList.size());
```
Kehrlann: **[suggestion, blocking]** From what I understand we're aligning on AssertJ assertions in gh-3195.

Please use `assertThat(federationList).hasSize(2)` and other assertJ assertions in tests.

server/src/main/java/org/cloudfoundry/identity/uaa/authentication/ClientDetailsAuthenticationProvider.java
```diff
@@ -71,10 +71,10 @@ protected void additionalAuthenticationChecks(UserDetails userDetails, UsernameP
                         setAuthenticationMethod(authentication, CLIENT_AUTH_NONE);
                         break;
                     } else if (isPrivateKeyJwt(authentication.getDetails())) {
+                        setAuthenticationMethod(authentication, CLIENT_AUTH_PRIVATE_KEY_JWT);
```
Kehrlann, line 74: **[question, non-blocking]** Is the order important here? If so, why?

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/clients/ClientAdminEndpointDocs.java
```diff
@@ -80,10 +80,13 @@ class ClientAdminEndpointDocs extends AdminClientCreator {
...
+            fieldWithPath("jwks").constrained("Optional only, if jwks_uri is used. Required otherwise. OIDC standard.").type(STRING).description("A valid JSON string according JSON Web Key Set standard, see [RFC 7517](https://www.rfc-editor.org/rfc/rfc7517), e.g. content of /token_keys endpoint from UAA"),
+            fieldWithPath("jwks_uri").constrained("Optional only, if jwks is used. Required otherwise. OIDC standard.").type(STRING).description("A valid URI to token keys endpoint. Must be compliant to jwks_uri from [OpenID Discovery](https://openid.net/specs/openid-connect-discovery-1_0.html)."),
+            fieldWithPath("iss").constrained("Optional only, if OIDC standard is used. Required otherwise. OAuth2 standard.").type(STRING).description("OIDC complaint issuer URI, matching the retreived OAuth2 complaint JWT (iss claim)."),
```
Kehrlann: **[typo]**

```suggestion
            fieldWithPath("iss").constrained("Optional only, if OIDC standard is used. Required otherwise. OAuth2 standard.").type(STRING).description("OIDC compliant issuer URI, matching the retrieved OAuth2 compliant JWT (iss claim)."),
```

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/clients/ClientAdminEndpointDocs.java
```diff
@@ -80,10 +80,13 @@ class ClientAdminEndpointDocs extends AdminClientCreator {
...
+            fieldWithPath("sub").constrained("Optional only, if OIDC standard is used. Required otherwise. OAuth2 standard.").type(STRING).description("A subject string, matching the retreived OAuth2 complaint JWT (sub claim)."),
```
Kehrlann: **[typo]**

```suggestion
            fieldWithPath("sub").constrained("Optional only, if OIDC standard is used. Required otherwise. OAuth2 standard.").type(STRING).description("A subject string, matching the retrieved OAuth2 compliant JWT (sub claim)."),
```

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/clients/ClientAdminEndpointDocs.java
```diff
@@ -80,10 +80,13 @@ class ClientAdminEndpointDocs extends AdminClientCreator {
...
+            fieldWithPath("aud").constrained("Optional. Otherwise UAA token endpoint is expected. OAuth2 standard.").type(STRING).description("An audience for the retreived OAuth2 complaint JWT (part of aud claim array)."),
```
Kehrlann: **[typo]**
```suggestion
            fieldWithPath("aud").constrained("Optional. Otherwise UAA token endpoint is expected. OAuth2 standard.").type(STRING).description("An audience for the retreived OAuth2 compliant JWT (part of aud claim array)."),
```

uaa/slateCustomizations/source/index.html.md.erb
```diff
@@ -2216,10 +2216,29 @@ instead of secret based client authentication. See details for client authentica
...
+
+<aside class="success">
+  Added in 77.24.0
+</aside>
+
 
 The client can send a client_assertion for authentication instead of client_secret or Authorization header.
-The signature of the JWT from client_assertion is validated according to OpenID Connect standard either with the jwks_uri or the internal jwks public keys.
+The signature of the JWT from client_assertion is validated and assures the client authentication. The structure of the JWT
+sent with parameter client_assertion differentiates between the OAuth2 and OpenID Connect standard and ith this, both standards
```
Kehrlann: **[typo]**

```suggestion
sent with parameter client_assertion differentiates between the OAuth2 and OpenID Connect standard and with this, both standards
```

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/jwt/JwtClientAuthentication.java
```diff
@@ -137,35 +150,121 @@ public boolean validateClientJwt(Map<String, String[]> requestParameters, Client
...
+
+    private static boolean isAudienceSupported(String audience, List<String> audList) {
+        return audience == null || ObjectUtils.isEmpty(audList) || audList.contains(audience);
+    }
+
+    private static JWT parseClientAssertion(String clientAssertion) throws ParseException {
+        return JWTParser.parse(clientAssertion);
+    }
+
+    private static JWTClaimsSet parseClientJWT(JWT clientJWT) throws ParseException {
```
Kehrlann: **[issue]** This method is not parsing. Rename to `getClaimSet(...)` or something similar.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/jwt/JwtClientAuthentication.java
```diff
@@ -137,35 +150,121 @@ public boolean validateClientJwt(Map<String, String[]> requestParameters, Client
...
+        return clientJWT != null ? clientJWT.getJWTClaimsSet() : null;
+    }
+
+    private static String getClientId(JWTClaimsSet clientToken) {
+        if (clientToken != null && clientToken.getSubject() != null && clientToken.getIssuer() != null &&
+                clientToken.getSubject().equals(clientToken.getIssuer()) && clientToken.getAudience() != null && clientToken.getJWTID() != null &&
+                clientToken.getExpirationTime() != null) {
```
Kehrlann, line 202: **[suggestion, non-blocking]** Maybe we should take the opportunity and split out the part of this method that does "get client ID", and the part that does "validate required claims"... which should probably go in `validateClientJWToken` rather than returning null here.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/jwt/JwtClientAuthentication.java
```diff
@@ -137,35 +150,121 @@ public boolean validateClientJwt(Map<String, String[]> requestParameters, Client
...
+    private JWKSet getUaaJWKSet() throws ParseException {
+        Map<String, KeyInfo> keyInfoMap = keyInfoService.getKeys();
+        List<JWK> jwkList = new ArrayList<>(keyInfoMap.size());
+        for (KeyInfo entry : keyInfoMap.values()) {
+            jwkList.add(JWK.parse(entry.getJwkMap()));
+        }
+        return new JWKSet(jwkList);
+    }
+
+    private JWKSet getTrustedJwksSet(JWTClaimsSet clientClaims) throws OidcMetadataFetchingException, ParseException, MalformedURLException {
```
Kehrlann: **[suggestion]** Only pass the `issuer` value as a parameter, not the whole `JWTClaimSet`.

model/src/main/java/org/cloudfoundry/identity/uaa/oauth/client/ClientJwtCredential.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @JsonIgnore
```
Kehrlann: **[question]** is `@JsonIgnore` required on static methods?

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/PrivateKeyJwtClientAuthIT.java
```diff
@@ -111,6 +111,41 @@ public void testClientCredentialGrantWithClientUsingPrivateKeyJwtAndExpectValidT
...
+                "client_id", usedClientId,
+                "sub", usedClientId,
+                "zid", "uaa",
+                "iss", "http://localhost:8080/uaa/oauth/token",
+                "grant_type", "client_credentials"
+        );
+        // When
+        String accessToken = getClientCredentialsGrantToken(usedClientId, "access_token", OK);
+        // Then
+        assertNotNull(accessToken);
```
Kehrlann:
```suggestion
        assertThat(accessToken).isNotNull();
```

uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/PrivateKeyJwtClientAuthIT.java
```diff
@@ -111,6 +111,41 @@ public void testClientCredentialGrantWithClientUsingPrivateKeyJwtAndExpectValidT
...
+                "client_id", usedClientId,
+                "sub", usedClientId,
+                "zid", "uaa",
+                "iss", "http://localhost:8080/uaa/oauth/token",
+                "grant_type", "client_credentials"
+        );
+        // When
+        accessToken = getClientCredentialsGrantTokenFromJwt(usedClientId, accessToken, "access_token", OK);
+        // Then
+        assertNotNull(accessToken);
```
Kehrlann:
```suggestion
        assertThat(accessToken).isNotNull();
```

server/src/test/java/org/cloudfoundry/identity/uaa/zone/MultitenantJdbcClientDetailsServiceTests.java
```diff
@@ -568,6 +569,40 @@ void deleteClientJwt() {
...
+    void updateFederatedClientJwt() {
+        UaaClientDetails clientDetails = new UaaClientDetails();
+        clientDetails.setClientId("newClientIdWithNoDetails");
+        service.addClientDetails(clientDetails);
+        service.addClientJwtCredential(clientDetails.getClientId(), new ClientJwtCredential("subject", "issuer", null), currentZoneId, true);
+
+        Map<String, Object> map = jdbcTemplate.queryForMap(SELECT_SQL,
+                "newClientIdWithNoDetails");
+
+        assertEquals("newClientIdWithNoDetails", map.get("client_id"));
```
Kehrlann: **[suggestion, blocking]** From what I understand we're aligning on AssertJ assertions in gh-3195.

Use `assertThat(map).containsEntry("client_id", "newClientIdWithNoDetails")` or `assertThat(map.get("client_id")).isEqualTo("newClientIdWithNoDetails")`.

Same for the rest of the file.

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/token/KeyInfoServiceTests.java
```diff
@@ -91,6 +96,17 @@ void testSignedProviderAsymmetricKeys() {
...
+        List<JWK> jwkList = new ArrayList<>();
+        keyInfoService.getKeys().values().forEach(keyInfo -> {
+            try {
+                jwkList.add(JWK.parse(keyInfo.getJwkMap()));
+            } catch (ParseException e) {
+                // ignore
+            }
+        });
+        jwkSet = new JWKSet(jwkList);
+        assertNotNull(jwkSet);
```
Kehrlann: **[question]** What does this test actually do?

I have not tested it but if. you remove the `.forEach` bit, you get an empty `jwkList`, the `JWKSet` is not null and the test still passes?

server/src/main/java/org/cloudfoundry/identity/uaa/client/ClientAdminBootstrap.java
```diff
@@ -251,6 +258,14 @@ private void addNewClients() {
         }
     }
 
+    private static void writeJwtClientConfiguration(ClientJwtConfiguration keyConfig, UaaClientDetails client) {
```
Kehrlann, line 261: **[suggestion, non-blocking]** I like re-using the method here, but I think in the case of `JWT_CREDS`, the validation is always true. 

The is JWT credentials are parsed with `ClientJwtCredential.parse` ; and `getCleanString()` re-writes it as JSON with `JsonUtils.writeValueAsString`. So I don't think `getCleanString()` can ever fail in this case.

Therefore, I would not use `getCleanString()` when validating the `JWT_CREDS` case, and I would not modify the `getCleanString()` method at all - `getCleanString()` is really just a way of validating some config and should probably be called something different.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/jwt/JwtClientAuthentication.java
```diff
@@ -137,35 +150,121 @@ public boolean validateClientJwt(Map<String, String[]> requestParameters, Client
...
+                    }
                     throw new BadCredentialsException("Wrong client_assertion");
                 }
-                return clientId.equals(validateClientJWToken(JWTParser.parse(clientAssertion), oidcMetadataFetcher == null ? new JWKSet() :
+                // validate token according to private_key_jwt with OIDC
+                return clientId.equals(validateClientJWToken(clientJWT, oidcMetadataFetcher == null ? new JWKSet() :
                                 JWKSet.parse(oidcMetadataFetcher.fetchWebKeySet(clientJwtConfiguration).getKeySetMap()),
-                        clientId, keyInfoService.getTokenEndpointUrl()).getSubject());
-            } catch (ParseException | URISyntaxException | OidcMetadataFetchingException e) {
+                        clientId, clientId, keyInfoService.getTokenEndpointUrl()).getSubject());
```
Kehrlann: **[suggestion, non-blocking]** Document the two flows as comments in the code, with links to the specs. When reading this, it took me a little while to understand why `expectedIssuer == clientId` here (OIDC flow) and not the issuer.

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/jwt/JwtClientAuthentication.java
```diff
@@ -137,35 +150,121 @@ public boolean validateClientJwt(Map<String, String[]> requestParameters, Client
         if (GRANT_TYPE.equals(UaaStringUtils.getSafeParameterValue(requestParameters.get(CLIENT_ASSERTION_TYPE)))) {
             try {
                 String clientAssertion = UaaStringUtils.getSafeParameterValue(requestParameters.get(CLIENT_ASSERTION));
-                if (!clientId.equals(getClientId(clientAssertion))) {
+                JWT clientJWT = parseClientAssertion(clientAssertion);
+                JWTClaimsSet clientClaims = parseClientJWT(clientJWT);
+                if (!clientId.equals(getClientId(clientClaims))) {
```
Kehrlann: **[suggestion]** Worth clarifying that this is for "authorization grant" case of the OAuth2 flow, where the `sub` is the subject of the user who got the token, not the client ID.

server/src/test/java/org/cloudfoundry/identity/uaa/audit/JdbcUnsuccessfulLoginCountingAuditServiceTests.java
```diff
@@ -29,7 +29,7 @@
 import static org.mockito.Mockito.*;
 
 @WithDatabaseContext
-class JdbcUnsuccessfulLoginCountingAuditServiceTests {
+class       JdbcUnsuccessfulLoginCountingAuditServiceTests {
```
Kehrlann:
```suggestion
class JdbcUnsuccessfulLoginCountingAuditServiceTests {
```

server/src/test/java/org/cloudfoundry/identity/uaa/audit/LoggingAuditServiceTest.java
```diff
@@ -37,7 +37,7 @@ void log_format_whenThereIsAnAuthType() {
         ArgumentCaptor<String> stringCaptor = ArgumentCaptor.forClass(String.class);
         verify(mockLogger).info(stringCaptor.capture());
         String logMessage = stringCaptor.getValue();
-        assertThat(logMessage, is("PasswordChangeFailure ('theData'): principal=thePrincipalId, origin=[theOrigin], identityZoneId=[theZoneId], authenticationType=[theAuthType]"));
+        assertThat(logMessage, is("PasswordChangeFailure ('theData'): principal=thePrincipalId, origin=[theOrigin], identityZoneId=[theZoneId], authenticationType=[theAuthType], detailedDescription=[theDescription]"));
```
Kehrlann: **[suggestion]** use assertJ instead, see other comments - there will probably be conflicts here with gh-3195

server/src/test/java/org/cloudfoundry/identity/uaa/client/ClientAdminBootstrapTests.java
```diff
@@ -296,6 +296,20 @@ void simpleInvalidClientWithJwkSet() throws Exception {
...
+    void simpleAddClientWithClientJwtCredendial() {
+        Map<String, Object> map = new HashMap<>();
+        map.put("id", "foo-jwks");
+        map.put("secret", "bar");
+        map.put("scope", "openid");
+        map.put("authorized-grant-types", GRANT_TYPE_AUTHORIZATION_CODE);
+        map.put("authorities", "uaa.none");
+        map.put("redirect-uri", "http://localhost/callback");
+        map.put("jwt_creds", "[{\"iss\":\"http://localhost:8080/uaa/oauth/token\",\"sub\":\"foo-jwt\"}]");
+        UaaClientDetails clientDetails = (UaaClientDetails) doSimpleTest(map, clientAdminBootstrap, multitenantJdbcClientDetailsService, clients);
```
Kehrlann, line 304: **[thought, non-blocking]** `doSimpleTest` is a questionable name 😅

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/jwt/JwtClientAuthentication.java
```diff
@@ -137,35 +150,121 @@ public boolean validateClientJwt(Map<String, String[]> requestParameters, Client
         if (GRANT_TYPE.equals(UaaStringUtils.getSafeParameterValue(requestParameters.get(CLIENT_ASSERTION_TYPE)))) {
```
Kehrlann, line 150: **[suggestion]** noticed that this is NOT the grant_type as per spec, it's the client_assertion_type ; the grant type exist and is defined in `TokenConstants#GRANT_TYPE_JWT_BEARER`. We should probably use this PR to rename this parameter, maybe `CLIENT_ASSERTION_TYPE_JWT_BEARER` or something similar.

Kehrlann (changes requested):
First of all, great work, this is a complicated topic. I had a hard time wrapping my head around it 🤯 


My main issue is that this JWT authentication + grant type is complex and hard to understand. There are three things handled in the same place (OAuth2 `jwt-bearer` grant type + OAuth2 client authentication + OIDC client authentication). It makes it really hard to understand what should happen and when. 
**[suggestion]** While we are touching this, would be super helpful to decompose the whole flow a bit rather than add more code in the same place. Maybe have multiple `ClientJwtAuthentication` classes or methods that map to a specific flow with a link to the spec.
I don't have an easy suggestion of what should be done but we should at least consider it.


I also left a few suggestions (use assertJ, a typo here and there, suggestions for method names), some are easy fixes and some are just non-blocking suggestions.

model/src/main/java/org/cloudfoundry/identity/uaa/oauth/client/ClientJwtCredential.java
strehle: not sure, but json ignore is what I know from the past, do you want me to remove it ?

Kehrlann: I tried without `@JsonIgnore` and it works ; let's remove it for that static method.

---

Experiment:


```java
static class FooData {

    private final String name;
    private final int age;


    FooData(String name, int age) {
        this.name = name;
        this.age = age;
    }

    public String getName() {
        return name;
    }

    public int getAge() {
        return age;
    }

    @JsonIgnore
    public String getEmail() {
        return name + "@example.com";
    }

    public static String staticData() {
        return "some-stuff";
    }
}
```


`new FooData("daniel", 36)` serializes to `{"name":"daniel","age":36}`, static methods do not need to be `@JsonIgnore`'d.

model/src/main/java/org/cloudfoundry/identity/uaa/oauth/client/ClientJwtCredential.java
strehle: changed

server/src/main/java/org/cloudfoundry/identity/uaa/authentication/ClientDetailsAuthenticationProvider.java
strehle: needed to have the value in audit log , also in error cases

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/clients/ClientAdminEndpointDocs.java
strehle: done

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/jwt/JwtClientAuthentication.java
strehle: changed as discussed

server/src/main/java/org/cloudfoundry/identity/uaa/client/ClientAdminBootstrap.java
strehle: we can discuss, but getCleanString returns for an empty config null, otherwise the serializer from json returns something like {} for objects or maps

Kehrlann: We could use something like Spring's `CollectionUtils.isEmpty` to check this rather than going back-and-forth with this writing and parsing.

But this suggestion is non-blocking and the code is fine as-is.

Kehrlann (approved):
LGTM

duanemay (approved):

PR #3222, opened 2025-01-07
Do not add whitespaces in XML signature of /saml/metadata

### LINKED ISSUE #3220: Review SAML metadata changes
We noticed that the ID attribute of the SAML metadata changed between two versions of the UAA. 

When setting entityID=`https://localhost:443` in the UAA configuration file (`uaa.yml setting login.entitytID`)
 
UAA 77.19.0

```
ID="https___localhost_443" entityID="https://localhost:443"
```

UAA 77.23.0

```
ID="https://localhost:443" entityID="https://localhost:443"
```

The ID attribute should be a NCName and not contain colons
https://www.datypic.com/sc/xsd/t-xsd_ID.html (type="ID")

I'm opening this issue to track if other changes have occurred, and if we need to be backwards compatible

strehle:
Add property, e.g. https://stackoverflow.com/questions/4728300/how-to-produce-xml-signature-with-no-whitespaces-and-line-breaks-in-java

Signature value ONLY of endpoint /saml/metadata contain the line breaks, other signatures dont show such chars

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/IdentityZoneConfig.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/IdentityZoneConfig.java
@@ -26,6 +26,7 @@ public class IdentityZoneConfig {
             log.info("New securerandom.strongAlgorithms: {}", newStrongAlgorithm);
             Security.setProperty("securerandom.strongAlgorithms", newStrongAlgorithm);
         }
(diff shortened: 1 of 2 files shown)


strehle:
@Kehrlann I am wondering why the line breaks ( &#13; ) only in metadata and therefore do you know how spring-security creates the signature, so that there are no extra line breaks.

Kehrlann:
@strehle I don't have the specifics in mind but this is probably generated by OpenSAML, not Spring Security itself.

Kehrlann (approved):

PR #3230, opened 2025-01-09
Helper scripts for certificates

fhanik:
I anticipate work with certificates, trusting certificates etc. 
This adds a simple directories of how to create a Certificate Authority and self sign certificates using the CA.


fhanik:
I will make this part of the LDAP requests, since I'm working on TLS/Certificates in there

PR #3233, opened 2025-01-16
Allow jwt bearer with empty secret

### LINKED ISSUE #3286: Allow jwt bearer with empty secret - password grant allows empty secrets
JWT bearer should replace password grant. The authentication flow is similar, because with one call you get a USER token from UAA.

JWT bearer has limits and more limits then password grant has, e.g. the cf client allows to get a password grant token with cf client and empty secret. The creation of a oauth2 client in uaa with an empty secret and jwt bearer is dis-allowed, but I dont see a security reason for this.

strehle:
Remove checks in REST call

Move the check for existing secrets to the
grant flow (runtime) and remove it from REST (configuration).


strehle:
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=3233

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/token/JwtTokenGranter.java
```diff
@@ -13,11 +13,18 @@
...
+
+import static org.cloudfoundry.identity.uaa.oauth.token.TokenConstants.CLIENT_AUTH_PRIVATE_KEY_JWT;
+import static org.cloudfoundry.identity.uaa.oauth.token.TokenConstants.CLIENT_AUTH_SECRET;
 import static org.cloudfoundry.identity.uaa.oauth.token.TokenConstants.GRANT_TYPE_JWT_BEARER;
 
 public class JwtTokenGranter extends AbstractTokenGranter {
     final DefaultSecurityContextAccessor defaultSecurityContextAccessor;
 
+    private static final List<String> ALLOWED_AUTH_METHODS = List.of(CLIENT_AUTH_SECRET, CLIENT_AUTH_PRIVATE_KEY_JWT);
+            //comment I would add here CLIENT_AUTH_EMPTY to allow empty secret for jwt bearer, same as allowed for password
```
strehle: With this, the change is not change but only a refactoring.

As mentioned currently we reject in clients for JWT bearer empty secret but that is strange and we should discuss

strehle:
Not needed

PR #3234, opened 2025-01-16
Solve issue 3226

### LINKED ISSUE #3226: SAML "invalid destination" error message should include `location`, as well as `destination`
When SAML authentication fails because of "invalid destination", because the `destination` does not match the `location` (see patch below), the resulting error message does not include the `location`.

The patch below updates the error message to include the value of `location` which is being compared against `destination`.

```patch
diff --git a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/OpenSaml4AuthenticationProvider.java b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/OpenSaml4AuthenticationProvider.java
index e484a0f13..b5398e269 100644
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/OpenSaml4AuthenticationProvider.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/OpenSaml4AuthenticationProvider.java
@@ -224,8 +224,8 @@ public final class OpenSaml4AuthenticationProvider implements AuthenticationProv
             String destination = response.getDestination();
             String location = token.getRelyingPartyRegistration().getAssertionConsumerServiceLocation();
             if (StringUtils.hasText(destination) && !destination.equals(location)) {
-                String message = "Invalid destination [" + destination + "] for SAML response [" + response.getID()
-                        + "]";
+                String message = "Invalid destination [%s], location [%s] combo for samle response [%s]"
+                        .formatted(destination, location, response.getID());
                 result = result.concat(new Saml2Error(Saml2ErrorCodes.INVALID_DESTINATION, message));
             }
             String assertingPartyEntityId = token.getRelyingPartyRegistration()
```

strehle:


duanemay:
The issue seems to be that when UAA is running on a standard port :80 or :443 and that port is included in the metadata. The destination within the SAML response includes the port, but the location from the relying party registration does not. Therefore the comparison does not match and authentication fails.

aramprice:
Agreed that the default port stripping is probably at the root of this.

The purpose behind this change is to expose both of the elements being compared in the conditional so that it is clearer _why_ this particular error is returned by UAA.

Having only half of the information in the error message makes it impossible to understand the root cause when this error occurs.

duanemay (approved):
debug info is fine, need to circle back on fix

PR #3240, opened 2025-01-16
Migrate DB configuration from XML to Java

### LINKED ISSUE #3238: Consider removing/deprecating FlywayMigrationRunner.java
The probable use case was to run a database migration prior to UAA starting.

Notes:
- https://github.com/cloudfoundry/uaa/commit/55c858f80a9bfc36b320dae4b5a75d8c14c06acf
- Pivotal Tracker: https://www.pivotaltracker.com/n/projects/997278/stories/175232621

Kehrlann:
## Context

Migrating XML config to java-based config. This PR focuses on moving a big chunck of the database configuration classes to Java config. In the process, it removes `data-source.xml`.

### Environment / profiles / database platform

Most profile data was moved from `env.xml` into `DatabaseConfiguration` + `application-{profiles}.properties` files. Database specific details are now encapsulated in `DatabasePlatform` (a superset of the former `Vendor` class).

The defaults have been moved from XML to `application-{profile}` files, with explicit references in the configuration classes (e.g. `PostgresConfiguration` references `application-postgresql.properties` explicitly).

Introduced a "default-url" property in the same place ; it is not intended to be exposed.

### Properties

Introduced a central `DatabaseProperties` class to gather all database-specific properties. It has a lot of properties, and it's unclear if they are all used - most of them are undocumented. But they have been exposed for a long time so I prefer to keep them.

### Removed `UaaDatabaseName` and `DatabaseUrlModifier`

The first of these classes was a test utility ; removed it from prod. The second was a bean to add `connectTimeout` at the end of the jdbc URL. Aligned both classes on `JdbcUrlCustomizer`, which update a jdbc-url with code. This was, the DB name customization can happen in the test package only.

### Removed `FlywayMigrationRunner`

Closes gh-3238 ; this class is never used.

### Removed some explicit beans in favor of Boot auto-configuration

Some beans e.g. `JdbcTemplate` do not need to be explicitly wired, as Boot will provide some for you. Referenced all the beans and configuration classes in `DatabaseConfiguration`'s javadoc.

This requires importing some auto-configurations in `@WithDatabaseContext`, and enabling all auto-configuration in `@DefaultTestContext`. There are some configuration conflicts in `@DefaultTestContext` and so some auto-configurations must be excluded.


server/src/main/java/org/cloudfoundry/identity/uaa/client/JdbcQueryableClientDetailsService.java
```diff
@@ -41,7 +41,7 @@ public class JdbcQueryableClientDetailsService
 
     public JdbcQueryableClientDetailsService(
             final @Qualifier("jdbcClientDetailsService") MultitenantJdbcClientDetailsService delegate,
-            final @Qualifier("namedJdbcTemplate") NamedParameterJdbcTemplate jdbcTemplate,
+            final NamedParameterJdbcTemplate jdbcTemplate,
```
strehle, line 44: what is the bean name if you remove the own name ?

strehle:
here sonar 
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=3240

if you use lombok the coverage should be ok, because you dont have setters/getters in there to be covered.

strehle (commented):
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=3240

server/src/main/java/org/cloudfoundry/identity/uaa/client/JdbcQueryableClientDetailsService.java
Kehrlann: Now we're relying on Spring Boot autoconfiguration, through `JdbcTemplateAutoConfiguration` > `NamedParameterJdbcTemplateConfiguration`. In the [source of that configuration class](https://github.com/spring-projects/spring-boot/blob/88eb74884a191902cfd6a54f01f221d5d25709f7/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/NamedParameterJdbcTemplateConfiguration.java#L40), we can see the name is `namedParameterJdbcTemplate`. But since we only have one bean candidate for this type, we should let the container do dependency resolution by type and not by name. This avoids coupling the the Spring Boot auto-configuration details via a string.


For info I put a message in [the associated commit](https://github.com/cloudfoundry/uaa/pull/3240/commits/01ce31722f2022bc42474a7d82a1ed5cd85efe43) that should help developers realize why we don't need the qualifier anymore.

strehle: ok, was more a question and now understand, before  we had an own NamedParameterJdbcTemplate  instance and now we rely on the default, which is more ore less same implementation, thanks

Kehrlann:
Addressed Sonar coverage warning by using lombok in `DatabaseProperties`.

strehle (approved):
LGTM

duanemay (approved):
👍

PR #3246, opened 2025-01-18
More waiting time for webdriver

strehle:
5 sec is not enough in some cases
In other scenarios we have about 30 sec.

https://github.com/cloudfoundry/uaa/blob/develop/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/feature/DefaultIntegrationTestConfig.java#L35-L37

### DIFF
--- a/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/pageObjects/Page.java
+++ b/uaa/src/test/java/org/cloudfoundry/identity/uaa/integration/pageObjects/Page.java
@@ -20,7 +20,8 @@ public class Page {
 
     public Page(WebDriver driver) {
         this.driver = driver;
-        driver.manage().timeouts().implicitlyWait(Duration.ofSeconds(5));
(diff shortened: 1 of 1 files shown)


duanemay:
The implicit wait increases test script execution time, if we use explicit waits or convert to using asserts on the selenium tests to awaitility then we can use the `await().atMost(Duration.ofSeconds(30)).untilAsserted(() -> ...)` which will check the condition repeatedly until it passes then proceed. 

You can see the these Pages objects have been moved to awaitility, lower in the file. We can definitely increase that timeout and should only affect the test execution times where it fails to load the element.

strehle:
> The implicit wait increases test script execution time, if we use explicit waits or convert to using asserts on the selenium tests to awaitility then we can use the `await().atMost(Duration.ofSeconds(30)).untilAsserted(() -> ...)` which will check the condition repeatedly until it passes then proceed.
> 
> You can see the these Pages objects have been moved to awaitility, lower in the file. We can definitely increase that timeout and should only affect the test execution times where it fails to load the element.

ok, so we can also revert this..

I have issues with latest chrom... 132 see new PRs
https://github.com/cloudfoundry/uaa-ci/pull/25
https://github.com/cloudfoundry/uaa/pull/3248

Issue #3145, opened 2024-11-20
[Feature] Paging for retrieval of all identity providers for a given identity zone


torsten-sap:
### What version of UAA are you running?

77.18.0

### How are you deploying the UAA?

I am deploying the UAA
- using cf-deployment

### What did you do?

Retrieval of all identity providers via endpoint [`/identity-providers`](https://docs.cloudfoundry.org/api/uaa/version/77.18.0/index.html#retrieve-all) for a given identity zone.

### What did you expect to see? What goal are you trying to achieve with the UAA?

Besides the option to retrieve all identity providers, the caller of `/identity-providers` has the option to use paging to restrict the result set of the UAA response. The paging may be similar to the SCIM API ([`/Users`](https://docs.cloudfoundry.org/api/uaa/version/77.18.0/index.html#list), [`/Groups`](https://docs.cloudfoundry.org/api/uaa/version/77.18.0/index.html#list-2)). With regards to the page size, the API may define a default (e.g. 100) and a maximum value (e.g. 500) if paging is requested.

The default behaviour of the API should stay as is, namely returning all identity providers for the given identity zone.

The goals for identity zones with a very high amount of identity providers are:
* Allow the caller of the API to use paging to ensure appropriate response times

### What did you see instead?

No paging mechanism supported for endpoint `/identity-providers`.

PR #3251, opened 2025-01-21
Introduce optional Cache for Login Info IdPs

adrianhoelzl-sap:
Introduce a cache for the SAML or OAuth identity providers returned in the login info endpoint. By default, the cache is disabled.

If enabled, the cached IdPs are neither synchronized between the different UAA instances and are not invalidated whenever an IdP is created, updated or deleted. Therefore, during a cache period, the returned set might be inconsistent.


strehle:
I am not convinced about cache here... in the past there were several PR´s from you in order to optimize the logic and I thought now we don't have situations ( only error situations ) where all IdPs are fetched ( means select * from identity_provider XXX) 

Instead of caching IdPs I would work on https://github.com/cloudfoundry/uaa/issues/3145 and would Disable  option `allowOriginLoop `, see 
https://github.com/cloudfoundry/uaa/blob/develop/model/src/main/java/org/cloudfoundry/identity/uaa/zone/UserConfig.java#L86-L87, so that either you login with origin or issuer or error. no further lookup to all IDPs from type SAML / OIDC / OAUTH etc.

duanemay:
Replaced by #3254

PR #3252, opened 2025-01-21
fix: concurrent group membership race conditions

mikeroda:
- catch the DuplicateKeyException when adding a member to a group. Needed due to https://github.com/cloudfoundry/uaa/commit/f8188564facddee35600a23580eb845194337753

- catch MemberNotFoundException when attempting to remove member from a group.


strehle:
@mikeroda Thanks for your PR. We will include the exception ignore, your test has a problem, can you please fix test ScimUserBootstrapTests > concurrentAuthEventsRaceCondition() FAILED

mikeroda:
> @mikeroda Thanks for your PR. We will include the exception ignore, your test has a problem, can you please fix test ScimUserBootstrapTests > concurrentAuthEventsRaceCondition() FAILED

With the latest commit I'm not able to reproduce the failure, and I can't see what is failing here. We could revert the change I made to that test case and then it will always work, but it doesn't really test effectively then. Do you have any suggestions?

strehle (approved):

strehle:
merged this and do the cleanup with https://github.com/cloudfoundry/uaa/pull/3275

PR #3253, opened 2025-01-21
Deprecate `default` profile

fhanik:


strehle (approved):
LGTM, I would prefer to wait with merge until we have a green pipeline, but no blocker

duanemay (approved):

strehle:
Hi, 
I underestimated this change because of deprecate term. CI is broken because deployment on https://login.uaa-acceptance.cf-app.com/ does not know marissa (our standard user).

So maybe we should stay with the integrated setup defaults but remove default everywhere we dont need it... or use a term like demo, test, etc.


@Kehrlann FYI

PR #3254, opened 2025-01-22
Login Info Endpoint: optimize OAuth IdPs for JSON Response

adrianhoelzl-sap:
(alternative for https://github.com/cloudfoundry/uaa/pull/3251)

This PR optimizes the login info endpoint so that it does not fetch all OAuth IdPs when a JSON response is requested.

Prior to the optimization, all OAuth IdPs are queried and then used for determining a potential redirect to the corresponding external IdP. However, for JSON responses the redirect is ignored anyway.

The optimization introduces the following changes to the JSON response, which are in our opinion acceptable:

- `showLoginLinks` is no longer returned if there is also no SAML IdP with `showSamlLink` enabled
  - this property is only relevant for the login page and therefore obsolete for the JSON response
  - in the JSON response, only SAML IdPs are part of the `idpDefinitions` anyway
- the `passcode` prompt is no longer removed if there is no SAML/OIDC IdP with `showSamlLink`/`showLinkText` in the IdP definitions list
  - password grants with a one-time passcode also work with the UAA user store

The first few commits in this PR are refactorings of the endpoint handler, they don't change any behavior. We advise to go through the PR from commit to commit.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java
@@ -73,7 +73,6 @@ import java.sql.Timestamp;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.Comparator;
 import java.util.Date;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


strehle (approved):

duanemay:
Should #3251 be closed, with this one?

strehle (approved):
approved already... do you need anything from other approvers ?
btw. can you answer question from @duanemay

duanemay (approved):

PR #3255, opened 2025-01-23
Ensure that ldap is fully backwards compatible, including user01

fhanik:
@strehle Here are the LDAP updates


strehle:
can we omit the name
https://github.com/fhanik/uaa/blob/pr/fix-ldap-readme/scripts/docker-compose.yml#L1

I run my dev env on ubuntu 24 but use a docker-compose and receive error from 
https://stackoverflow.com/questions/73427249/name-does-not-match-any-of-the-regexes-x-while-trying-to-set-the-project

strehle (approved):

strehle:
@fhanik can you rebase this , please

duanemay (approved):

PR #3262, opened 2025-01-30
Move login-ui.xml to java configuration

Kehrlann:
## Context & goal

In the XML -> Java migration for UAA configuration, we are moving configuration out of `login-ui.xml`.

This PR addresses some of the challenges, and moves 25-30% of `login-ui.xml` to java configuration.

### Spring Security incompatibilities

Spring Security does not support XML configuration + Java configuration in parallel, particularly declaring filter chains in both XML and Java.
The java filter chains beans are created after the full Spring Security XML configuration is registered, and are therefore never called.

There are workarounds introduced in 90d4a184, and they require creating another `FilterChainProxy` that has all the chains inside, and wiring that in `web.xml` instead of the usual `springSecurityFilterChain`. While it does introduce a bit of complexity, the upside is being able to move filter chains one by one and not have to do everything wholesale (there are 68 filter chains total).

Please take a look at this commit.

### Spring Boot vs non-Boot differences

There are some behavior differences in Boot vs non-Boot, around including `Model` data in redirect responses and jackson configuration for datetimes, covered in f4d5ca74.

### YamlConfigurationValidator

It was broken, see gh-3279 . I noticed because I created a bean called `uaaConfig` that conflicted with the validator.
I fixed the basics and introduced tests for it in 1a36b32c.

### Dead code

There were a few redundant `<mvc:...>` and `<oauth:...>` annotations.

There was also a `catchAllFilterChainProxy` that we determined with @fhanik was never actually used in production and only there for testing during development.

### Sonar & code coverage

Sonar is unhappy with code coverage because the majority of new code here is configuration classes and bean, and JaCoCo can't find coverage for it. Not going to block merging because of this.

### DIFF
--- a/scripts/cargo/uaa.yml
+++ b/scripts/cargo/uaa.yml
@@ -177,12 +177,14 @@ oauth:
     override: true
   clients:
     admin:
+      id: admin
       authorized-grant-types: client_credentials
       scope: uaa.none
       authorities: 'uaa.admin,clients.read,clients.write,clients.secret,clients.trust,scim.read,scim.write,clients.admin'
       secret: "adminsecret"
       jwks: '{"alg":"RS256","e":"AQAB","kid":"cUiuzP1rw1zm9MV8F0vtrws7BLc","kty":"RSA","n":"rWuIqrVV8kuqeorvRuLio1_pdQm_z7HZJKIcCD5SQqGO0AsKyf1xa5TPzHM0lqEh2GcPTer4u7MYQZzXAAvzOsSaTmgSlenLKDYCDZy2bwOjK0izVLbJwYqiiqyiMGhKeWsYokyDNoYaefjz8izDrp47XDHnwC2eeyJ43cE8GP0JJXRyxIPFecO8rfpe3AzTrHszJ9lPSX9E8QGppSFmcnUDUQYDRipNMzXXp2FHdR7T2MZkvxzjFhVSSMiaDTmAca-Wv_Uct2HpOfC3IuKSy1jpu8yr_GT6aBsDkt1XC1iARuFf9dE83R39oNgvVMICPjeWgNoyhK-ddQAUnRDeqw"}'
     cf:
+      id: cf
       secret: ''
       authorized-grant-types: 'implicit,password,refresh_token'
       scope: 'uaa.user,cloud_controller.read,cloud_controller.write,openid,password.write,scim.userids,cloud_controller.admin,scim.read,scim.write'
@@ -190,6 +192,7 @@ oauth:
       authorities: uaa.none
       autoapprove: 'true'
     app:
+      id: app
       secret: appclientsecret
       authorized-grant-types: password,implicit,authorization_code,client_credentials,refresh_token
       scope: cloud_controller.read,cloud_controller.write,openid,password.write,scim.userids,organizations.acme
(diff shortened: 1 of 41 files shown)


duanemay:
Interesting that the `id:` fields should have been required in uaa.yml, but currently works without. I do see the messages in the logs now, when id is missing. My only concern is if we intend to fail startup with the validation, it will it break existing customer configuration. Fine with a major release

Kehrlann:
@duanemay - that's a good point, we don't want to break existing customer's setup. This PR is safe in this regard.

Currently, for validation to crash the startup, users need to opt-in by using the `strict` profile ; otherwise it's a no-op. I think very few people are using this because our validation is out of sync with the actual acceptable configuration (see #3279). Maybe if users are not using any of the "new" features, the `strict` profile still works, but I don't know when the cutoff date is.  Lastly, the `strict` profile is undocumented.

This PR would enable users to use the `strict` profile again even with newer features. This PR also adds some testing ensuring that the validation works in `strict` mode with our cargo config, so that user may opt-in to that if they wish.


Specifically re: the "id" requirement, it has been in the code base for a very long time ([source](https://github.com/cloudfoundry/uaa/blame/ba3b9ca80d62a27033b167d465f176e6fe6b7e72/server/src/main/java/org/cloudfoundry/identity/uaa/impl/config/UaaConfiguration.java#L216-L217), git blame says that the last commit on this was 11 years ago with title "Formatting"). Our cargo / test configs were broken according to the validation rules already in place ; but the app could function nonetheless. And since we never run the `strict` profile, we had not noticed.

duanemay (approved):

scripts/cargo/uaa.yml
```diff
@@ -177,19 +177,22 @@ oauth:
     override: true
   clients:
     admin:
+      id: admin
```
strehle, line 180: is it now required to have id in yaml ? -> if yes we have to communicate it to cf-deployment and we need to check our oidc-integration test -> https://github.com/cloudfoundry/uaa/blob/develop/scripts/saml/saml-legacy-uaa.yml#L162-L308

strehle (commented):
Did you decided to have a release build of UAA before or after this ?

We should have a release build soon, because we have some feature/changes now

scripts/cargo/uaa.yml
Kehrlann: Please see [this comment](https://github.com/cloudfoundry/uaa/pull/3262#issuecomment-2647487525)

The tl;dr is:
- The validation code _already_ requires this `id` prop, and that has been the case for over 10 years. So if you run with profile=`strict` the app crashes without an `id` ([source](https://github.com/cloudfoundry/uaa/blame/ba3b9ca80d62a27033b167d465f176e6fe6b7e72/server/src/main/java/org/cloudfoundry/identity/uaa/impl/config/UaaConfiguration.java#L216-L217)).
- But the UAA actually works fine without the `id` (apart from validation)
- The PR does not change anything ; it "just" ensures that if we run our cargo-based app with profile=`strict` the app doesn't crash, but it does not impact customers in any way.

strehle: ok, so no adoption for https://bosh.ci.cloudfoundry.org/teams/uaa/pipelines/uaa-acceptance-gcp/jobs/push-oidc10/builds/465

strehle (approved):
you can merge and then we can have a new release

PR #3272, opened 2025-02-04
Dockerize SimpleSAMLPhp for Local Integration Tests

fhanik:
This PR 

1. Creates a docker container for SimpleSAMLPHP that launches on localhost:9090
2. Allows tests to be run using `-Dintegration.test.saml.url=http://localhost:9090/simplesaml`

Notes:

- As of now, uaa-acceptance is running an older version of SimpleSAMLPhp, so the code has to take that into account. Once we have upgraded that, there are some simple if-then-else statements that can be removed

- We are working towards the ability to run all tests on a developer workstation. We have completed MySQL, PostgreSQL, OpenLDAP and now with this SAML. Only thing missing are our `OIDC10` tests 

- The authsources.php and saml20-sp-remote.php files are taken from our original SimpleSAMLPhp deployment under github user cf-identity


Kehrlann (approved):
Nice! I like the double commit, one for the container and one for the code.

PR #3301, opened 2025-02-19
Create client with jwt bearer with empty secret

### LINKED ISSUE #3286: Allow jwt bearer with empty secret - password grant allows empty secrets

strehle:
Remove checks in REST call

Alternative PR to https://github.com/cloudfoundry/uaa/pull/3233


strehle:
@duanemay @fhanik This is a smaler PR but similar to https://github.com/cloudfoundry/uaa/pull/3233
Here I only removed the check in REST calls about empty secrets in combination with JWT bearer (for password grant we dont care so JWT should not reject it).

In PR https://github.com/cloudfoundry/uaa/pull/3233 I also do explicit validation of client authentication.
To me, this PR here is enough but with https://github.com/cloudfoundry/uaa/pull/3233 we would have insurance that missing secret cannot succeed in JWT bearer, but at least an empty secret

duanemay (approved):
LGTM

PR #3305, opened 2025-02-20
Prevent secret matches with more than 72 chars.

strehle:
Before:

uaac client add test --name test --authorized_grant_types client_credentials -s 'this-is-a-secret-which-is-not-longer-then-what-bcrypt-can-handle-matches'

uaac client client get test -s 'this-is-a-secret-which-is-not-longer-then-what-bcrypt-can-handle-matches'
works but also 
uaac client client get test -s 'this-is-a-secret-which-is-not-longer-then-what-bcrypt-can-handle-matchesXXXXXXXXXX'
uaac client client get test -s 'this-is-a-secret-which-is-not-longer-then-what-bcrypt-can-handle-matches-more-than-wanted'

After
uaac client add test --name test --authorized_grant_types client_credentials -s 'this-is-a-secret-which-is-not-longer-then-what-bcrypt-can-handle-matchesXXX'

```
{
  "error": "invalid_client",
  "error_description": "Client Secret must be no more than 72 characters in length."
}
```

Checks with secrets > 72 end always in bad credentials

more details
https://sendoh-daten.medium.com/using-bcrypt-as-password-encoder-can-be-problematic-when-passwords-are-non-ascii-characters-a5ac0478aeed

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/client/ClientAdminBootstrap.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/client/ClientAdminBootstrap.java
@@ -295,7 +295,7 @@ public class ClientAdminBootstrap implements
 
     private static void checkSecretLength(String secret) {
         if (ofNullable(secret).map(String::length).orElse(0) > 72) {
-            throw new InvalidClientDetailsException("Client must not have a secret with more than 72 characters");
+            throw new InvalidClientDetailsException("Client secret must be no more than 72 characters in length.");
         }
     }
 
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/client/ClientAdminBootstrap.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/client/ClientAdminBootstrap.java
@@ -230,7 +230,10 @@ public class ClientAdminBootstrap implements
             try {
                 clientRegistrationService.addClientDetails(client, IdentityZone.getUaaZoneId());
                 if (secondSecret != null) {
+                    checkSecretLength(secondSecret);
                     clientRegistrationService.addClientSecret(clientId, secondSecret, IdentityZone.getUaaZoneId());
+                } else {
+                    checkSecretLength(client.getClientSecret());
                 }
             } catch (ClientAlreadyExistsException e) {
                 if (override) {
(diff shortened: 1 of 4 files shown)


strehle:
@Kehrlann  since you are our spring security expert. Question: is this limit known or do you recommend the workaround ,e.g. sha512(password)  to store not more than 72 ?
We discussed the issue in OSS sync and from UAA side we reject this but not further action

Kehrlann:
@strehle that's a bcrypt limtation, max 72 chars: https://security.stackexchange.com/questions/39849/does-bcrypt-have-a-maximum-password-length

I don't think the proposed change in `BackwardsCompatibleDelegatingPasswordEncoder`works: what if a user has a client registered, with a password > 72 chars ? If they try their password today it works (although only the first 72 characters are checked). If they try after this change they get an exception.

I think the check should be in the creation logic (user provisioning?) and password changes. And you have to be careful, what happens if you have non-ascii characters, how are they encoded, does the length check actually work?

Kehrlann (changes requested):
@strehle that's a bcrypt limtation, max 72 chars: https://security.stackexchange.com/questions/39849/does-bcrypt-have-a-maximum-password-length

I don't think the proposed change in `BackwardsCompatibleDelegatingPasswordEncoder`works: what if a user has a client registered, with a password > 72 chars ? If they try their password today it works (although only the first 72 characters are checked). If they try after this change they get an exception.

I think the check should be in the creation logic (user provisioning?) and password changes. And you have to be careful, what happens if you have non-ascii characters, how are they encoded, does the length check actually work?

server/src/main/java/org/cloudfoundry/identity/uaa/client/ClientAdminBootstrap.java
```diff
@@ -230,7 +230,10 @@ private void addNewClients() {
             try {
                 clientRegistrationService.addClientDetails(client, IdentityZone.getUaaZoneId());
                 if (secondSecret != null) {
+                    checkSecretLength(secondSecret);
                     clientRegistrationService.addClientSecret(clientId, secondSecret, IdentityZone.getUaaZoneId());
+                } else {
+                    checkSecretLength(client.getClientSecret());
```
adrianhoelzl-sap, line 236: Why do we only perform the check if no second secret is set? Any why only after writing it to the DB?

server/src/main/java/org/cloudfoundry/identity/uaa/util/beans/BackwardsCompatibleDelegatingPasswordEncoder.java
```diff
@@ -1,12 +1,15 @@
...
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder;
 import org.springframework.security.crypto.password.PasswordEncoder;
 
 public class BackwardsCompatibleDelegatingPasswordEncoder implements PasswordEncoder {
 
     private static final String OPTIONAL_BCRYPT_PREFIX = "bcrypt";
     private final BCryptPasswordEncoder defaultPasswordEncoder;
```
adrianhoelzl-sap, line 11: Would it make sense to introduce a new password encoder that performs `bcrypt(sha512(password))` (as noted in the article you linked) and use it for any new client? To distinguish between old and new encodings, we could introduce a new non-optional prefix.

strehle:
close this since we should fix it, e.g. https://github.com/cloudfoundry/uaa/issues/3333

PR #3309, opened 2025-02-25
Proxy jwt bearer to support corporate trust

### LINKED ISSUE #3289: Proxy / Forward JWT Bearer in UAA to allow central trust anchor
Problem.

Trust anchor for CF deployments can be centralized, this means, that your password is not in CF but in an IDP (SAML or OIDC). With OIDC IDP you can do password forwarding, means cf login -u user -p password --oirigin idp is generating a password grant to UAA and UAA can forward this the next IDP to forward / check password on IDP level.

Moving away from passwords towards to token based authentication should allow same possiblity.
If you do cf auth --assertion <your-JWT-to-Authenticate> --oirigin idp generates a JWT to UAA, but if the trust of the JWToken is not setup in UAA but in the trusted IDP then this breaks the flow.

Therefore we should support same option for JWT bearer, means, if --oirigin is setup then login_hint in JWT bearer should be check and if the JWT is not trusted ( simply we have no trusted IdP in UAA with this issuer ) then we forward this JWT to the origin IDP with a JWT bearer to this IDP very similar to what we do with password grant proxy.
The result of this JWT bearer should be processed very similar to password grant proxy.

strehle:
Feature: Forward JWT bearer grant to trusted identity provider. Very similar to password grant.

Refactorings done, so that REST method to call trusted identity provider is re-used (from password grant) .
Reason: JWT bearer and Password grant need similar checks and parameters.

Why:
In a community CF usage the trust maybe is done origin by origin. 
trust-by-origin

SAP uses the corporate origin trust model, so that access to CF is granted by a dedicated IdP which is controlled by SAP. Customers control their own corporate IdP and therefore the authentication requests need to be forward through the corporate Origin to the customer IdP. This model was created with https://github.com/cloudfoundry/uaa/issues/2505
See
trust-by-corporate-origin


server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
```diff
@@ -808,14 +827,150 @@ public KeyInfoService getKeyInfoService() {
...
-    public void fetchMetadataAndUpdateDefinition(OIDCIdentityProviderDefinition definition) {
+    protected void fetchMetadataAndUpdateDefinition(OIDCIdentityProviderDefinition definition) {
         try {
             oidcMetadataFetcher.fetchMetadataAndUpdateDefinition(definition);
         } catch (OidcMetadataFetchingException e) {
             logger.warn("OidcMetadataFetchingException", e);
         }
     }
 
+    public IdentityProvider<OIDCIdentityProviderDefinition> getOidcProxyTokenExchange(HttpServletRequest request) {
```
adrianhoelzl-sap: Can we improve the name of this method? IMO, it indicates that we already perform a token exchange, even though that is not the case.

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
```diff
@@ -808,14 +827,150 @@ public KeyInfoService getKeyInfoService() {
...
+        return retrieveTokenExchangeIdp(UaaLoginHint.parseRequestParameter(request.getParameter("login_hint")), getAllowedProviders());
+    }
+
+    public List<String> getAllowedProviders() {
+        Authentication clientAuth = SecurityContextHolder.getContext().getAuthentication();
+        if (clientAuth == null) {
+            throw new BadCredentialsException("No client authentication found.");
+        }
+        List<String> allowedProviders = null;
```
adrianhoelzl-sap: Please eliminate this variable by directly returning the result or null

server/src/main/java/org/cloudfoundry/identity/uaa/authentication/BackwardsCompatibleTokenEndpointAuthenticationFilter.java
```diff
@@ -254,7 +256,17 @@ protected Authentication attemptTokenAuthentication(HttpServletRequest request,
             log.debug(GRANT_TYPE_JWT_BEARER + " found. Attempting authentication with assertion");
             String assertion = request.getParameter("assertion");
             if (assertion != null && externalOAuthAuthenticationManager != null) {
-                log.debug("Attempting OIDC JWT authentication for token endpoint.");
+                IdentityProvider<OIDCIdentityProviderDefinition> oidcProxy = externalOAuthAuthenticationManager.getOidcProxyTokenExchange(request);
+                if (oidcProxy != null) {
```
adrianhoelzl-sap, line 260: Do I understand the general idea correctly?

We receive a JWT bearer request with a `login_hint` parameter. If the IdP described by that param has `tokenExchangeEnabled`, we forward the incoming JWT to the corporate IdP (also via a JWT bearer grant).

In contrast to that: if the flag is not enabled, we directly try to issue an access token for the user described by the incoming JWT.

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
adrianhoelzl-sap: for example: "getOidcProxyIdpForTokenExchange"

model/src/main/java/org/cloudfoundry/identity/uaa/provider/OIDCIdentityProviderDefinition.java
```diff
@@ -32,6 +32,7 @@ public class OIDCIdentityProviderDefinition extends AbstractExternalOAuthIdentit
     private URL discoveryUrl;
     private boolean passwordGrantEnabled;
     private boolean setForwardHeader;
+    private Boolean tokenExchangeEnabled;
```
adrianhoelzl-sap, line 38: Can you please add a Javadoc comment that describes the purpose of this flag?

server/src/main/java/org/cloudfoundry/identity/uaa/authentication/BackwardsCompatibleTokenEndpointAuthenticationFilter.java
adrianhoelzl-sap: The "regular" JWT bearer flow against UAA can still be used if no `login_hint` parameter is passed in the request.

server/src/main/java/org/cloudfoundry/identity/uaa/authentication/BackwardsCompatibleTokenEndpointAuthenticationFilter.java
```diff
@@ -254,7 +256,17 @@ protected Authentication attemptTokenAuthentication(HttpServletRequest request,
...
+                    log.debug("Forward OIDC JWT authentication to oidc proxy");
+                    String idpAssertion = externalOAuthAuthenticationManager.oidcJwtBearerGrant(
+                            (UaaAuthenticationDetails) authenticationDetailsSource.buildDetails(request),
+                            oidcProxy, assertion);
+                    assertion = idpAssertion != null ? idpAssertion : assertion;
+                } else {
+                    log.debug("Attempting OIDC JWT authentication for token endpoint.");
+                }
+
                 ExternalOAuthCodeToken token = new ExternalOAuthCodeToken(null, null, null, assertion, null, null);
```
adrianhoelzl-sap, line 270: Would it make sense to introduce a static factory method here? e.g. `ExternalOAuthCodeToken.forAssertion(...)`

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
```diff
@@ -808,14 +827,150 @@ public KeyInfoService getKeyInfoService() {
...
         } catch (OidcMetadataFetchingException e) {
             logger.warn("OidcMetadataFetchingException", e);
         }
     }
 
+    public IdentityProvider<OIDCIdentityProviderDefinition> getOidcProxyTokenExchange(HttpServletRequest request) {
+        return retrieveTokenExchangeIdp(UaaLoginHint.parseRequestParameter(request.getParameter("login_hint")), getAllowedProviders());
+    }
+
+    public List<String> getAllowedProviders() {
```
adrianhoelzl-sap, line 842: Can we make this method `static`?

server/src/main/java/org/cloudfoundry/identity/uaa/authentication/BackwardsCompatibleTokenEndpointAuthenticationFilter.java
strehle: yeah we can do, it not really related to this change, because we dont have that many factory methods in UAA...

server/src/main/java/org/cloudfoundry/identity/uaa/authentication/BackwardsCompatibleTokenEndpointAuthenticationFilter.java
strehle: Yes, login_hint is an optional parameter, very similar to password grant.
The only difference to password grant is.

If the default IdP is OIDC, then we dont proxy the call. The password grant handler is doing that but we run this behavior now some year. In JWT bearer logic I only proxy the assertion if login_hint is available as parameter.
With this it is ensured, that no change to version is done. 
So proxy is only done if you request it via login_hint, even if default IdP is OIDC but even then you must add login_hint with value of defaultIdP

fhanik (approved):

PR #3316, opened 2025-02-27
Move login-ui.xml to java config, second part

Kehrlann:
Follow up of #3262 

## Context

Modernizing UAA and moving XML config to java config. This address the `login-ui.xml` entirely.

### Resulting filter chains

Reduced the number of filter chains from 12 to 5. This was done in multiple steps - first moved all existing filter chains to java, then started merging what can be merged.

This required touching various filters, tests and components so the behavior is clearer. When in doubt of why e.g. a test was removed, look at the specific "Fold XYZ filter chain into ..." commit. The commit message has explanations.

There are probably a few other merges that are possible, but doing so would require more work.

### Properties

Moved some properties into proper properties file, so that they are easier to find than in SpEL expressions. This allows to standardize the default values. See for example `${require_https}` or `${servlet.session-cookie.max-age}`.

### ResetPasswordAuthenticationFilter

Now has a `matcher` so it is only executed on certain requests, here `POST /reset_password.do`. This allows to put this filter in a more general filter chain and avoid it intercepting other requests.

This should technically be a controller endpoint rather than a Filter, but that's a lot more work.

### LocalUaaRestTemplate

This is a leftover from the "notify user when account is created / password is changed / etc". When it is done over HTTP, it uses a `LocalUaaRestTemplate`. It's only used in this particular case. To avoid dependency issues, this now only imported in the `NotificationConfiguration` when some beans require it. Otherwise it's not loaded in the application context.

### SamlAuthenticationFilterConfig

This was only touched because some beans definitions made more sense in `LoginSecurityConfiguration` where they are actually used.

I cleaned up unused `@Autowired` annotations.

### Sonar

Any class called `*SecurityConfiguration.java` is now excluded from Sonar, because it mostly contains `@Bean UaaFilterChains`, which don't require code coverage. Sonar gets its code coverage wrong on beans.


server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginSecurityConfiguration.java
```diff
@@ -38,6 +39,20 @@ ResourcePropertySource messagePropertiesSource() throws IOException {
         return new ResourcePropertySource("messages.properties");
     }
 
+    @Bean
+    @Order(FilterChainOrder.INVITATIONS_ACCEPT)
+    UaaFilterChain acceptInvitation(HttpSecurity http) throws Exception {
+        var originalChain = http
+                .securityMatcher("/invitations/accept")
+                .authorizeHttpRequests(auth -> auth.anyRequest().permitAll())
```
fhanik: **NitPick**

Isn't 
```
auth -> auth.anyRequest().permitAll()
```
different from 
```
IS_AUTHENTICATED_ANONYMOUSLY
```

Wouldn't the latter be represented by `.anonymous()` ?

I marked it nitpick since we don't have tests or code that distinguishes from the two.

server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginSecurityConfiguration.java
```diff
@@ -42,6 +40,23 @@ ResourcePropertySource messagePropertiesSource() throws IOException {
...
+    UaaFilterChain forgotPassword(HttpSecurity http) throws Exception {
+        var originalChain = http
+                .securityMatcher(
+                        "/forgot_password",
+                        "/forgot_password.do"
+                )
+                .authorizeHttpRequests(auth -> auth.anyRequest().permitAll())
+                .csrf(CsrfConfigurer::disable)
+                .exceptionHandling(exception -> {
+                    exception.authenticationEntryPoint(new CsrfAwareEntryPointAndDeniedHandler("/invalid_request", "/login?error=invalid_login_request"));
```
fhanik:
```
new CsrfAwareEntryPointAndDeniedHandler("
```

If we're using this construct in many places, why not make it a bean like it was in XML?

server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginSecurityConfiguration.java
```diff
@@ -44,6 +52,37 @@ ResourcePropertySource messagePropertiesSource() throws IOException {
...
+    UaaFilterChain inviteUser(
+            HttpSecurity http,
+            @Qualifier("resourceAgnosticAuthenticationFilter") OAuth2AuthenticationProcessingFilter oauth2ResourceFilter
+    ) throws Exception {
+        var originalChain = http
+                .securityMatcher("/invite_users/**")
+                .authorizeHttpRequests(auth -> {
+                    auth.requestMatchers(HttpMethod.POST, "/**").access(
+                            (authSup, ctx) -> {
+                                // TODO: dgarnier move to AuthorizationManagersUtils
```
fhanik: I'm realizing that by reviewing `by commit` - I'm looking at code no longer present in the PR.

server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginSecurityConfiguration.java
```diff
@@ -62,6 +64,10 @@ class LoginSecurityConfiguration {
 
     private static final CsrfAwareEntryPointAndDeniedHandler LOGIN_ENTRYPOINT = new CsrfAwareEntryPointAndDeniedHandler("/invalid_request", "/login?error=invalid_login_request");
     private static final CsrfAwareEntryPointAndDeniedHandler ACCESS_DENIED_HANDLER = LOGIN_ENTRYPOINT;
+    private static final Customizer<ExceptionHandlingConfigurer<HttpSecurity>> EXCEPTION_HANDLING = exceptionHandling -> {
+        exceptionHandling.accessDeniedHandler(ACCESS_DENIED_HANDLER);
+        exceptionHandling.authenticationEntryPoint(LOGIN_ENTRYPOINT);
+    };
```
fhanik, line 70: We briefly chatted about this on March 6. 

This is a deviation from the XML, as in it makes hard coded constants instead of beans. How do we decide when to do this, and by hard coding constants, prevent a future user from overriding the exception handler by simply providing their own bean?

fhanik (approved):

duanemay (approved):

server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginSecurityConfiguration.java
Kehrlann: Excellent question - it did puzzle me, but after investigating it, it's strictly equivalent. 

For context, I confirmed with Rob from Spring Security when doing the migration, his answer was

> I don't really understand the idea of being authenticated anonymously. I think that anonymous was historically a thing for hiding a logout button if the user was "authenticated anonymously"
> [...]
> IS_AUTHENTICATED_ANONYMOUSLY maps to anonymous || fullyAuthenticated (authenticated but not remember me) || remember me. This can be reduced to anonymous or not anonymous which is permitAll
> [...]
> Again...lots of things that need cleaned up but just take a long time to migrate off of due to history

server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginSecurityConfiguration.java
Kehrlann: As a rule of thumb, I prefer constants over IoC managed beans. Unless there is a specific reason to make something a bean, I try to steer clear from it.

If it:
1. Does not depend on other beans, configuration or the "environment"
2. Is not required to be exposed as a bean to be picked up by Boot or other libs
3. Has a single instance, and a single type, that is used everywhere
4. Is not overloaded in "component" tests (`@DefaultTestContext`)

Then I would make a constant. Advantages:
- No need to be discovered by the Spring container
- Simplifies the bean graph
    - As you mentioned, we are less likely to introduce overrides or collisions
    - We are less likely to introduce a dependency path with side-effects. For example, initializing a bean "too early" [excludes some beans from post-processing](https://docs.spring.io/spring-framework/reference/core/beans/factory-extension.html#beans-factory-extension-bpp) which has bitten in the UAA (there are currently 158 such beans in a MockMvcTest, untangling it is a pain).

I find it easier for my mental model when reading the code too: when I see `foo.setThing(THING)` in a bean function, I read "use the THING constant", whereas when I see `foo.setThing(thing)` I have to jump through "ok `thing` is a parameter, injected by type - which beans do I have of this type ; which beans do I have that _implement_ this type? is it overridden anywhere that I should be aware of?"

PR #3329, opened 2025-03-10
replace Spring Security filters in scim-endpoints.xml

fhanik:


server/src/main/java/org/cloudfoundry/identity/uaa/scim/beans/ScimSecurityConfiguration.java
```diff
@@ -0,0 +1,198 @@
...
+
+    @Autowired
+    @Qualifier("oauthAccessDeniedHandler")
+    OAuth2AccessDeniedHandler oauthAccessDeniedHandler;
+
+    @Autowired
+    CookieBasedCsrfTokenRepository csrfTokenRepository;
+
+    @Bean
+    @Order(FilterChainOrder.SCIM)
```
Kehrlann: The filter chains should not have the same order in general

**[issue]**  In particular here, `scimUserPassword` and `scimUsers` conflict ; if for some reason `scimUsers` gets picked up first, it will swallow all calls to `/Users/...`

server/src/main/java/org/cloudfoundry/identity/uaa/scim/beans/ScimSecurityConfiguration.java
```diff
@@ -0,0 +1,198 @@
...
+    UaaFilterChain scimUserPassword(HttpSecurity http) throws Exception {
+        SecurityFilterChain chain = http
+                .securityMatcher("/User*/*/password", "/User*/*/password/**")
+                .authorizeHttpRequests( auth -> {
+                    auth.requestMatchers("/**").access(anyOf(true).hasScopeWithZoneId("password.write"));
+                    auth.anyRequest().denyAll();
+                })
+                .authenticationManager(emptyAuthenticationManager)
```
Kehrlann: **[nitpick, follow-up]** I found that `emptyAuthenticationManager` may not be required, dependening on the FC. Consider removing entirely.

server/src/main/java/org/cloudfoundry/identity/uaa/scim/beans/ScimSecurityConfiguration.java
```diff
@@ -0,0 +1,198 @@
...
+    OAuth2AccessDeniedHandler oauthAccessDeniedHandler;
+
+    @Autowired
+    CookieBasedCsrfTokenRepository csrfTokenRepository;
+
+    @Bean
+    @Order(FilterChainOrder.SCIM)
+    UaaFilterChain scimUserPassword(HttpSecurity http) throws Exception {
+        SecurityFilterChain chain = http
+                .securityMatcher("/User*/*/password", "/User*/*/password/**")
```
Kehrlann: **[nitpick, follow-up]** AFAICT there are no `/User*` endpoints, only `/Users` ; align the matchers on `/Users`

server/src/main/java/org/cloudfoundry/identity/uaa/scim/beans/ScimSecurityConfiguration.java
```diff
@@ -0,0 +1,198 @@
...
+import org.springframework.security.web.authentication.www.BasicAuthenticationFilter;
+
+import static org.cloudfoundry.identity.uaa.web.AuthorizationManagersUtils.anyOf;
+
+@Configuration
+@EnableWebSecurity
+class ScimSecurityConfiguration {
+
+    @Autowired
+    @Qualifier("tokenServices")
```
Kehrlann, line 34: **[nit, follow-up]** I think there's only one instance of this, called `tokenServices`, remove the qualifier?

(There might be a bug in the existing code where the `tokenServices` is defined both in XML and in `@Component`)

server/src/main/java/org/cloudfoundry/identity/uaa/security/SelfCheckAuthorizationManager.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.util.function.Supplier;
+import javax.servlet.http.HttpServletRequest;
+
+import org.springframework.security.authorization.AuthorizationDecision;
+import org.springframework.security.authorization.AuthorizationManager;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.web.access.intercept.RequestAuthorizationContext;
+
+public class SelfCheckAuthorizationManager  implements AuthorizationManager<RequestAuthorizationContext> {
```
Kehrlann: **[nit]** Should probably be in the same package as `AuthorizationManagerUtils`

server/src/main/java/org/cloudfoundry/identity/uaa/scim/beans/ScimSecurityConfiguration.java
```diff
@@ -0,0 +1,198 @@
...
+                                .accessDeniedHandler(oauthAccessDeniedHandler)
+                )
+                .build();
+
+        return new UaaFilterChain(chain, "groupEndpointSecurity");
+    }
+
+    @Bean
+    @Order(FilterChainOrder.SCIM)
+    UaaFilterChain scimUsers(HttpSecurity http, @Qualifier("self") IsSelfCheck selfCheck) throws Exception {
```
Kehrlann, line 125: **[nit, follow-up]** probably don't need the qualifier

server/src/main/java/org/cloudfoundry/identity/uaa/web/AuthorizationManagersUtils.java
```diff
@@ -24,21 +29,45 @@ public static AnyOfAuthorizationManager anyOf() {
         return new AnyOfAuthorizationManager();
     }
 
+    public static AnyOfAuthorizationManager anyOf(boolean throwOnError) {
```
Kehrlann: **[nit, follow-up]** it used to be `throwOnError` but scoped to `oauth2` ; now it's just `throwOnError` and the meaning is less clear. Consider renaming to `throwOnMissingScope` ; or at least explaining in the javadoc where this comes from?

server/src/main/java/org/cloudfoundry/identity/uaa/web/AuthorizationManagersUtils.java
```diff
@@ -96,10 +125,24 @@ public AnyOfAuthorizationManager hasScopeWithZoneId(String scope) {
...
                         var securityMethods = new ContextSensitiveOAuth2SecurityExpressionMethods(auth.get());
-                        return new AuthorizationDecision(securityMethods.hasScopeInAuthZone(scope));
+                        return new ScopeTrackingAuthorizationDecision(securityMethods.hasScopeInAuthZone(scope), scope);
                     }
             );
             return this;
         }
     }
+
+    private static class ScopeTrackingAuthorizationDecision extends AuthorizationDecision {
```
Kehrlann, line 135: **[praise]** Clever encapsulation :)

server/src/main/java/org/cloudfoundry/identity/uaa/web/FilterChainOrder.java
```diff
@@ -6,13 +6,13 @@
...
-    // Order of filters handling user login features, formerly defined by
-    // ordering filter chains in login-ui.xml
+    // Order of filters in login-ui.xml
     public static final int AUTOLOGIN_CODE = 1200;
     public static final int AUTOLOGIN = 1201;
     public static final int INVITATIONS = 1202;
     public static final int INVITE = 1203;
     public static final int LOGIN_PUBLIC_OPERATIONS = 1204;
-    public static final int UI_SECURITY = 1205;
+    public static final int SCIM = 1205;
```
Kehrlann: **[issue]** This is different from the previous order ; all the SCIM filter chains were registered before anything in `login-ui.xml` ; although it may have no impact. If we're trying to replicate the exact behavior, then we should bump SCIM above any of the other FCs.

The ordering here is not explained (my bad), but I had in mind:
ORDER = 100 * (file position in spring-servlet.xml) + (http position within that file)

So here SCIM_USER_PASSWORDS would be 400, SCIM_USER_IDS 401, etc.

server/src/main/java/org/cloudfoundry/identity/uaa/web/UaaFilterChain.java
```diff
@@ -20,9 +20,15 @@
 public class UaaFilterChain implements SecurityFilterChain {
 
     private final SecurityFilterChain chain;
+    private final String name;
 
-    public UaaFilterChain(SecurityFilterChain chain) {
+    public UaaFilterChain(SecurityFilterChain chain, String name) {
```
Kehrlann, line 25: **[praise]** so useful in debugging ❤️

Kehrlann (commented):
Two issues re: filter chain ordering, one which is important in `ScimSecurityConfiguration`, and one less so in `FilterChainOrder` but the behavior diverges from the old implementation.

I marked a few things as **nit, follow-up** : consider doing it, either here in another commit, or in a follow-up PR. They're non-structural, but they allow use to make simpler configuration (less things), which I find easier to understand. Feel free to disregard though.

Other things:
- **[nit, follow-up]** All filter chains seem to have the same rules, except the `authorizeHttpRequests`. I would fold them all into a single chain.
- **[nit]** There is an empty `/remove.me`file, is that intended?

server/src/main/java/org/cloudfoundry/identity/uaa/scim/beans/ScimSecurityConfiguration.java
fhanik: Good point. I do need all chains to have a order, so it ends up before the swallow all UI. In between most of them, the order doesn't mean anything with the exception that you mention. 

I will fix this.

server/src/main/java/org/cloudfoundry/identity/uaa/scim/beans/ScimSecurityConfiguration.java
fhanik: This appears to hold true. I will remove the references.

server/src/main/java/org/cloudfoundry/identity/uaa/scim/beans/ScimSecurityConfiguration.java
fhanik: You are right. I had just copied the XML

server/src/main/java/org/cloudfoundry/identity/uaa/scim/beans/ScimSecurityConfiguration.java
fhanik: I'll leave this for now. We can simplify later.

server/src/main/java/org/cloudfoundry/identity/uaa/security/SelfCheckAuthorizationManager.java
fhanik: Class has been moved

server/src/main/java/org/cloudfoundry/identity/uaa/scim/beans/ScimSecurityConfiguration.java
fhanik: I'll leave qualifiers in place for now. We can do a larger refactor later when all XML is gone.

server/src/main/java/org/cloudfoundry/identity/uaa/web/AuthorizationManagersUtils.java
fhanik: good point. I have renamed the method.

server/src/main/java/org/cloudfoundry/identity/uaa/web/FilterChainOrder.java
fhanik: I have put the SCIM filters in 400 class. we don't need a unique order for each, we just need to ensure that the password change goes before the rest, and that all of them go for UI catch all

PR #3330, opened 2025-03-10
Integration tests: add an option for the web driver to click and wait for a page reload

Kehrlann:


strehle (approved):
you can go with yours

PR #3337, opened 2025-03-11
Move codestore-endpints endpoints javaconfig

Kehrlann:
Do not merge, waiting on gh-3329


server/src/main/java/org/cloudfoundry/identity/uaa/codestore/CodeStoreSecurityConfiguration.java
```diff
@@ -0,0 +1,52 @@
...
+class CodeStoreSecurityConfiguration {
+
+    @Bean
+    @Order(FilterChainOrder.CODESTORE)
+    UaaFilterChain codestore(
+            HttpSecurity http,
+            @Qualifier("resourceAgnosticAuthenticationFilter") OAuth2AuthenticationProcessingFilter oauth2ResourceFilter
+    ) throws Exception {
+        var originalFilterChain = http
+                .securityMatcher("/Codes/**")
```
fhanik, line 31: I've noticed a difference between XML and Java, where in XML

```
"/Codes/**"
```
matches both `/Codes` and `/Codes/`

I vaguely remember that `/Codes` would not match in Java.

I'm RTO, so limited equipment for a deep dive, but should this be:

```
.securityMatcher("/Codes", "/Codes/**")
```

asking for a friend!

fhanik (approved):

server/src/main/java/org/cloudfoundry/identity/uaa/codestore/CodeStoreSecurityConfiguration.java
Kehrlann: Good point. I just tested by running the UAA, and `/Codes/**` actually matches requests to `/Codes`, `/Codes/` and `/Codes/foobar`.

Validated with a simple test:

```java
var request = MockMvcRequestBuilders.get("/Codes").buildRequest(new MockServletContext())
new MvcRequestMatcher(new HandlerMappingIntrospector(), "/Codes/**")
  .matches(request); // -> returns true
```

Kehrlann: Found it!

The changes you are mentioning are introduced for Spring Framework 6.x, and are about trailing slashes ([spring-framework docs](https://docs.spring.io/spring-framework/reference/web/webmvc/filters.html#filters.url-handler)).

fhanik: Good find!

PR #3338, opened 2025-03-11
Remove the rate limiter filter from the resource-endpoints.xml

fhanik:


server/src/main/java/org/cloudfoundry/identity/uaa/ratelimiting/beans/RateLimiterSecurityConfiguration.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Autowired
+    @Qualifier("clientAuthenticationFilter")
+    ClientBasicAuthenticationFilter clientAuthenticationFilter;
+
+    @Bean
+    @Order(FilterChainOrder.RATE_LIMIT)
+    UaaFilterChain ratelimitSecurity(HttpSecurity http) throws Exception {
+        SecurityFilterChain chain = http
+                .securityMatcher("/RateLimitingStatus/**", "/RateLimitingStatus" )
```
Kehrlann: **[nit]** As discussed in gh-3337, this is equivalent to `.securityMatcher("/RateLimitingStatus/**")`

Kehrlann (approved):
👍 thanks!

PR #3343, opened 2025-03-12
resource-endpoints.xml to java config

fhanik:


server/src/main/java/org/cloudfoundry/identity/uaa/client/beans/ClientInfoSecurityConfiguration.java
```diff
@@ -0,0 +1,63 @@
...
+    UaaFilterChain clientInfoSecurity(HttpSecurity http) throws Exception {
+        SecurityFilterChain chain = http
+                .securityMatcher("/clientinfo/**")
+                .authorizeHttpRequests( auth -> {
+                    auth.requestMatchers("/**").access(anyOf().fullyAuthenticated());
+                    auth.anyRequest().denyAll();
+                })
+                .sessionManagement(session -> session.sessionCreationPolicy(SessionCreationPolicy.STATELESS))
+                //TODO is the auth manager needed?
+                .authenticationManager(clientAuthenticationManager)
```
Kehrlann, line 52: **[nit]** as you mentioned, not needed because it's already injected in the filter at bean creation time

Kehrlann (approved):
LGTM, one minor nit on the authentication manager.

PR #3345, opened 2025-03-12
move approvalsSecurity filterchain to java

fhanik:


server/src/main/java/org/cloudfoundry/identity/uaa/scim/beans/ScimSecurityConfiguration.java
```diff
@@ -34,7 +34,7 @@ class ScimSecurityConfiguration {
     @Qualifier("tokenServices")
     private UaaTokenServices tokenServices;
 
-    @Autowired()
+    @Autowired
```
duanemay, line 37: is Autowired still needed here?

duanemay (approved):

PR #3350, opened 2025-03-13
Move /oauth/clients/** filter chains to java configuration

Kehrlann:


server/src/main/java/org/cloudfoundry/identity/uaa/client/ClientAdminSecurityConfiguration.java
```diff
@@ -0,0 +1,100 @@
...
+     * <li>is Zone admin</li>
+     * <li>has scope clients.admin</li>
+     * <li>has any of the scopes in {@code scopes}</li>
+     * </ul>
+     */
+    public static AuthorizationManager<RequestAuthorizationContext> isAdminOrHasScopes(String... scopes) {
+        String[] requiredScopes = new String[scopes.length + 1];
+        requiredScopes[0] = "clients.admin";
+        System.arraycopy(scopes, 0, requiredScopes, 1, scopes.length);
+        return anyOf()
```
fhanik, line 94: I believe this should be `anyOf(true)`

server/src/main/java/org/cloudfoundry/identity/uaa/client/event/ClientAdminEventPublisher.java
```diff
@@ -9,6 +9,7 @@
 import org.springframework.context.ApplicationEventPublisherAware;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.stereotype.Component;
```
fhanik: orphaned import?

uaa/src/main/webapp/WEB-INF/spring-servlet.xml
```diff
@@ -339,6 +339,11 @@
     </bean>
 
 
+    <bean id="clientDetailsValidator" class="org.cloudfoundry.identity.uaa.client.ClientAdminEndpointsValidator">
```
fhanik, line 342: this has happened to me as well, where the beans are difficult to move. once XML is all gone, maybe we can go over the bean diagram and see if we can simplify it

fhanik (approved):

PR #3351, opened 2025-03-13
Prevent ClassCastException in IdentityZoneSwitcherFilter

strehle:
Check instance for access

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/zone/IdentityZoneSwitchingFilter.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/zone/IdentityZoneSwitchingFilter.java
@@ -87,7 +87,11 @@ public class IdentityZoneSwitchingFilter extends OncePerRequestFilter {
                     new UaaAuthenticationDetails(servletRequest),
                     true, userAuthentication.getAuthenticatedTime());
         }
-        oa = new UaaOauth2Authentication(((UaaOauth2Authentication) oa).getTokenValue(), IdentityZoneHolder.get().getId(), request, userAuthentication);
+        String tokenValue = null;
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/cloudfoundry/identity/uaa/zone/IdentityZoneSwitchingFilter.java
```diff
@@ -87,7 +87,11 @@ protected OAuth2Authentication getAuthenticationForZone(String identityZoneId, H
                     new UaaAuthenticationDetails(servletRequest),
                     true, userAuthentication.getAuthenticatedTime());
         }
-        oa = new UaaOauth2Authentication(((UaaOauth2Authentication) oa).getTokenValue(), IdentityZoneHolder.get().getId(), request, userAuthentication);
+        String tokenValue = null;
+        if (oa instanceof UaaOauth2Authentication uaaOauth2Authentication) {
+            tokenValue = uaaOauth2Authentication.getTokenValue();
+        }
+        oa = new UaaOauth2Authentication(tokenValue, IdentityZoneHolder.get().getId(), request, userAuthentication);
         oa.setDetails(oaDetails);
```
fhanik, line 95: @strehle Looking at the request causing the error, it should either return `400 Bad Request` or `401 Forbidden` for API clarity. 

The request in the `curl` command simply shouldn't be allowed. Albeit, the example does _not_ switch the actual zone, the use of the header is invalid.

Can I submit another PR for you to review?

strehle: Sure you can take the branch and fix it, since you are the author of the filter.
Everything different to 500 is ok to me

strehle:
close because of alternatives, e.g. https://github.com/cloudfoundry/uaa/pull/3401

PR #3352, opened 2025-03-14
Fix #3349 - IdentitySwitchFilter 500 Error

### LINKED ISSUE #3349: Token Endpoint with HTTP 500 response
The identity zone switching filter causes HTTP 500.

Example token request with curl

`curl 'http://localhost:8080/uaa/oauth/token' -i -X POST -H 'Content-Type: application/x-www-form-urlencoded' -H 'Accept: application/json' -H 'X-Identity-Zone-Id: uaa' -d 'client_id=cf&client_secret=&grant_type=password&username=marissa&password=koala&token_format=jwt'
`

Reason: class cast exception in
https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/zone/IdentityZoneSwitchingFilter.java#L90

fhanik:
IdentityZoneSwitchFilter - You cannot switch to the default zone, instead simply continue.

Turns out that the zone switch headers are abused, and that a switch from `zone=uaa` to `zone=uaa` translating the permission `zone.uaa.admin` to `uaa.admin` is quite frequent in our test suite

This will take some thinking about....

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/zone/IdentityZoneSwitchingFilter.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/zone/IdentityZoneSwitchingFilter.java
@@ -19,12 +19,14 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
 import java.util.Collection;
+import java.util.Collections;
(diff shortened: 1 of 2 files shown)


fhanik:
Not a workable solution....too many tests are switching to UAA zone (by mistake)

PR #3355, opened 2025-03-18
Move multitenant-endpoints.xml to javaconfig

Kehrlann:


server/src/main/java/org/cloudfoundry/identity/uaa/zone/beans/IdentityZoneSecurityConfiguration.java
```diff
@@ -22,11 +21,94 @@
...
+                .securityMatcher("/identity-zones/**")
+                .authenticationManager(emptyAuthenticationManager)
+                .authorizeHttpRequests(auth -> {
+                    auth.requestMatchers(HttpMethod.GET, "/identity-zones").access(
+                            anyOf()
+                                    .isUaaAdmin()
+                                    .isZoneAdmin()
+                                    .hasScopeWithZoneId("zones.read")
+                                    .hasScope("zones.write")
+                                    .throwOnMissingScope()
```
fhanik, line 56: [non blocking for merge]

Do we want to have to call `throwOnMissingScope? Or do we simply want to assume it to be `true`

fhanik (approved):

PR #3361, opened 2025-03-21
fix saml validation and allow again only encrypted assertions

### LINKED ISSUE #3353: Ignores setting `config.samlConfig.wantAssertionSigned` when validating SAML response
We need to configure an Identity Zone with a legacy identity provider that returns SAML responses that are encrypted but not signed. Until UAA v77.18 login with this SAML-Provider worked as expected, but ever since updating to v77.22 or newer we keep getting error messages `did not decrypt response (...) since it was not signed`. 

We figured that the error is thrown by [this part](https://github.com/cloudfoundry/uaa/blob/11fa350f4f0d080628a3a5f1e18545612a30001b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/OpenSaml4AuthenticationProvider.java#L367) of the code. There is a configuration option `config.samlConfig.wantAssertionSigned` which is documented for the identity-zone. When setting `config.samlConfig.wantAssertionSigned: false` I would expect UAA to ignore the SAML response missing a signature. However, the configuration has no effect on the actual behavior and we keep getting error messages. The code block appears to ignore this identity-zone setting

strehle:
Tested side effect: In former UAAs the option

`wantAssertionSigned: false`

was too dangerous because you could end up in a validation without any crypto part.

Now you can set in UAA default zone wantAssertionSigned: false, because there is an extra check of situations without crypto part. But if an assertion is encrypted and if we can read it, then we don't need the signature, because the encrypted is good enough (or even better than signature only).

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/OpenSaml4AuthenticationProvider.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/OpenSaml4AuthenticationProvider.java
@@ -20,6 +20,8 @@ import lombok.Getter;
 import net.shibboleth.utilities.java.support.xml.ParserPool;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.cloudfoundry.identity.uaa.zone.IdentityZone;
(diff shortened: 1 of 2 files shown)


duanemay (approved):
Thank you

PR #3377, opened 2025-03-30
fix federated credential administration

strehle:
fix administration of federated credentials, e.g. example client client_federated_jwt_trust

uaac client get client_federated_jwt_trust
 -> one entry
uaac client jwt add client_federated_jwt_trust --issuer http://localhost:8080/uaa/oauth/token --subject admin
uaac client jwt delete client_federated_jwt_trust --issuer http://localhost:8080/uaa/oauth/token --subject admin
uaac client get client_federated_jwt_trust
 -> one entry

### DIFF
--- a/model/src/main/java/org/cloudfoundry/identity/uaa/oauth/client/ClientJwtChangeRequest.java
+++ b/model/src/main/java/org/cloudfoundry/identity/uaa/oauth/client/ClientJwtChangeRequest.java
@@ -7,7 +7,6 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 
 import static org.cloudfoundry.identity.uaa.oauth.client.ClientJwtChangeRequest.ChangeMode.ADD;
 import static org.cloudfoundry.identity.uaa.oauth.client.ClientJwtChangeRequest.ChangeMode.DELETE;
-import static org.cloudfoundry.identity.uaa.oauth.client.ClientJwtChangeRequest.ChangeMode.UPDATE;
 
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @JsonIgnoreProperties(ignoreUnknown = true)
@@ -124,8 +123,8 @@ public class ClientJwtChangeRequest {
 
     @JsonIgnore
     public boolean isFederated() {
-        return ((changeMode == ADD || changeMode == UPDATE) && issuer != null && subject != null) ||
-                (changeMode == DELETE && (issuer != null && subject != null));
+        // is private_key_jwt according to RFC 7523. audience is addition supported, but optional
+        return (issuer != null && subject != null);
     }
 
     @JsonIgnore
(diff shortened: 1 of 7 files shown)


model/src/main/java/org/cloudfoundry/identity/uaa/oauth/client/ClientJwtCredential.java
```diff
@@ -45,4 +45,21 @@ public static List<ClientJwtCredential> parse(String clientJwtCredentials) {
...
     }
 
+    @Override
+    public boolean equals(Object object) {
+        if (this == object) return true;
+        if (object == null || getClass() != object.getClass()) return false;
+        ClientJwtCredential that = (ClientJwtCredential) object;
+        return subject.equals(that.subject) &&
+               issuer.equals(that.issuer) &&
+               (audience != null ? audience.equals(that.audience) : that.audience == null);
```
adrianhoelzl-sap:
```suggestion
               Objects.equals(audience, that.audience);
```

model/src/main/java/org/cloudfoundry/identity/uaa/oauth/client/ClientJwtChangeRequest.java
```diff
@@ -125,7 +125,7 @@ public String getChangeValue() {
     @JsonIgnore
     public boolean isFederated() {
         return ((changeMode == ADD || changeMode == UPDATE) && issuer != null && subject != null) ||
-                (changeMode == DELETE && (issuer != null || subject != null));
+                (changeMode == DELETE && (issuer != null && subject != null));
```
adrianhoelzl-sap: Can we simplify the whole expression to `return issuer != null && subject != null;`?

model/src/main/java/org/cloudfoundry/identity/uaa/oauth/client/ClientJwtChangeRequest.java
```diff
@@ -125,7 +125,7 @@ public String getChangeValue() {
     @JsonIgnore
     public boolean isFederated() {
```
adrianhoelzl-sap, line 125: Can we add a comment here that this corresponds to the RFC 7523 approach?

model/src/main/java/org/cloudfoundry/identity/uaa/oauth/client/ClientJwtChangeRequest.java
```diff
@@ -125,7 +125,7 @@ public String getChangeValue() {
...
         return ((changeMode == ADD || changeMode == UPDATE) && issuer != null && subject != null) ||
-                (changeMode == DELETE && (issuer != null || subject != null));
+                (changeMode == DELETE && (issuer != null && subject != null));
     }
 
     @JsonIgnore
```
adrianhoelzl-sap, line 130: Can we add a comment here that no explicit JWKS URI is required as the issuer is already a URL (that can be used to look up the token keys)?

model/src/main/java/org/cloudfoundry/identity/uaa/oauth/client/ClientJwtCredential.java
strehle: was done by github co pilot

adrianhoelzl-sap (approved):

PR #3380, opened 2025-04-01
fix class cast issue

strehle:
@fhanik found in a landscape after activating the new release

### DIFF
--- a/scripts/cargo/uaa.yml
+++ b/scripts/cargo/uaa.yml
@@ -51,7 +51,10 @@ jwt:
       format: opaque
       rotate: false
       unique: false
-
+    claims:
+      exclude:
+        - excluded-claim1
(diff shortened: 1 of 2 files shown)


scripts/cargo/uaa.yml
```diff
@@ -51,7 +51,10 @@ jwt:
       format: opaque
       rotate: false
       unique: false
-
+    claims:
```
strehle, line 54: only an example but with the example we ensure that there is no error

strehle:
@fhanik @Kehrlann want do a release today or tomorrow and this should be in , because it is a regression - if someone has exclude claims configured, therefore can you please check if ok for you

Kehrlann (approved):
👍

PR #3389, opened 2025-04-07
Fix #3349

### LINKED ISSUE #3349: Token Endpoint with HTTP 500 response

fhanik:
If an attempt is made to switch to the default UAA(system) zone, the header is removed and ignored

https://github.com/cloudfoundry/uaa/issues/3349


strehle (approved):
thanks, please rebase, to get tests green

PR #3393, opened 2025-04-10
fix dev tests by pinning chrome to 134

strehle:


strehle:
not working thus close it

PR #3422, opened 2025-04-28
revert single use of RestTemplate in proxy calls

### LINKED ISSUE #3411: NoHttpResponseException appear after version bump from 77.27.0 to 77.29.0

strehle:
Re-establish the beans and usage fromm 77.27.x and earlier

Use rest config as bean 
- use in SAML IDP xml -> https://github.com/cloudfoundry/uaa/blob/76.31.0/uaa/src/main/webapp/WEB-INF/spring/saml-idp.xml#L129
- later use in password grant definition only but obviously created an extra bean -> https://github.com/cloudfoundry/uaa/blob/77.27.0/uaa/src/main/webapp/WEB-INF/spring/oauth-endpoints.xml#L645

It was not intended to do changes with 77.27.-> 77.29.0 but only refactoring was wanted because of re-use of the proxy calls in password grant flow also in jwt bearer flow.
The use of the configuration as extra method was not covered.

Using RestTemplateConfig instead of beans changes behaviour because of
https://docs.spring.io/spring-framework/reference/core/beans/java/basic-concepts.html

This PR make use again of RestTemplateConfig in the proxy REST call, but only there. This was the behaviour in 77.27.0 .


server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
```diff
@@ -146,12 +148,14 @@ public class ExternalOAuthAuthenticationManager extends ExternalLoginAuthenticat
...
+                                              RestTemplateConfig restTemplateConfig,
                                               RestTemplate trustingRestTemplate,
                                               RestTemplate nonTrustingRestTemplate,
                                               TokenEndpointBuilder tokenEndpointBuilder,
                                               KeyInfoService keyInfoService,
                                               OidcMetadataFetcher oidcMetadataFetcher) {
         super(providerProvisioning);
+        this.restTemplateConfig = restTemplateConfig;
         this.trustingRestTemplate = trustingRestTemplate;
         this.nonTrustingRestTemplate = nonTrustingRestTemplate;
```
adrianhoelzl-sap, line 160: Can we also get rid of these two properties and replace the last usage of them with the same pattern?

strehle: If we would do that then we would have again a different behaviour to 77.27.0.

Please check simply to verify my researches.

We used the single rest templates trustingRestTemplate  and nonTrustingRestTemplate  for Authorization Code flow and updating Metadata content. , e.g. .well-known/openid-configuration and/or token_key

For password grant proxy we used restTemplateConfig 

The refactoring with jwt bearer moved coding from passwordgrant manager to this class and that was one reason why I removed restTemplateConfig because I thougt using the restTemplate beans is good enough.

This PR simply changes the usages to what we have had with 77.27.0 ....

We definitly should refactor the use of REST templeates , but to me this is something we should discuss, simply because we have duplicates of trustingRestTemplate  and nonTrustingRestTemplate in SAML config and OAUTH config.

adrianhoelzl-sap (approved):

strehle:
Debugged UAA with version 
76.31.0 -> password grant manager
77.20.x -> password grant manager
77.27.0 -> password grant manager
77.29.0 -> ExternalOAuthAuthenticationManager proxy call

In all version it is the case that (non)TrustingRestTemplate is a single bean with the Pool 
In versions 77.22 -> now we have the problem that nonTrustingRestTemplate and trustingRestTemplate is duplicated and in worse the SAML configuration wins. In SAML there is no Pool configuration and with that the default is used.

This PR would change runtime so that  (non)TrustingRestTemplate is recreated always, but that is now not wanted anymore... thus close PR

PR #3427, opened 2025-04-29
Identity Zone Selection via X-Zid Header

adrianhoelzl-sap:


fhanik:
@adrianhoelzl-sap How does this impact cookies and security?

What If I do a request like this

```
GET /some/path
Host: some.host.com
Header: X-Zid=zone-id-for-someother.host.com
```

and this sets up my cookies (including authentication stored in the session)
and then I do this request (and continue using the same JSESSIONID since the host hasn't changed)
```
GET /some/path
Host: some.host.com
```

Do I just gain access across zones?

adrianhoelzl-sap:
> @adrianhoelzl-sap How does this impact cookies and security?
> 
> What If I do a request like this
> 
> ```
> GET /some/path
> Host: some.host.com
> Header: X-Zid=zone-id-for-someother.host.com
> ```
> 
> and this sets up my cookies (including authentication stored in the session) and then I do this request (and continue using the same JSESSIONID since the host hasn't changed)
> 
> ```
> GET /some/path
> Host: some.host.com
> ```
> 
> Do I just gain access across zones?

This scenario should be addressed by this check here: https://github.com/cloudfoundry/uaa/blob/df507238543931915a1bff4873e89f26aba6798e/server/src/main/java/org/cloudfoundry/identity/uaa/authentication/SessionResetFilter.java#L60

The zone resolved through the subdomain or `X-Zid` header (`IdentityZoneHolder.getCurrentIdentityZoneId()`) is compared to the zone for which the session was set up (`authentication.getPrincipal().getZoneId()`). If they do not match, the session is invalidated.

fhanik:
The `SessionResetFilter` is added at the end of each filter chain. it is the last filter that is used before an MVC endpoint is invoked.

But all the filters before that, that may contain logic and send redirects prior to the completion of the filter chain are still vulnerable to this. 

A lot of the SAML/OAuth/OIDC happens in just filters, and not in MVC endpoints, and those filters would be working under incorrect assumptions. 

Given the risk, is this header really needed?

If the HTTP client can set a header, it can set the "Host" header, and if there is a browser involved, cookie handling will be accurate.

model/src/main/java/org/cloudfoundry/identity/uaa/oauth/common/util/RandomValueStringGenerator.java
```diff
@@ -15,6 +15,8 @@ public class RandomValueStringGenerator {
 
     private static final char[] DEFAULT_CODEC = "1234567890ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz-_"
             .toCharArray();
+    private static final char[] CODEC_LETTERS_ONLY = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
```
strehle: pls use https://github.com/cloudfoundry/uaa/blob/b449b5b20aa59e2771d6c2cfa03b529afd32b5e6/server/src/main/java/org/cloudfoundry/identity/uaa/util/AlphanumericRandomValueStringGenerator.java#L7

model/src/main/java/org/cloudfoundry/identity/uaa/oauth/common/util/RandomValueStringGenerator.java
```diff
@@ -90,4 +92,7 @@ public void setLength(int length) {
         this.length = length;
     }
 
+    public static RandomValueStringGenerator lettersOnly(final int length) {
```
strehle: see above

strehle (commented):
My hope was that we can consolidate the amount of filter (per each request) and optimize but can we integration this logic into one of other IdentityZoneXYZ filters ? ... maybe rename one and do more ?

model/src/main/java/org/cloudfoundry/identity/uaa/oauth/common/util/RandomValueStringGenerator.java
adrianhoelzl-sap: fixed

model/src/main/java/org/cloudfoundry/identity/uaa/oauth/common/util/RandomValueStringGenerator.java
adrianhoelzl-sap: fixed

scripts/cargo/uaa.yml
```diff
@@ -56,6 +56,7 @@ jwt:
         - excluded-claim1
         - excluded-claim2
 login:
+  zidHeaderEnabled: true
```
strehle, line 59: I checked docs and we have X-Identity-Zone-Id and X-Identity-Zone-Subdomain already , so what is X-zid  in comparison ?

The header allows new possiblities, correct ? then we should have thing to tell admin, e.g.
allowZoneSwitchByHeader or allowZoneSwitchByZidHeader

server/src/main/java/org/cloudfoundry/identity/uaa/zone/IdentityZoneMismatchCheckFilter.java
```diff
@@ -0,0 +1,106 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpSession;
+import java.io.IOException;
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * Checks whether there is a mismatch between ...
+ * <ul>
```
strehle, line 25: What is the difference to X-Identity-Zone-Id, X-Identity-Zone-Subdomain ?

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/zones/IdentityZoneResolvingMockMvcTest.java
```diff
@@ -99,4 +115,199 @@ void isNotFound(String hostname) throws Exception {
...
+                                .header("X-Zid", zone2Id)
+                                .header("Accept", "application/json")
+                ).andExpect(resolvedZoneWithName(zone2Subdomain));
+            }
+
+            @Test
+            void subdomainSetToZone1_ZidHeaderSetToZone2_BothZonesExist_ShouldReturnZone2() throws Exception {
+                mockMvc.perform(
+                        get("/login")
+                                .header("Host", zone1Subdomain + "." + HOST_NO_SUBDOMAIN)
```
strehle, line 171: this means, even if we have a subdomain but if there is a X-Zid we overrule the subdomain ?

adrianhoelzl-sap: Yes, the `X-Zid` header overrides the zone specified by the subdomain.

server/src/main/java/org/cloudfoundry/identity/uaa/zone/IdentityZoneMismatchCheckFilter.java
adrianhoelzl-sap: With the subdomain or the `X-Zid` header, one selects the identity zone to log in to. With the other two headers, IdZ switching can be performed (only possible if logged in to "uaa" zone).

Example for IdZ switching:

- user has a group `zones.custom.scim.read` in the "uaa" zone
- log in to "uaa" zone (here, the `X-Zid` can be used as an override), receive token
- use the token for performing actions according to the `scim.read` scope in the zone with the ID "custom"
  - In the SCIM requests, he would then need to pass the `X-Identity-Zone-Id` header with the value "custom"

PR #3428, opened 2025-04-29
fix SAML IdP initiated SSO

### LINKED ISSUE #3423: SAML IdP-Initiated SSO fails when using entity ID alias
The UaaRelyingPartyRegistrationResolver is used to resolve the relying party from the request such as with IdP-initiated SSO flow. The request comes in on the assertion consumer service URL. This works when the entity ID is in the path, but when using an entity ID alias, they don't match and the entity ID cannot be resolved. This results in a "No relying party registration found" situation. This is because it expects the entity ID in the path and not an alias.

An entity ID alias such as "cloudfoundry-saml-login" may be used if the entity ID is of a URI form like "https://cloudfoundry-saml-login" since the alias results in a cleaner ACS URL.

strehle:
Use both entityIdAlias and/or entityID, similar to bean creation relyingPartyRegistrationRepository, -> https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/SamlRelyingPartyRegistrationRepositoryConfig.java#L53

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/SamlRelyingPartyRegistrationRepositoryConfig.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/SamlRelyingPartyRegistrationRepositoryConfig.java
@@ -16,6 +16,7 @@ import org.springframework.security.saml2.provider.service.registration.RelyingP
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


strehle:
@duanemay any concerns if I merge ?

duanemay (approved):

PR #3430, opened 2025-04-30
Fix duplicate bean definition of rest templates

### LINKED ISSUE #3411: NoHttpResponseException appear after version bump from 77.27.0 to 77.29.0

strehle:
before we had duplicates of:
- nonTrustingRestTemplate
- trustingRestTemplate

With this there is a separation of RestTemplate usage between SAML and OIDC/OAUTH
Main reason is, that there are SAML timesouts and other time outs.
With opensaml-2 and spring-saml2 there was obviously the same separate usages


server/src/main/java/org/cloudfoundry/identity/uaa/oauth/beans/OauthEndpointBeanConfiguration.java
```diff
@@ -550,11 +551,17 @@ CurrentUserCookieRequestFilter currentUserCookieFilter(
         return new CurrentUserCookieRequestFilter(currentUserCookieFactory);
     }
 
+    @Bean
+    RestTemplateConfig restTemplateConfig() {
```
fhanik, line 555: Does this not conflict with 
```
@Configuration
public class RestTemplateConfig
```

Should this be a `@Configuration` or a `@Bean` but maybe not both?

fhanik: What happens if you do

```
    @Autowired
    RestTemplateConfig restTemplateConfig
```

strehle: If I remove the extra Bean config, then the singleton is used, means, 

`RestTemplate rt = config.isSkipSslValidation() ? restTemplateConfig.trustingRestTemplate() : restTemplateConfig.nonTrustingRestTemplate();`

is equal to 

`RestTemplate rt = config.isSkipSslValidation() ? trustingRestTemplate : nonTrustingRestTemplate;`

This is what I see in debugger.

I am not sure, when our problems begun, but in past we had this extra bean configuration already, see
76.31.0 -> https://github.com/cloudfoundry/uaa/blob/76.31.0/uaa/src/main/webapp/WEB-INF/spring/saml-idp.xml#L129

Then with 77.22.0 the duplicated beans (trustingRestTemplate  / nonTrustingRestTemplate) started as problem mainly because if the SAML bean is used, then there is no pooling but we have seen that then we run into issues in large systems with high amount of traffic.

@adrianhoelzl-sap said that with version 77.27.0 the call

`RestTemplate rt = config.isSkipSslValidation() ? restTemplateConfig.trustingRestTemplate() : restTemplateConfig.nonTrustingRestTemplate();`

resulted in newly created resttemplates, but I could not see this, but maybe at this time the beans from OAuth where used more often than from SamlConfig

strehle: > @Autowired
>     RestTemplateConfig restTemplateConfig

Tested, but without extra bean configuration for the `@Configuration` I get a singleton... so it is the same ref.

To me we can also do this because at least we would have again a RestTemplete with Pooled connections... as we had before..

If you compare to 77.20.x and before then ok.

If you compare a 76.31.0 UAA with ours then as mentioned, because of extra bean configuration in saml-idp.xml the call to the method create always new Resttemplates and Pools (maybe also not that what we want)

strehle: Please check https://github.com/cloudfoundry/uaa/pull/3434 . 

I think it is better to go with https://github.com/cloudfoundry/uaa/pull/3434 and stay with using the beans for RestTemplates.

In debugger I see that SamlConfig beans e.g. nonTrustingRestTemplate wins in start and internally with HttpClient there is a default pooled connection manager created, but only with the default of 10 (all) and 5 (per route). And that would explain our issue in large systems with huge load.

strehle:
used https://github.com/cloudfoundry/uaa/pull/3434

PR #3434, opened 2025-04-30
Fix duplicate bean definition of rest templates -Alternative to PR https://github.com/cloudfoundry/uaa/pull/3430

### LINKED ISSUE #3411: NoHttpResponseException appear after version bump from 77.27.0 to 77.29.0

strehle:
before we had duplicates of:

- nonTrustingRestTemplate
- trustingRestTemplate

With this there is a separation of RestTemplate usage between SAML and OIDC/OAUTH. Duplication is removed in SamlConfig.

Main reason is, that there are SAML timesouts and other time outs.
With opensaml-2 and spring-saml2 there was obviously the same separate usages

In addition to https://github.com/cloudfoundry/uaa/pull/3430 here we add new options to give Operations the power to change settings. 2 new settings

1. validateAfterInactivity - default 2000 (sec) . Is the current default in PoolManger already. In case of load with many Idle connections decrease the value to 100 (ms)
2. retryCount - default 0 . This will add a retry handler to HttpBuilder. This retry will be done to ignore NoHttpResponseException 

Both new options are created because of researches and best practices from https://stackoverflow.com


fhanik (approved):
Much easier to follow along. Approved.

PR #3439, opened 2025-05-02
Refactor SAML authentication

### LINKED ISSUE #3432: SAML Attributes not stored if SAML Bearer is used
Unfortunately there are 2 implementation for SAML SSO and SAML Bearer but only in SAML SSO we validate and store SAML attributes, but not in SAML Bearer flows.

Best is, to combine the user tasks in SamlUaaAuthenticationUserManager which is used for both parts already.

strehle:
SAML Response to UAA user and SAML Bearer Assertion to Uaa user was implemented twice with some different logic.

SAML Response SSO has all we need, but SAML Bearer has no attribute processing and no user attribute store. With the refactoring this should be fixed and should be less code now

Sonar
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=3439

Main refacotoring is: getUaaAuthentication in SamlUaaAuthenticationUserManager.java now, instead of Saml2BearerGrantAuthenticationConverter.java and SamlUaaResponseAuthenticationConverter.java


duanemay (approved):
Thank you for reducing the lines of code and cleaning up around this. LGTM

PR #3445, opened 2025-05-07
Refactor (prepare) before Spring Update

strehle:
- RequestMapping to GetXYZ method


strehle:
This change is recommended before spring framework update 6

duanemay (approved):

PR #3476, opened 2025-05-19
Convert MockMvc tests to @SpringBootTest

fhanik:


uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/audit/AuditCheckMockMvcTests.java
```diff
@@ -826,7 +828,6 @@ void generateUserModifiedEvent_whenUserCreatedByClient() throws Exception {
     }
 
     @Nested
-    @DefaultTestContext
```
fhanik, line 829: This annotation creates another context, and this new context doesn't have the required listeners

uaa/src/test/java/org/cloudfoundry/identity/uaa/mock/audit/AuditCheckMockMvcTests.java
```diff
@@ -986,7 +987,6 @@ void generateUserCreatedEvent_DuringLoginServerAuthorize() throws Exception {
...
```
fhanik, line 989: This annotation creates another context, and this new context doesn't have the required listeners

PR #3487, opened 2025-05-25
Thymeleaf upgrade

strehle:
will come with spring update anyhow


duanemay (approved):
Just noting we should probably remove the version numbers from dependencies after the upgrade

strehle:
> Just noting we should probably remove the version numbers from dependencies after the upgrade

sure, after spring boo3 we can remove most of the versions... beginning from spring framework down to tomcat etc.

PR #3490, opened 2025-05-27
Logging Refactorings

strehle:
Mainly https://docs.openrewrite.org/recipes/java/logging/slf4j/log4jtoslf4j Then cleanup of Sonar smells of classes in edit mode (no general smell fix)

First change was done by openrewrite with log4jtoslf4j 
But then there were many sonar smells about identityZoneManager and since without XMLs the migration is much easier I did it in this PR for the files in edit.


duanemay (approved):
👍 LGTM

PR #3504, opened 2025-06-02
Potential fix for code scanning alert no. 44: HTTP response splitting

strehle:
Potential fix for [https://github.com/cloudfoundry/uaa/security/code-scanning/44](https://github.com/cloudfoundry/uaa/security/code-scanning/44)

To fix the issue, the `userId` parameter should be sanitized to remove or encode any special characters that could lead to HTTP response splitting. A common approach is to URL-encode the value before using it in the `Cookie` constructor. This ensures that any potentially malicious characters are safely encoded and cannot interfere with the HTTP response structure.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._


server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java
```diff
@@ -624,7 +624,8 @@ private Map.Entry<String, AbstractIdentityProviderDefinition> evaluateLoginHint(
 
     @RequestMapping(value = {"/delete_saved_account"})
     public String deleteSavedAccount(HttpServletRequest request, HttpServletResponse response, String userId) {
-        Cookie cookie = new Cookie("Saved-Account-%s".formatted(userId), "");
+        String sanitizedUserId = URLEncoder.encode(userId, UTF_8);
+        Cookie cookie = new Cookie("Saved-Account-%s".formatted(sanitizedUserId), "");
```
fhanik: We also have this [utility method](https://github.com/cloudfoundry/uaa/blob/0c0f9c2c61adfbc43453e0384b39d2e0411ebb0c/server/src/main/java/org/cloudfoundry/identity/uaa/login/AccountSavingAuthenticationSuccessHandler.java#L94-L102) called when the cookie is first set

Maybe break this out into a utility class so that we use the same encoding everywhere?

PR #3505, opened 2025-06-03
Ensure that all Singular URLs can be used. Singular app sometimes app…

fhanik:
…ends "/" which was implicitly mapped previously. Now it needs to be explicitly mapped


duanemay (approved):
Thank you

PR #3535, opened 2025-06-26
Require Java 21

### LINKED ISSUE #3534: [Question] When the support for java 21 will be available?
We are using CF UAA version 75.18.0 with Java 11 and planning to upgrade to latest CF UAA with Java 21. I see current development happening with [Java 17](https://github.com/cloudfoundry/uaa/blob/develop/build.gradle#L59-L62), are there any plans for java 21 upgrade?

duanemay:
Update build to require Java 21
Begin using Java 21 APIs and remove deprecations

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -6,7 +6,7 @@ on:
   pull_request:
     branches: [ develop ]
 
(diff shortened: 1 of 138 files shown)


strehle (approved):
first, was not sure if we should bring it into next release already, but ok, I incr. the major version and then we may should go this step as well, so you can decide, but ok for me

PR #3551, opened 2025-07-14
Push UAA Docker images to cfidentity/uaa

### LINKED ISSUE #3508: [Question] UAA Dockerfile Missing and No Image Updates in 4+ Months – Expected?
## Hi team,

While reviewing the Docker images for UAA, I noticed that the latest published tag on Docker Hub appears to be from **4 months ago**:  
📦 [`cloudfoundry/uaa:77.21.0`](https://hub.docker.com/r/cloudfoundry/uaa/tags)

I had a couple of questions:

- Is this gap in releases **intentional or by design**?
- Is there a **new release cadence**, or perhaps images are now being published somewhere else?
- Is there a difference between **rolling (e.g. `v77.34`)** and **stable (e.g. `v77.21.0`)** releases?
- Just trying to understand if **newer versions** are being tracked elsewhere or under **different tags**.

I also no longer see a **Dockerfile** in the UAA repo:
- Has the Dockerfile been **moved to a different repository**, or is the image build now managed through an **internal/private pipeline**?

Any context would be greatly appreciated — thank you so much for maintaining this!

strehle:
Push to 
https://hub.docker.com/r/cfidentity/uaa/tags

instead of
https://hub.docker.com/r/cloudfoundry/uaa/tags

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -28,7 +28,7 @@ jobs:
         run: |
           echo running on branch ${GITHUB_REF##*/}
           if [ "${GITHUB_REF##*/}" == "develop" ]; then
-            export BP_JVM_TYPE=JDK && ./gradlew --no-daemon --max-workers 1 bootBuildImage --imageName docker.io/cloudfoundry/uaa:${{ github.run_number }} --publishImage
+            export BP_JVM_TYPE=JDK && ./gradlew --no-daemon --max-workers 1 bootBuildImage --imageName docker.io/cfidentity/uaa:${{ github.run_number }} --publishImage
           else
-            export BP_JVM_TYPE=JDK && ./gradlew --no-daemon --max-workers 1 -Pversion=${GITHUB_REF##*/} bootBuildImage --imageName docker.io/cloudfoundry/uaa:${GITHUB_REF##*/} --publishImage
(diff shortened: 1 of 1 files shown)


strehle:
@beyhan  @rkoster Hope you have the knowledge where we store the credentials to docker hub, e.g. 

We would like to push docker images from UAA releases https://hub.docker.com/r/cloudfoundry/uaa/tags but we have only the credentials / access to cfidentity.

cfidentity is used for our internal pipeline mainly, but if we dont get cloudfoundry access we will use this repo for now.

If you have access/credentials to cloudfoundry can you update please the secret section in
https://github.com/cloudfoundry/uaa/settings/secrets/actions

@duanemay has update it to cfidentity so if we dont get cloudfoundry we have to to with cfidentity

beyhan:
In bosh areas we startet to publish images in https://github.com/orgs/cloudfoundry/packages?tab=packages&q=bosh. We don't use docker hub any more. It is simple because you can do with the same github user/token and doesn't generate costs (currently). If you don't resist on docker hub it could be an option for you. If you would like to go with docker hub I would suggest to approach @christopherclark regarding access.

strehle:
> In bosh areas we startet to publish images in https://github.com/orgs/cloudfoundry/packages?tab=packages&q=bosh. We don't use docker hub any more. It is simple because you can do with the same github user/token and doesn't generate costs (currently). If you don't resist on docker hub it could be an option for you. If you would like to go with docker hub I would suggest to approach @christopherclark regarding access.

Thank you @beyhan we will discuss then in team about both or one of the options

strehle:
> Thank you @beyhan we will discuss then in team about both or one of the options

discussed and we go with docker hub first with cfidentity

PR #3552, opened 2025-07-14
Add "urn:ietf:params:oauth:grant-type:token-exchange" as a grant

fhanik:
Partial Implemention of RFC 8693 - 

Highlights

1. Implements impersonation token exchange
2. Client requires `urn:ietf:params:oauth:grant-type:token-exchange` grant type
3. Client authentication supported are `basic`, `form` or `client_assertion`
4. Functions similar to the `urn:ietf:params:oauth:grant-type:jwt-bearer` grant type, with an added ability to impersonate the client that performed the grant (like the `user_token` grant type)
5. Resulting tokens have a new claim `act` the represents the origin of the request
6. Downstream implementations can override the `authentication manager` (bean name: `tokenExchangeAuthenticationManager`; for user authentications) and the `TokenExchangeGranter` (bean name: `tokenExchangeGranterBean`) to override the grant itself


fhanik:
Items left

1. Add the `act` claim with the identity of the caller
2. Allow the `audience` parameter to override the client_id in the resulting token (privilege escalation). This MUST have some sort of authorized scope on the calling client to allow this impersonation to take place
3. client_assertion tests

strehle:
FYI request review for this PR from copilot

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/TokenExchangeWrapperForExternalOauth.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public TokenExchangeWrapperForExternalOauth(final ExternalOAuthAuthenticationManager authenticationManager) {
+        this.delegate = authenticationManager;
+    }
+
+    @Override
+    public Authentication authenticate(Authentication authentication) throws AuthenticationException {
+        if (authentication instanceof ExternalOAuthCodeToken token) {
+            if (token.getIdToken() == null) {
+                if (token.getIdToken() == null) {
```
duanemay: already checked token.getIdToken() == null

duanemay (dismissed):

duanemay (approved):

PR #3567, opened 2025-07-23
Fix StaleUrlCache test

fhanik:
We can wait until the removal thread has done its job.


strehle (dismissed):
thanks for taking care of it

strehle (approved):

PR #3571, opened 2025-07-30
parse JWK from RSA public key value only if exists

### LINKED ISSUE #3570: JWT signature validation fails with x5t#s256
JWK is described by [RFC 7517](https://datatracker.ietf.org/doc/html/rfc7517).

There is a bug in UAA SignatureVerifier where it does a conversion from JsonWebKey to a Nimbus JWK:
https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/jwt/SignatureVerifier.java#L42

After reading the JWK set from an external OIDC provider, UAA deserializes the JSON to a Java class named JsonWebKey. This class has only the fields for a JWK that it needs so it does not have x5t#s256. This is not the problem because the class hangs onto the original json of the JWK. Right before calling into the Nimbus JOSE JWT library to do the actual validation, the SignatureVerifier in UAA needs to convert the JsonWebKey to a Nimbus JWK class. This conversion is where the bug is. For RSA keys it looks for a non-standard "value" field containing a PEM-encoded RSA public key. The JWK from the external OIDC provider doesn't contain the "value" field but the JsonWebKey code [here](https://github.com/cloudfoundry/uaa/blob/develop/model/src/main/java/org/cloudfoundry/identity/uaa/oauth/jwk/JsonWebKey.java#L164) tries to be helpful and creates a PEM-encoded RSA public key from the fields it has and returns that. The Nimbus JWK is parsed from this and consequently only has the limited RSA public key fields. Note that if the value were null or the SignatureVerifier just ignored the value and parsed the JWK from the key properties, which contains all the JSON fields in a map, it would have the x5t#s256 attribute.

Anyway, later on when Nimbus is validating a JWT with the x5t#s256 attribute, it requires the JWK to contain an x5c to compute the hash from. But the JWK it gets passed from UAA doesn't have that field due to the conversion bug, and the validation fails.

This probably worked prior to the switch to use Nimbus JOSE JWT because I'm guessing the out-dated Spring Security JWT was ignoring x5t#s256. Once UAA upgraded to Nimbus JOSE JWT, it started applying it correctly.

mikeroda:
- if the JWK json doesn't have a value, parse the JWK from the raw json to get all the attributes like x5t#s256, etc
- Fixes #3570

### DIFF
--- a/model/src/test/java/org/cloudfoundry/identity/uaa/oauth/jwk/JsonWebKeyTests.java
+++ b/model/src/test/java/org/cloudfoundry/identity/uaa/oauth/jwk/JsonWebKeyTests.java
@@ -36,4 +36,49 @@ class JsonWebKeyTests {
         Map<String, Object> jsonMap = Map.of("kid", "uaa-key");
         assertThatExceptionOfType(IllegalArgumentException.class).isThrownBy(() -> new JsonWebKey(jsonMap));
     }
+
+    @Test
+    void rsaPublicKey() {
+        // given
+        String value = "-----BEGIN PUBLIC KEY-----\n" +
+                "MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAvc5lyy8d+oOBDdgB8XsG\n" +
+                "Nw4Zj7W6nbTF03qs/bL0CdgaIuMCK1hHDIBInHCRDYykIKwnwSdXi52QGeVAnUZa\n" +
+                "86Tf/LyV+MNY7OOgTiQOP0fKwIt45MTGJX/FxKYMt3KZUkD9Xp+Pfj8/EeehvdY+\n" +
+                "shOG6TwU1OayElu+WLQd7/qDsAjnNzhEv35R1BwMoMoSrN33d1HQSvFnNlX2R39q\n" +
+                "X7YEc/bLG7JMBP/4MW3my2MK9AqC08w7P4dJkOVz+Zcf1wtV1oWX2Ddt4N4MdhGy\n" +
+                "Fmg5S1XaZodaAL0T79VOOKVQ3fIxRaQjPZ522crpc6/AIFEPD+e4ezh5UcVaBI2N\n" +
+                "uQIDAQAB\n" +
+                "-----END PUBLIC KEY-----";
+        Map<String, Object> jsonMap = Map.of(
+                "kid", "uaa-key",
+                "kty", "RSA",
+                "e", "AQAB",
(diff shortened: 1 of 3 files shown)


model/src/test/java/org/cloudfoundry/identity/uaa/oauth/jwk/JsonWebKeyTests.java
```diff
@@ -36,4 +36,49 @@ void webKeyPublicNoTypeException() {
         Map<String, Object> jsonMap = Map.of("kid", "uaa-key");
         assertThatExceptionOfType(IllegalArgumentException.class).isThrownBy(() -> new JsonWebKey(jsonMap));
     }
+
+    @Test
+    void rsaPublicKey() {
+        // given
+        String value = "-----BEGIN PUBLIC KEY-----\n" +
```
strehle: please use text blocks

https://sonarcloud.io/project/issues?issueStatuses=OPEN%2CCONFIRMED&sinceLeakPeriod=true&pullRequest=3571&id=cloudfoundry-identity-parent

model/src/test/java/org/cloudfoundry/identity/uaa/oauth/jwk/JsonWebKeyTests.java
```diff
@@ -36,4 +36,49 @@ void webKeyPublicNoTypeException() {
...
+        JsonWebKey jsonWebKey = new JsonWebKey(jsonMap);
+        assertThat(jsonWebKey.hasValue()).isFalse();
+        assertThat(jsonWebKey.getValue()).isEqualTo(value);
+    }
+
+    @Test
+    void rsaPublicKeyWithValue() {
+        // given
+        String value = "-----BEGIN PUBLIC KEY-----\n" +
+                "MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAvc5lyy8d+oOBDdgB8XsG\n" +
```
strehle: please use text blocks https://sonarcloud.io/project/issues?issueStatuses=OPEN%2CCONFIRMED&sinceLeakPeriod=true&pullRequest=3571&id=cloudfoundry-identity-parent

strehle (dismissed):
ok, only 2 minor sonar issues

strehle (dismissed):

strehle (approved):

strehle:
https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=3571

PR #3575, opened 2025-08-07
Refactor ExternalOAuthAuthenticationManager

adrianhoelzl-sap:


server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
```diff
@@ -970,43 +974,12 @@ public String oauthTokenRequest(UaaAuthenticationDetails details, final Identity
         return idToken;
     }
 
+    @Setter
```
duanemay: Could this be `@Data`

adrianhoelzl-sap: fixed with ce26932

duanemay (approved):

PR #3582, opened 2025-08-12
The jwt-bearer grant confuses internal/external groups 

fhanik:
the internal (mapped) authorities(scope) are different from the incoming(external) groups

The way the flow should work is

1. Extract external groups via the external_groups attribute mapping
2. Filter the external groups via the external group white list (Note Nr.1)
3. Store the filtered external groups (external authorities)
4. Map the filtered external groups to scope, and store the scopes (authorities)

Note Nr/1
For LDAP, and empty whitelist means no external groups are allowed For SAML and OAUTH/OIDC an empty whitelist means all external groups are allowed


strehle:
review sonar: https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=3582

strehle (dismissed):
requested sonar and copilot, but ok to me

fhanik:
@strehle I have added the copilot suggestions, and I have fixed the sonarcloud critical complaints (namely the <? extends GrantedAuthority>

The latter required me to refactor some of the methods to use the subclass, SimpleGrantedAuthority, instead. 
I also deleted ExternalOAuthUserAuthority.java as it was basically a duplicate of the class mentioned above. This is a breaking change, so I can undo it if needed, but it makes the sonar fix a bit more difficult.

duanemay (approved):
👍

strehle (approved):
all good with sonar

https://sonarcloud.io/summary/new_code?id=cloudfoundry-identity-parent&pullRequest=3582

co-pilot is just something you can think about but nothing which should block you

PR #3595, opened 2025-08-22
Update boot from 3.5.4 to 3.5.5

strehle:


strehle:
here is works https://github.com/cloudfoundry/cf-identity-acceptance-tests-release/pull/97 but not in UAA

PR #3611, opened 2025-09-05
Update database matrix to include MySQL 8.4 and 9, remove MySQL 5

duanemay:
Also adds testing for 'postgresql-18'. Refactor the test scripts.


scripts/start_docker_database.sh
```diff
@@ -0,0 +1,50 @@
...
+    --privileged \
+    --tty \
+    --interactive \
+    --platform linux/amd64 \
+    --shm-size=1G \
+    --volume "${uaa_dir}":"${container_script_dir}" \
+    --volume "${gradle_lock_dir}" \
+    --env DB="${DB}" \
+    --publish "${db_port}:${db_port}" \
+    #--publish "5005:5005" \
```
strehle: the comment here is mainly from co pilot but a good point.

Maybe we add a check for env variable or argument here and then create pubsh list based on debug wanted or not.

I think support debug or xdebug in spring.profiles.active property so maybe we can use this ?

strehle (dismissed):
thanks, in general we can go with this but only the debug is something you can think about

scripts/start_docker_database.sh
duanemay: went with the env variables for now

strehle (dismissed):

fhanik (approved):

PR #3614, opened 2025-09-08
info endpoint providers

fhanik:
We'd like a bit more public visibility into the federation of providers in the /info endpoint. This PR displays an JSON object named `providers` with three attributes (name, origin, type) on the /info endpoint when JSON is requested.

You can today, go to the /login page, and extract origins for both SAML and OIDC providers, so we do believe it is ok to display this information.

To be discussed September 11, 2025 in UAA meeting

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java
@@ -361,6 +361,21 @@ public class LoginInfoEndpoint {
             returnLoginPrompts = false;
         }
 
+        if (jsonResponse && request.getRequestURI().endsWith("/info")) {
+            List<Map<String, String>> infoIdentityProviders = new java.util.ArrayList<>();
+            providerProvisioning.retrieveAll(true, IdentityZoneHolder.get().getId())
+                    .forEach(provider -> {
+                        infoIdentityProviders.add(
+                                Map.of(
+                                        OriginKeys.ORIGIN, provider.getOriginKey(),
+                                        "name", provider.getName(),
+                                        "type", provider.getType()
+                                )
+                        );
+                    });
+            model.addAttribute("providers", infoIdentityProviders);
+        }
+
         // ldap or uaa not part of allowedIdentityProviderKeys
         if (allowedIdentityProviderKeys != null &&
                 !allowedIdentityProviderKeys.contains(OriginKeys.LDAP) &&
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java
```diff
@@ -361,6 +361,21 @@ private String login(Model model, Principal principal, List<String> excludedProm
             returnLoginPrompts = false;
         }
 
+        if (jsonResponse && request.getRequestURI().endsWith("/info")) {
+            List<Map<String, String>> infoIdentityProviders = new java.util.ArrayList<>();
+            providerProvisioning.retrieveAll(true, IdentityZoneHolder.get().getId())
```
strehle, line 366: We have tweaked this endpoint the last years not to retrrieve always all IdP because we (SAP) have many IdPs in one zone and with cf login there is the situation, that /info is called before the login and with that we had DB issues

@torsten-sap correct or is my assumption no longer valid ?

strehle: https://github.com/cloudfoundry/uaa/pull/3254

server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java
```diff
@@ -361,6 +361,21 @@ private String login(Model model, Principal principal, List<String> excludedProm
...
+                    .forEach(provider -> {
```
strehle, line 367: I think we have to filter , because SAML and OAUTH use a property to determine if visible or not

fhanik: The`true` in `retrieveAll(true` is that property, isn't it? It means active providers only.

strehle: not only acitve, but only provider with either
* showSamlLink
* showLinkText

server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpoint.java
fhanik: Sounds like a property to explicit enable this feature is needed, with the default being set to `false`

strehle: problem is that we had many changes in this area, some because to fix DB performance, and later some to repair some regressions, we should discuss in meeting

fhanik:
I am abandoning this effort. We're going a different route for the use case. No need to add more complexity to /login and /info

PR #3619, opened 2025-09-10
Clear Origin ThreadLocal after AuthN in ExternalOAuthAuthenticationManager

adrianhoelzl-sap:


server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
```diff
@@ -178,6 +178,16 @@ public void setOrigin(String origin) {
         this.origin.set(origin);
     }
 
+    @Override
+    public Authentication authenticate(final Authentication request) throws AuthenticationException {
+        final Authentication authentication = super.authenticate(request);
+
+        // clear ThreadLocal holding the origin key
+        origin.remove();
```
strehle: I was not aware about the thread local part... where do we access this ?

adrianhoelzl-sap: It is used in various methods in this and the superclass, e.g., when querying a potentially existing shadow user. The origin is set early in `getExternalAuthenticationDetails`.

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
```diff
@@ -178,6 +178,16 @@ public void setOrigin(String origin) {
...
+
+        return authentication;
```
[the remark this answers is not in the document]
adrianhoelzl-sap: fixed with commit [29d9265](https://github.com/cloudfoundry/uaa/pull/3619/commits/29d926551cd15d7fa225b4d235f5fd748707abf8)

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
fhanik: Let me think about this. It appears that the `setOrigin` would be the proper way to clear this threadlocal. I know we have sub classes of this class, and this change may impact those subclasses negatively, especially since this commit changes the nature of `super.authenticate` in subclasses.

adrianhoelzl-sap: Good point, I decided to introduce a separate method for clearing the origin which is called in a `try-finally` in `ExternalLoginAuthenticationManager.authenticate`.

Then, every subclass can override the three methods depending on how it wants to manage the origin.

strehle (approved):

PR #3628, opened 2025-09-19
Move test database properties to test resources

duanemay:
## What

`application-mysql.properties`/`application-postgresql.properties` in `server/src/main/resources` shipped hardcoded `database.username=root`/`database.password=changeme`/a `127.0.0.1`/local JDBC URL — these look like local test/dev fixtures, not production defaults, and end up packaged into the WAR. This moves those values into `src/test/resources` (kept for the test suite, which needs them), leaving only `database.driverClassName` in the packaged `application-mysql.properties`/`application-postgresql.properties`.

This is not a fix for an exploitable default: UAA falls back to the `hsqldb` profile with zero configuration (`YamlServletProfileInitializer.applySpringProfiles`), so these values are only ever read if an operator explicitly activates the `mysql`/`postgresql` profile. It's hygiene — don't ship plausible-looking credentials in the production artifact.

## Follow-up fix included

Removing these from `src/main/resources` broke three local/CI flows that relied on them being present in the packaged artifact when the `mysql`/`postgresql` profile is active, with no other credentials supplied:

- `./gradlew run` / `bootRun` with `-Dspring.profiles.active=postgresql` or `mysql` (documented in `README.md`)
- the embedded `integrationTest` auto-start (`java -jar` launch in `uaa/build.gradle.kts`)
- CI's external-Tomcat integration test matrix (`scripts/tomcat/integration_tests_tomcat.sh`, `.github/workflows/integration-tests.yml`, mysql/postgresql matrix)

All three now supply the same `root`/`changeme` local-dev credentials via explicit `-D` system properties (still overridable) instead of relying on the packaged properties files. Verified by running `bootRun -Dspring.profiles.active=postgresql` against a real local Postgres — UAA started and Flyway created all tables as expected.

## Test plan

- [x] `./gradlew :cloudfoundry-identity-uaa:bootRun -Dspring.profiles.active=postgresql -Ddatabase.url=jdbc:postgresql://localhost:5433/uaa` against a local Postgres with a `root`/`changeme` superuser — UAA started, Flyway created all 16 tables
- [x] `shellcheck scripts/tomcat/integration_tests_tomcat.sh` — no new findings
- [x] CI mysql/postgresql integration test matrix (this PR)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -2,6 +2,28 @@ import org.apache.tools.ant.filters.ReplaceTokens
 import java.nio.file.Path
 import java.nio.file.Paths
 
+// Local-dev-only DB credentials for the mysql/postgresql profiles, matching the database
+// bootstrapped by README.md / scripts/lib_db_helper.sh / scripts/docker-compose.yml.
+// These are intentionally not shipped in application-mysql.properties/application-postgresql.properties
+// (which are packaged into the WAR), so callers that boot a real mysql/postgresql locally must
+// supply them here instead. Override any of them with the matching -D system property.
+fun localDatabaseCredentialArgs(activeProfiles: String): Map<String, String> {
+    val profiles = activeProfiles.split(",").map { it.trim() }
+    return when {
+        profiles.contains("mysql") -> mapOf(
+            "database.username" to System.getProperty("database.username", "root"),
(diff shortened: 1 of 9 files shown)


fhanik:
rebased on develop branch

PR #3635, opened 2025-09-30
Boot Migration - Backwards compatible request mappings (end with slash)

fhanik:
we went from 

```
@RequestMapping(value = "/oauth/clients/restricted", method = RequestMethod.POST)
```

to

```
@PostMapping("/oauth/clients/restricted")
```

**It breaks here**, urls ending with `/` no longer work properly, whereas they worked before

and it needs to be

```
@PostMapping({"/oauth/clients/restricted", "/oauth/clients/restricted/"})
```

The following URLs now support ending with `/`. The only endpoint that did *not* support an ending slash prior to the boot migration was `/Users`

1. /Codes
1. /Groups
1. /Groups/External
1. /Groups/zones
1. /Users
1. /authenticate
1. /force_password_change
1. /identity-zones
1. /identity-zones/{zoneId}/clients
1. “{zoneId}/clients
1. /invite_users
1. /oauth/clients
1. /oauth/clients/tx
1. /oauth/clients/tx/delete
1. /oauth/clients/tx/modify
1. /oauth/clients/tx/secret
1. /password_change
1. /password_resets
1. /profile

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/account/PasswordResetEndpoint.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/account/PasswordResetEndpoint.java
@@ -63,7 +63,7 @@ public class PasswordResetEndpoint {
         this.identityZoneManager = identityZoneManager;
     }
 
-    @PostMapping("/password_resets")
+    @PostMapping({"/password_resets", "/password_resets/"})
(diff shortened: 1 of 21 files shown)


strehle (dismissed):

duanemay (approved):
Thank you

PR #3643, opened 2025-10-08
Fix broken OAUTH2.0 authorization_code flow.

fhanik:
When configuring an Oauth2.0 provider that returns opaque access tokens, the UAA throws an error trying to parse it as it assumes it is a JWT id_token

Easy to reproduce by adding Github as an OAuth provider in uaa.yml

```
login:
  oauth:
    providers:
      github:
        type: oauth2.0
        authUrl:     https://github.com/login/oauth/authorize
        tokenUrl:     https://github.com/login/oauth/access_token
        userInfoUrl:     https://api.github.com/user
        issuer: https://github.com
        relyingPartyId: <your-github-app-client-id>
        relyingPartySecret: <your-github-app-client-secret>
        performRpInitiatedLogout: false
        scopes:
          - openid
        linkText: Login with Github
        showLinkText: true
        attributeMappings:
          user_name: login
        clientAuthInBody: true
        externalGroupsWhitelist:
          - "*"
```

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManager.java
@@ -116,6 +116,7 @@ import static java.util.Collections.singletonList;
 import static java.util.Collections.singletonMap;
 import static java.util.Objects.isNull;
 import static java.util.stream.Collectors.toSet;
+import static org.cloudfoundry.identity.uaa.constants.OriginKeys.OAUTH20;
 import static org.cloudfoundry.identity.uaa.oauth.token.ClaimConstants.SUB;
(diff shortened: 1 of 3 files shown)


strehle:
I wonder about because github use case it not what I tested that much but there were some changes and improvements in past, e.g. https://github.com/cloudfoundry/uaa/commits/7e671a0ea1ba4c74b096dcfd33685edadbf82dce/docs/github-oauth2-provider.md so that I was thinking we have all, e.g. tests for this are in
 https://github.com/cloudfoundry/uaa/blob/develop/server/src/test/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthAuthenticationManagerGithubTest.java

fhanik:
@strehle @adrianhoelzl-sap It appears that we no longer have an actual Github OAuth2.0 test, so I added one to make it clear.

You can take commit 4ab33f74f87e91155e44f9b07d9d272fd7df4cd7 against develop and see it failing.

strehle (approved):

PR #3652, opened 2025-10-24
Remove duplicate integration-test workflow

duanemay:
As of the removal of cargo, boot-integration-tests.yml and integration-tests.yml are now doing the same thing


duanemay:
Included in #3654

Issue #3630, opened 2025-09-24
cf login --sso no longer display passcode url


haochenhu233:
Since the commit d8c10f0 in version v77.27.0 of uaa-release, running the command cf login --sso with okta as IdP doesn't show the passcode url anymore.

Before the commit:
```
$ cf login --sso
API endpoint: api.cf.example.com
Temporary Authentication Code ( Get one at https://login.cf.example.com/passcode ):
```
After the commit:
```
$ cf login --sso
API endpoint: api.cf.example.com
:
```

With env var CF_TRACE=true
We have:
before the commit:
```
HTTP/1.1 200 OK
Cache-Control: no-store
Content-Language: en-US
Content-Security-Policy: script-src 'self'
Content-Type: application/json;charset=UTF-8
Date: Wed, 13 Aug 2025 18:03:51 GMT
Set-Cookie: [PRIVATE DATA HIDDEN]
Strict-Transport-Security: max-age=31536000
X-Content-Type-Options: nosniff
X-Frame-Options: DENY
X-Vcap-Request-Id: 80942449-30f7-4577-4260-ab4aeb90789e
X-Xss-Protection: 1; mode=block
{
  "app": {
    "version": "77.17.0"
  },
  "commit_id": "e42b84e",
  "entityID": "http://login.example.com/",
  "idpDefinitions": {},
  "links": {
    "login": https://login.example.com/ ,
    "passwd": https://login.example.com/forgot_password,
    "register": https://login.example.com/create_account,
    "uaa": https://uaa.example.com/
  },
  "prompts": {
    "passcode": [
      "password",
      "Temporary Authentication Code ( Get one at https://login.example.com/passcode )"
    ],
    "password": "[PRIVATE DATA HIDDEN]",
    "username": [
      "text",
      "Email"
    ]
  },
  "showLoginLinks": true,
  "timestamp": "2024-10-08T06:53:39+0000",
  "zone_name": "uaa"
}

Temporary Authentication Code ( Get one at https://login.example.com/passcode ):
```

After the commit d8c10f0:
```
HTTP/1.1 200 OK
Cache-Control: no-store
Content-Language: en-US
Content-Security-Policy: script-src 'self'
Content-Type: application/json;charset=UTF-8
Date: Wed, 24 Sep 2025 20:47:18 GMT
Set-Cookie: [PRIVATE DATA HIDDEN]
Strict-Transport-Security: max-age=31536000
X-Content-Type-Options: nosniff
[...]

PR #3660, opened 2025-10-31
Revert to legacy behavior of showing `passcode` as a prompt in /info JSON response.

fhanik:
Goal: Maintain backwards compatible behavior

To disable this behavior, example for performance reasons, set `login.allowOriginLoop=false`

Also referenced
https://github.com/cloudfoundry/uaa/issues/3630

### DIFF
--- a/server/src/test/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpointTests.java
+++ b/server/src/test/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpointTests.java
@@ -31,6 +31,7 @@ import org.cloudfoundry.identity.uaa.zone.beans.IdentityZoneManager;
 import org.cloudfoundry.identity.uaa.zone.beans.IdentityZoneManagerImpl;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.springframework.dao.DataAccessException;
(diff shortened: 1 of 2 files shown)


server/src/test/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpointTests.java
```diff
@@ -826,14 +826,34 @@ void oauth_provider_links_shown() throws Exception {
...
+
+        RawExternalOAuthIdentityProviderDefinition definition = new RawExternalOAuthIdentityProviderDefinition()
+                .setAuthUrl(URI.create("http://auth.url").toURL())
+                .setTokenUrl(URI.create("http://token.url").toURL());
+
+        IdentityProvider<AbstractExternalOAuthIdentityProviderDefinition> identityProvider = MultitenancyFixture.identityProvider("oauth-idp-alias", "uaa");
+        identityProvider.setConfig(definition);
+
+        when(mockIdentityProviderProvisioning.retrieveActiveByTypes(anyString(), eq(OriginKeys.OIDC10), eq(OriginKeys.OAUTH20)))
+                .thenReturn(singletonList(identityProvider));
```
adrianhoelzl-sap: Would it be possible to extract this into a method, e.g., `arrangeActiveOAuthIdpExists`, and reuse it in the different test cases?

server/src/test/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpointTests.java
adrianhoelzl-sap: If I didn't miss anything, this should be the same code

server/src/test/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpointTests.java
```diff
@@ -826,14 +826,34 @@ void oauth_provider_links_shown() throws Exception {
...
-        when(mockIdentityProviderProvisioning.retrieveActiveByTypes(anyString(), any(), any())).thenReturn(singletonList(identityProvider));
+        when(mockIdentityProviderProvisioning.retrieveActiveByTypes(anyString(), any(String[].class))).thenReturn(singletonList(identityProvider));
         endpoint.loginForHtml(extendedModelMap, null, new MockHttpServletRequest(), singletonList(MediaType.TEXT_HTML));
 
         assertThat((Boolean) extendedModelMap.get("showLoginLinks")).isTrue();
     }
 
+    @Test
+    void passcode_prompt_present_whenThereIsAtleastOneActiveOauthProvider_AllowOriginLoop() throws Exception {
+        LoginInfoEndpoint endpoint = getEndpoint(IdentityZoneHolder.get());
```
adrianhoelzl-sap: Can we add a comment here that the default value for the `allowOriginLoop` setting is true?

server/src/test/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpointTests.java
```diff
@@ -851,8 +871,29 @@ void no_passcode_prompt_present_whenThereIsAtleastOneActiveOauthProvider() throw
...
+
+        RawExternalOAuthIdentityProviderDefinition definition = new RawExternalOAuthIdentityProviderDefinition()
+                .setAuthUrl(URI.create("http://auth.url").toURL())
+                .setTokenUrl(URI.create("http://token.url").toURL());
+
+        IdentityProvider<AbstractExternalOAuthIdentityProviderDefinition> identityProvider = MultitenancyFixture.identityProvider("oauth-idp-alias", "uaa");
+        identityProvider.setConfig(definition);
+
+        when(mockIdentityProviderProvisioning.retrieveActiveByTypes(anyString(), eq(OriginKeys.OIDC10), eq(OriginKeys.OAUTH20)))
+                .thenReturn(singletonList(identityProvider));
```
adrianhoelzl-sap: here as well

server/src/test/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpointTests.java
```diff
@@ -826,14 +826,34 @@ void oauth_provider_links_shown() throws Exception {
...
 
-        when(mockIdentityProviderProvisioning.retrieveActiveByTypes(anyString(), any(), any())).thenReturn(singletonList(identityProvider));
+        when(mockIdentityProviderProvisioning.retrieveActiveByTypes(anyString(), any(String[].class))).thenReturn(singletonList(identityProvider));
         endpoint.loginForHtml(extendedModelMap, null, new MockHttpServletRequest(), singletonList(MediaType.TEXT_HTML));
 
         assertThat((Boolean) extendedModelMap.get("showLoginLinks")).isTrue();
     }
 
+    @Test
+    void passcode_prompt_present_whenThereIsAtleastOneActiveOauthProvider_AllowOriginLoop() throws Exception {
```
adrianhoelzl-sap: Would it make sense to group the three test cases in an inner class annotated with `@Nested`?

server/src/test/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpointTests.java
fhanik: Absolutely

server/src/test/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpointTests.java
fhanik: Absolutely. I have added two nested classes 

- WhenAllowOriginLoopHasDefaultValueOfTrue
- WhenAllowOriginLoopIsSetToFalse

This makes it clear what the default value is true, and that we are setting it to false

adrianhoelzl-sap (approved):

Issue #3661, opened 2025-11-04
SAML Metadata - incorrect SAML metadata is generated when entityID is a URL


fhanik:
in uaa.yml if we have

```
login:
  entityID: http://some.saml.com/entityID
```

Previous version generated SAML metadata:

```
<?xml version="1.0" encoding="UTF-8"?>
<md:EntityDescriptor 
    ID="http___some.saml.com_entityID" 
    entityID="http://some.saml.com/entityID" 
    xmlns:md="urn:oasis:names:tc:SAML:2.0:metadata">
...
    <md:AssertionConsumerService 
        Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST" 
        Location="http://localhost:8080/uaa/saml/SSO/alias/some.saml.com" 
        index="0" 
        isDefault="true"/>
    <md:AssertionConsumerService 
        Binding="urn:oasis:names:tc:SAML:2.0:bindings:URI" 
        Location="http://localhost:8080/uaa/oauth/token/alias/some.saml.com" 
        index="1"/>
...
</md:EntityDescriptor>
```

The 78.x release train generates the following (the `AssertionConsumerService.Location` attribute is incorrect)

```
<?xml version="1.0" encoding="UTF-8"?>
<md:EntityDescriptor 
    ID="http___some.saml.com_entityID"
    entityID="http://some.saml.com/entityID"
    xmlns:md="urn:oasis:names:tc:SAML:2.0:metadata">
...
    <md:AssertionConsumerService 
        Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST" 
        Location="http://localhost:8080/uaa/saml/SSO/alias/http:/some.saml.com/entityID"
        index="0" 
        isDefault="true"/>
    <md:AssertionConsumerService 
        Binding="urn:oasis:names:tc:SAML:2.0:bindings:URI" 
        Location="http://localhost:8080/uaa/saml/SSO/alias/http:/some.saml.com/entityID"
        index="1"/>
...
</md:EntityDescriptor>
```

and this must also work correctly for SAML metadata in a zone.

The old code used to do (in Spring XML - see `getHostIfArgIsURL` )

```
    <bean id="samlSPAlias" class="java.lang.String">
        <constructor-arg value="${login.saml.entityIDAlias:${login.entityID:unit-test-sp}}"/>
    </bean>

[...]

PR #3662, opened 2025-11-05
Fix SAML Metadata when EntityID is a URL

fhanik:
https://github.com/cloudfoundry/uaa/issues/3661

The alias cannot be a URL. Backwards compatible behavior is to use the hostname for the alias


joemahady-comm (commented):
LGTM

duanemay (approved):
👍

PR #3666, opened 2025-11-11
[bc-212-test]: testing memory pressure issues

joemahady-comm:


build.gradle
```diff
@@ -198,10 +212,6 @@ tasks.register('manifests', Copy) {
     into("build/sample-manifests")
 }
 
-tasks.register('integrationTest', Test) {
```
duanemay, line 201: This is actually not a duplicate of line 122, this is part of the parent, the other is in a subprojects block and applied to each subproject

build.gradle
```diff
@@ -120,8 +121,21 @@ subprojects {
     }
 
     tasks.register('integrationTest', Test) {
+        dependsOn subprojects.integrationTest
```
duanemay, line 124: dependsOn subprojects inside the subprojects block?

fhanik (approved):

PR #3679, opened 2025-11-20
feat: add index on group_membership(identity_zone_id, origin)

tack-sap:
This PR introduces an index on group_membership(identity_zone_id, origin) for PostgreSQL

We experienced an issue with deleting an identity provider when the group_membership table had a very large number of entries in a single zone, where we ran into our configured timeouts on DB side while trying to delete the group_memberships. The delete was triggered by this statement during idp deletion: https://github.com/cloudfoundry/uaa/blob/develop/server/src/main/java/org/cloudfoundry/identity/uaa/scim/jdbc/JdbcScimUserProvisioning.java#L521

Even when there was not a single entry with the same origin, the deletion failed as the DB needed to perform a sequential scan on the complete zone just to find that there is nothing to delete.

With this index this issue disappeared.

We saw that this index also took a long time to create, but it resolved our problem and we did not see negative impact on the DB performance so far.


duanemay:
I have done some testing with the mysql equivalent, and will add to this PR

duanemay (approved):

PR #3713, opened 2026-01-08
Refactor: Create Login State Values Once

### LINKED ISSUE #3701: Invalid state parameter when multiple login attempts are initiated with external IdP in different browser tabs
**Issue Summary**

When Cloud Foundry UAA identityZone is configured with an external Identity Provider (Okta, Entra, etc), users encounter an “Invalid state parameter” error if they initiate the login flow in two different browser tabs using the same browser.
The second login attempt overwrites the state stored in UAA’s session, causing the first login callback to fail.

**Steps to Reproduce**

1. Configure CF UAA identity zone with an external OIDC provider (Okta).
2. From the same browser session, open two tabs:
3. Tab 1 → navigate to /oauth/authorize endpoint
4. Tab 2 → navigate to /oauth/authorize endpoint
5. UAA generates a state value and stores it in the session for Tab 1 (S1).
6. User triggers login from Tab 2 → UAA overwrites the session state with S2.
7. Complete the login in Tab 1 by entering Okta credentials.
8. Okta redirects back to UAA with state=S1.
9. UAA rejects the request with “Invalid state parameter” because the session now contains state=S2.

**Expected Behavior**

Ideally, UAA should either:

1. Support multiple concurrent login transactions per session,

or

2. Reject one of the login attempts early with a clear warning,

or

3. Provide configuration options to manage state per request or store state per tab.

(Currently UAA supports only one login transaction per session.)

strehle:
If IdP selection is not active, but all IdP are listed in login page, then the state and PKCE values are generated with each request and stored in the current HTTP session. If there are parallel action, e.g. different browser tabs, then this can lead to login issues.

Refactor so that these states are created once per Http session and re-use them in other requests, e.g. browser tabs

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthProviderConfigurator.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthProviderConfigurator.java
@@ -82,8 +82,7 @@ public class ExternalOAuthProviderConfigurator implements IdentityProviderProvis
         var responseType = URLEncoder.encode(definition.getResponseType(), StandardCharsets.UTF_8);
         var relyingPartyId = definition.getRelyingPartyId();
 
-        var state = generateStateParam();
-        SessionUtils.setStateParam(request.getSession(), SessionUtils.stateParameterAttributeKeyForIdp(idpOriginKey), state);
+        var state = SessionUtils.setStateParam(request.getSession(), SessionUtils.stateParameterAttributeKeyForIdp(idpOriginKey), generateStateParam());
 
         UriComponentsBuilder uriBuilder = UriComponentsBuilder
                 .fromUriString(idpUrlBase)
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/cloudfoundry/identity/uaa/util/SessionUtils.java
```diff
@@ -56,8 +57,13 @@ public static UaaAuthentication getForcePasswordExpiredUser(HttpSession session)
...
     }
 
-    public static void setStateParam(HttpSession session, String stateParamKey, String state) {
-        session.setAttribute(stateParamKey, state);
+    public static String setStateParam(HttpSession session, String stateParamKey, String state) {
+        if (session.isNew() || session.getAttribute(stateParamKey) == null) {
+            session.setAttribute(stateParamKey, state);
+            return state;
+        } else {
+            return session.getAttribute(stateParamKey) instanceof String existingState ? existingState : null;
```
adrianhoelzl-sap, line 65: Can we add a comment or change the method's name to represent this behavior? If a value already exists, we do not update it.

server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/ExternalOAuthProviderConfigurator.java
```diff
@@ -96,9 +95,8 @@ public String getIdpAuthenticationUrl(
         // https://docs.spring.io/spring-security/site/docs/5.3.1.RELEASE/reference/html5/#initiating-the-authorization-request
         if (isPkceNeeded(definition)) {
             var pkceVerifier = new S256PkceVerifier();
-            var codeVerifier = generateCodeVerifier();
-            var codeChallenge = pkceVerifier.compute(codeVerifier);
-            SessionUtils.setStateParam(request.getSession(), SessionUtils.codeVerifierParameterAttributeKeyForIdp(idpOriginKey), codeVerifier);
+            var codeVerifier = SessionUtils.setStateParam(request.getSession(), SessionUtils.codeVerifierParameterAttributeKeyForIdp(idpOriginKey), generateCodeVerifier());
+            var codeChallenge = SessionUtils.setStateParam(request.getSession(), SessionUtils.codeChallengeParameterAttributeKeyForIdp(idpOriginKey), pkceVerifier.compute(codeVerifier));
```
adrianhoelzl-sap, line 99: Can we replace `var` with the concrete types here?

fhanik:
Suggesting to close this one, I am working on https://github.com/cloudfoundry/uaa/pull/3951

@strehle

PR #3718, opened 2026-01-15
fix: include NameID Format in SAML LogoutRequest

mikeroda:
Shibboleth generally requires that the NameID element in a SAML LogoutRequest matches the NameID issued during the initial authentication, including its Format attribute. 

Follows Spring documentation for customizing the SAML2 LogoutRequest resolution:

https://github.com/spring-projects/spring-security/blob/5a7d93ee3b6658fbd03f0f57daeb42d3f382016d/docs/modules/ROOT/pages/servlet/saml2/logout.adoc#customizing-saml2logoutrequest-resolution


duanemay (approved):

duanemay:
@mikeroda Thank you for the contribution

PR #3720, opened 2026-01-17
Reduce private_key_jwt expiry

strehle:
With private_key_jwt we create a client_assertion and use it as replacement for the secret towards another IDP for authentication.

Reason to reduce is that Auth0 has the max exp of 5 min.
https://auth0.com/docs/get-started/authentication-and-authorization-flow/authenticate-with-private-key-jwt


duanemay (approved):
updated asserts

strehle:
> updated asserts

thanks, nice

PR #3727, opened 2026-01-28
fix: log at debug when openid scope not included

mikeroda:
- Should be able to request an access token without openid scope (i.e. ID token not requested) and not log an info-level message. Make the log debug instead.


strehle (commented):
from my side ok if you dont want the id token

duanemay (approved):

strehle (approved):

PR #3730, opened 2026-01-28
Path based Identity Zone

### LINKED ISSUE #3729: Path based Identity Zone
### Proposal

Rather than using subdomains, for example `somedomain.login.system.com`,  this issue proposes using a path `/z/{subdomain}/oauth/token`

### What this actually means

Implementation is 100% backwards compatible. 

Paths are used to **express** an identity zone, they are not a new definition of an identity zone. 

Tokens issued via subdomains (Host header) and tokens issued via path based identity zones have the same issuer, ie tokens do not adjust claim values.

### Requirements

1. Zone path will only work when used against the default domain. If a zone path is added to a subdomain of the UAA, it will return a 400 error.
2. subdomain based zones will continue to work
3. There is no zone path for the default zone, unless the path is zone-id rather than subdomain
4. Cookies must have the path set correctly 
5. Cookie flapping can occur, two cookies with JSESSIONID being sent, so cookie filtering must occur on the server

### Example Test

```java
public class ZonePathTokenMockMvcTests extends AbstractTokenMockMvcTests {

    private enum ZoneResolutionMode {
        SUBDOMAIN(subdomain -> get("/oauth/token")
                .with(new SetServerNameRequestPostProcessor(subdomain + ".localhost"))
        ),
        ZONE_PATH(subdomain -> get("/z/{subdomain}/oauth/token", subdomain));

        private final Function<String, MockHttpServletRequestBuilder> requestBuilder;

        ZoneResolutionMode(Function<String, MockHttpServletRequestBuilder> requestBuilder) {
            this.requestBuilder = requestBuilder;
        }

        MockHttpServletRequestBuilder createRequestBuilder(String subdomain) {
            return requestBuilder.apply(subdomain);
        }
    }

    @ParameterizedTest
    @EnumSource(ZoneResolutionMode.class)
    void clientCredentialsGrant(ZoneResolutionMode mode) throws Exception {
        String subdomain = "testzone" + generator.generate().toLowerCase();
        IdentityZone testZone = setupIdentityZone(subdomain);
        IdentityZoneHolder.set(testZone);
        setupIdentityProvider(OriginKeys.UAA);

        String scopes = "uaa.admin";

        String clientId = "testclient" + generator.generate();
        setUpClients(clientId, scopes, scopes, "client_credentials", true, null, null);

        IdentityZoneHolder.clear();

        String tokenResult = mockMvc.perform(mode.createRequestBuilder(subdomain)
                        .accept(APPLICATION_JSON)
                        .contentType(APPLICATION_FORM_URLENCODED)
                        .param(OAuth2Utils.GRANT_TYPE, GRANT_TYPE_CLIENT_CREDENTIALS)
                        .param(OAuth2Utils.RESPONSE_TYPE, "token")
                        .param(OAuth2Utils.CLIENT_ID, clientId)
                        .param(CLIENT_SECRET, SECRET)
                        .param(OAuth2Utils.REDIRECT_URI, TEST_REDIRECT_URI))
                .andExpect(status().isOk())
                .andExpect(jsonPath("$.access_token").isNotEmpty())
                .andReturn().getResponse().getContentAsString();
        Map<String, Object> response = JsonUtils.readValueAsMap(tokenResult);
        Jwt tokenClaims = JwtHelper.decode((String) response.get("access_token"));
        Assertions.assertThat(tokenClaims.getClaimSet().getStringClaim(ISS))
                .isEqualTo("http://%s.localhost:8080/uaa/oauth/token".formatted(testZone.getSubdomain()));
    }
}
```

### Decisions

1. Should the path use subdomain or zone-id? `/z/{subdomain}/oauth/token` versus `/z/{zone-id}/oauth/token`? Or `/zid/{zone-id}/` and `/zs/{subdomain}/` to support both?
2. Do we support `/z/{subdomain}/` everywhere, or should we only support it for UI (cookie based) and simply rewrite the URLs for all other places?
3. Should the JSESSIONID cookie name be different based on domain? example JSESSIONID-foobar for the foobar domain? (special characters, encoding)

fhanik:
Feature is by default not enabled (see uaa.yml)

```yaml
zones:
  paths:
    enabled: false
```

Code Review of Feature: [zone-path-feature.md
](https://github.com/fhanik/uaa/blob/feature/path-based-zones/docs/zone-path-feature.md)

Session Logic: [zone-session-implementation.md](https://github.com/fhanik/uaa/blob/feature/path-based-zones/docs/zone-session-implementation.md)

See https://github.com/cloudfoundry/uaa/issues/3729


server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginSecurityConfiguration.java
```diff
@@ -523,26 +523,32 @@ UaaFilterChain uiSecurity(
...
+                    auth.requestMatchers(HttpMethod.OPTIONS, "/logout.do", "/z/*/logout.do").permitAll();
                     auth.requestMatchers("/**").fullyAuthenticated();
                 })
                 .formLogin(login -> {
                     login.loginPage("/login");
                     login.usernameParameter("username");
                     login.passwordParameter("password");
+                    // Support both /login.do and /z/{subdomain}/login.do for zone path-based authentication
                     login.loginProcessingUrl("/login.do");
                     login.defaultSuccessUrl("/"); // TODO is this exactly the same?
```
fhanik: This should redirect to `/z/{subdomain}`

docs/code-review-zone-path-feature.md
```diff
@@ -0,0 +1,212 @@
+# Code Review: Zone Path-Based Multi-Tenancy
```
adrianhoelzl-sap: I would propose to remove the "Code Review" here and in the file name

docs/code-review-zone-path-feature.md
```diff
@@ -0,0 +1,212 @@
...
+
+**Base commit:** `7d0802131af7d0d2401c080be5a47d546d8dce43`
```
adrianhoelzl-sap: Please also remove these lines

server/src/main/java/org/cloudfoundry/identity/uaa/util/UaaUrlUtils.java
```diff
@@ -80,6 +86,18 @@ private static UriComponentsBuilder getURIBuilder(
         return builder;
     }
 
+    private static boolean isZoneInRequestPath() {
+        try {
+            if (RequestContextHolder.getRequestAttributes() instanceof ServletRequestAttributes attrs) {
+                String contextPath = attrs.getRequest().getContextPath();
```
adrianhoelzl-sap: Can `getRequest()` return `null` and cause a NPE?

server/src/main/java/org/cloudfoundry/identity/uaa/util/UaaUrlUtils.java
```diff
@@ -80,6 +86,18 @@ private static UriComponentsBuilder getURIBuilder(
...
+                return contextPath != null && contextPath.contains("/z/");
```
adrianhoelzl-sap: Would it make sense to introduce a constant for the `z` path segment?

adrianhoelzl-sap: Is `contains` sufficiently precise here? Shouldn't we check for the path starting with `z`?

server/src/main/java/org/cloudfoundry/identity/uaa/login/ZoneAwareStaticResourceLinkBuilder.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+public class ZoneAwareStaticResourceLinkBuilder extends StandardLinkBuilder {
+
+    @Override
+    protected String computeContextPath(IExpressionContext context, String base, Map<String, Object> parameters) {
+        if (!(context instanceof IWebContext)) {
+            return super.computeContextPath(context, base, parameters);
+        }
+        var exchange = ((IWebContext) context).getExchange();
+        if (!(exchange instanceof IServletWebExchange)) {
```
adrianhoelzl-sap: We can get rid of the cast later on if we use `instanceof IServletWebExchange exchange` here

server/src/main/java/org/cloudfoundry/identity/uaa/zone/IdentityZoneResolvingFilter.java
```diff
@@ -46,36 +48,82 @@ public IdentityZoneResolvingFilter(final IdentityZoneProvisioning dao) {
...
+        }
+        return subdomainFromHost;
+    }
 
-            request.setAttribute("error_message_code", "zone.not.found");
-            response.sendError(HttpServletResponse.SC_NOT_FOUND, "Cannot find identity zone for subdomain " + subdomain);
+    /**
+     * Looks up identity zone by subdomain. On generic exception sends 500 and returns null; caller should check response.isCommitted().
+     */
+    private IdentityZone resolveZoneBySubdomain(String subdomain, HttpServletRequest request, HttpServletResponse response) throws IOException {
```
adrianhoelzl-sap: The `request` parameter is unused here

server/src/main/java/org/cloudfoundry/identity/uaa/zone/IdentityZoneResolvingFilter.java
```diff
@@ -84,7 +132,28 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
...
+    /**
+     * If context path ends with /z/{subdomain}, return the subdomain; otherwise null.
+     * Supports both /z/myzone and /uaa/z/myzone. Allows zone resolution when the request
+     * was already rewritten (e.g. tests simulating post-rewrite with context path set).
+     */
+    private String extractSubdomainFromContextPath(String contextPath) {
+        if (!StringUtils.hasText(contextPath) || !contextPath.contains(ZONE_PATH_PREFIX)) {
+            return null;
+        }
+        int idx = contextPath.lastIndexOf(ZONE_PATH_PREFIX);
```
adrianhoelzl-sap: What if I perform a request to an endpoint that contains a user-defined ID that is set to the value "z", e.g., `/z/some-zone/oauth/clients/z/secret`? Would we pick up "secret" as the zone subdomain?

server/src/main/java/org/cloudfoundry/identity/uaa/zone/ZoneContextPathSessionFilter.java
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * <p>After the filter chain completes, flushes all sub-session attribute maps back to the
+ * container session via {@code setAttribute} so that Spring Session's dirty-tracking
+ * detects the changes and persists them.
+ */
+public class ZoneContextPathSessionFilter extends OncePerRequestFilter {
+
+    public static final String BEAN_NAME = "zoneContextPathSessionFilter";
+
+    private static final String JSESSIONID = "JSESSIONID";
```
adrianhoelzl-sap: This constant seems to be defined more than once, can we streamline it to one definition?

server/src/main/java/org/cloudfoundry/identity/uaa/zone/ZonePathHttpSession.java
```diff
@@ -0,0 +1,111 @@
...
+                               String containerSessionAttributeName) {
+        this.containerSession = containerSession;
+        this.attributes = attributes;
+        this.containerSessionAttributeName = containerSessionAttributeName;
+        this.sessionIdSuffix = sessionIdSuffixFromContextPath(contextPathKey);
+    }
+
+    private static String sessionIdSuffixFromContextPath(String contextPathKey) {
+        if (contextPathKey == null || contextPathKey.isEmpty()) {
+            return "default";
```
adrianhoelzl-sap: Can we introduce a constant for this at a meaningful location?

server/src/main/java/org/cloudfoundry/identity/uaa/zone/ZonePathContextRewritingFilter.java
```diff
@@ -0,0 +1,368 @@
...
+            return;
+        }
+
+        //make sure we have the {subdomain} element after /z/
+        String subdomain = extractSubdomainFromPath(pathAfterContext);
+        if (!hasText(subdomain)) {
+            response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Invalid /z/ URL");
+            return;
+        }
+        if (RESERVED_SUBDOMAIN.equalsIgnoreCase(subdomain)) {
```
adrianhoelzl-sap: Do we also reject this value during zone creation?

server/src/main/java/org/cloudfoundry/identity/uaa/zone/ZonePathContextRewritingFilter.java
```diff
@@ -0,0 +1,368 @@
...
+        if ("/".equals(pathAfterZonePrefix) && pathAfterContext.endsWith("/")) {
+            // /z/{subdomain}/ → servlet path "", pathInfo "/"
+            servletPath = "";
+            pathInfo = "/";
+        } else {
+            // /z/{subdomain} (no trailing) or /z/{subdomain}/something → pathInfo null
+            servletPath = pathAfterZonePrefix;
+            pathInfo = null;
+        }
+        PathResult pathResult = new PathResult(servletPath, pathInfo);
```
adrianhoelzl-sap: We can directly return the value instead of declaring a var

server/src/main/java/org/cloudfoundry/identity/uaa/zone/ZonePathContextRewritingFilter.java
```diff
@@ -0,0 +1,368 @@
...
+
+    private record PathResult(String servletPath, String pathInfo) {
+    }
+
+    private String getNewContextPath(String contextPath, String subdomain) {
+        String baseContext =
+                (contextPath != null && contextPath.endsWith("/")) ?
+                        contextPath.substring(0, contextPath.length() - 1) :
+                        contextPath;
+        String newContextPath = (baseContext != null ? baseContext : "") + ZONE_PATH_PREFIX + subdomain;
```
adrianhoelzl-sap: Please directly return the result

server/src/main/java/org/cloudfoundry/identity/uaa/zone/ZonePathContextRewritingFilter.java
```diff
@@ -0,0 +1,368 @@
...
+        return newContextPath;
+    }
+
+    private String getPathAfterZonePrefix(String pathAfterContext, String subdomain) {
+        String pathAfterZonePrefix = pathAfterContext.substring(ZONE_PATH_PREFIX.length() + subdomain.length());
+        if (!pathAfterZonePrefix.startsWith("/")) {
+            pathAfterZonePrefix = "/" + pathAfterZonePrefix;
+        }
+        if (pathAfterZonePrefix.isEmpty()) {
```
adrianhoelzl-sap: The previous condition includes the empty case and sets the var's value to a non-empty string; we therefore never reach this condition's body

server/src/main/java/org/cloudfoundry/identity/uaa/zone/ZonePathContextRewritingFilter.java
```diff
@@ -0,0 +1,368 @@
...
+    private String getNewContextPath(String contextPath, String subdomain) {
+        String baseContext =
+                (contextPath != null && contextPath.endsWith("/")) ?
+                        contextPath.substring(0, contextPath.length() - 1) :
+                        contextPath;
+        String newContextPath = (baseContext != null ? baseContext : "") + ZONE_PATH_PREFIX + subdomain;
+        return newContextPath;
+    }
+
+    private String getPathAfterZonePrefix(String pathAfterContext, String subdomain) {
```
adrianhoelzl-sap, line 166: This method can be `static`

server/src/main/java/org/cloudfoundry/identity/uaa/zone/ZonePathContextRewritingFilter.java
```diff
@@ -0,0 +1,368 @@
...
+        if (!pathAfterZonePrefix.startsWith("/")) {
+            pathAfterZonePrefix = "/" + pathAfterZonePrefix;
+        }
+        if (pathAfterZonePrefix.isEmpty()) {
+            pathAfterZonePrefix = "/";
+        }
+        return pathAfterZonePrefix;
+    }
+
+    protected String getPathAfterContext(String requestURI, String contextPath) {
```
adrianhoelzl-sap, line 174: This method can be `static`

docs/zone-session-implementation.md
```diff
@@ -0,0 +1,242 @@
...
+single JSESSIONID cookie, a single underlying container/Spring Session, and
+**server-side namespacing by context path**. Each zone's session attributes live
+in an isolated `ConcurrentHashMap` stored as a single attribute on the container
+session, keyed by the request's context path.
+
+### Key classes
+
+| Class | Role |
+|---|---|
+| `ZonePathContextRewritingFilter` | Rewrites `/uaa/z/tenant/login` so context path becomes `/uaa/z/tenant` and servlet path becomes `/login`. Also wraps the response to normalize cookie paths. |
```
adrianhoelzl-sap, line 31: Shouldn't it be called `ZoneContextPathRewritingFilter` then?

server/src/main/java/org/cloudfoundry/identity/uaa/zone/ZonePathContextRewritingFilter.java
```diff
@@ -0,0 +1,368 @@
...
+        } else {
+            // /z/{subdomain} (no trailing) or /z/{subdomain}/something → pathInfo null
+            servletPath = pathAfterZonePrefix;
+            pathInfo = null;
+        }
+        PathResult pathResult = new PathResult(servletPath, pathInfo);
+        return pathResult;
+    }
+
+    private record PathResult(String servletPath, String pathInfo) {
```
adrianhoelzl-sap, line 155: Can we have a more descriptive name and/or comment for `pathInfo`?

server/src/main/java/org/cloudfoundry/identity/uaa/zone/ZoneContextPathSessionRequestWrapper.java
```diff
@@ -0,0 +1,114 @@
...
+        return snapshot;
+    }
+
+    private void restoreOurAttributes(HttpSession containerSession, Map<String, Object> snapshot) {
+        snapshot.forEach(containerSession::setAttribute);
+    }
+
+    private String contextPathKey() {
+        String cp = getContextPath();
+        return (cp != null && !cp.isEmpty()) ? cp : "";
```
adrianhoelzl-sap: Consider using `UaaStringUtils.EMPTY_STRING` at all locations where an empty string is intended to be used.

server/src/main/java/org/cloudfoundry/identity/uaa/zone/ZoneContextPathSessionRequestWrapper.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    @Override
+    public HttpSession getSession(boolean create) {
+        HttpSession containerSession = getDelegateRequest().getSession(create);
+        if (containerSession == null) {
+            return null;
+        }
+
+        String contextPathKey = contextPathKey();
```
adrianhoelzl-sap, line 64: Can we have a comment here that at this point, we expect the context path to be overwritten by `ZonePathContextRewritingFilter` in case the path-based zone change is used?

server/src/main/java/org/cloudfoundry/identity/uaa/zone/ZoneContextPathSessionFilter.java
```diff
@@ -0,0 +1,75 @@
...
+            if (names.nextElement().startsWith(prefix)) {
+                return; // at least one context-path sub-session still present
+            }
+        }
+        String path = request.getContextPath();
+        if (path == null || path.isEmpty()) {
+            path = "/";
+        }
+        HttpServletResponse rawResponse = (HttpServletResponse) response.getResponse();
+        rawResponse.addHeader("Set-Cookie", JSESSIONID + "=; Max-Age=0; Path=" + path);
```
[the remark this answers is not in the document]
fhanik: This is erasing the cookie from the browser. HttpOnly is irrelevant

server/src/main/java/org/cloudfoundry/identity/uaa/zone/ZoneContextPathSessionResponseWrapper.java
```diff
@@ -0,0 +1,77 @@
...
+    private HttpServletResponse getHttpResponse() {
+        return (HttpServletResponse) getResponse();
+    }
+
+    @Override
+    public void addCookie(Cookie cookie) {
+        if (wouldClearJSessionIdCookie(cookie.getName(), cookie.getValue(), cookie.getMaxAge())) {
+            return;
+        }
+        getHttpResponse().addCookie(cookie);
```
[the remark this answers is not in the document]
fhanik: What?

strehle (approved):

PR #3731, opened 2026-01-29
feat: omit explicit DB Statement for health check 

tack-sap:
With #2718 and #2763 we introduced an (asynchronous) check to the Database via a simple query to the health check of the UAA. As HSQLDB does not support the simple query "SELECT 1;" we used "SELECT 1 from identity_zone;" as the query that is executed in the health check. Usually this is still simple enough and it remains to be a good choice as default statement. 

However in some specific scenarios this statement produces too much work for a simple health check. Especially in an environment with a lot of identity_zones and many uaa instances that do such a health check in regular interval, we saw that it produces unnecessary load on the database. 

To avoid this unnecessary load on the database, we would like to make the statement that is used for the health check of the DB configurable with this PR. The default will stay the same, as this statement will work on all supported vendors. But for scenarios where you don't want this default statement, you have the possibility to change the statement to whatever suits your needs better. The property shares the same prefix as the rate of the DB checks, so you have both configurations in the same place.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/health/HealthzEndpoint.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/health/HealthzEndpoint.java
@@ -24,11 +24,14 @@ public class HealthzEndpoint {
     private volatile boolean stopping;
     private volatile Boolean wasLastConnectionSuccessful = null;
     private DataSource dataSource;
+    private String healthCheckStatement;
+
 
     public HealthzEndpoint(
             @Value("${uaa.shutdown.sleep:10000}") final long sleepTime,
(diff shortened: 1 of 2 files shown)


duanemay:
As discussed there is already a validationQuery in org.cloudfoundry.identity.uaa.db.DatabasePlatform, which is used in the datasource when getting a connection.

tack-sap:
@fhanik @duanemay As discussed I changed the implementation to rely on the validationQuery and testOnBorrow configuration of the DataSource. Hence the statement in the HealthzEndpoint is no longer needed and we only establish a connection to check the database.

adrianhoelzl-sap (approved):

PR #3737, opened 2026-02-05
CSS Trimming

lhardt:
### tl;dr

- Removed unused classes from `application.css`
- Created and adopted a minified version of `application.css` (101650 bytes , compared to the original's 235830 bytes, a ~56% reduction).
- Raised questions about OSS attribution 

### longer version

This PR is intended to expand on #1414 , by minifying the CSS file. Instead of minifying the CSS file directly, I preserved the original one, and created an `application.min.css` file. For the record, I used [this](https://github.com/tdewolff/minify) minification tool.

Inspecting a little further, It also seems to me that the `application.css` file composed of: 

- an old version of [normalize.css](https://github.com/necolas/normalize.css), 
- what appears to be bootstrap v2 or v3, 
- many rules imported from [pivotal-one-styles](https://github.com/vmware-archive/pivotal-one-styles); 
- some custom rules implemented "by hand" to satisfy business needs.

It is essentially overkill, accumulated from another project, when this [login-server commit](https://github.com/cloudfoundry-attic/login-server/commit/10d5d7c39128e9d4d2bc2343bc4960bc3fd45423#diff-06b290ffee12f6fd8375b0c6043fc2bcff7bc3e4284b21557697a020f15610e1) stripped away _Sass_ from the build process and merged all CSS files into what today is `application.css`.   Finally, it ended up here after [the contents of login-served were merged](https://github.com/lhardt/uaa/commit/327fbdc2c2f8595b1fbc09f5925c907b42f2d45a) into this repo.

The lack of attribution when squashing all Sass might be concerning, and is a possible license violation of the abovementioned CSS libraries. I removed a sizeable part of the "bootstrappy" code that appeared to be unused, as well as `gravatar` related classes, which could mitigate library risk.


server/src/main/resources/templates/web/layouts/main.html
```diff
@@ -22,7 +22,7 @@
     <meta name='keywords' content=""/>
     <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
     <link href="/vendor/font-awesome/css/font-awesome.min.css" th:href="@{/vendor/font-awesome/css/font-awesome.min.css}" rel="stylesheet"/>
-    <link href="/resources/oss/stylesheets/application.css" th:href="@{${assetBaseUrl}+'/stylesheets/application.css'}" rel="stylesheet"/>
+    <link href="/resources/oss/stylesheets/application.min.css" th:href="@{${assetBaseUrl}+'/stylesheets/application.min.css'}" rel="stylesheet"/>
```
lhardt: I'm not sure here if I understand this line's usage correctly.  

If the host provides external branding and provides their on `assetBaseUrl`,  do we expect all the `/resources/oss` contents to be present there or do we expect the host to provide their own custom `application.css`?

A remote possibility is, if the host created their own application.css (but not a minified version) this would ensure the css file is not found.

duanemay:
@lhardt Thank you, I think having a step to run minify, during the build process would be a better approach. The minified version is hard for a human to read to make edits, and it is unclear how to regenerate it when changes are made. Doing it in the build solves this.

lhardt:
In general, true, but given this file is very seldom changed (3 commits in 6 years) I also imagine it would add more maintenance overhead (because of longer pipeline, and version bumping), consequently saving less time than it would consume.

I could implement it either way, I just want the page to load slightly faster :) 

You could also have a github post-action to recommit the minified version in case the original changes. Then it wouldn't impact the main build time / tools.

duanemay:
@lhardt To expedite, this effort could be divided into two separate Pull Requests (PRs):

- Quick Size Fix (PR 1): Focus on immediate size reduction by removing unused CSS classes from application.css. This is expected to deliver a significant and fast improvement.
- Add CSS Minification (PR 2): Integrate CSS minification into the existing build process.

While running CSS minification as a post-build GitHub action saves overall build time, it complicates the CSS testing loop. To address this, we must either clearly document the manual execution steps for the minification process or, preferably, implement a mechanism (such as using file timestamps) to ensure the build only runs the minification step when necessary.

lhardt:
Done so. 

Now the stats are:
- Previous application.css size: 1691 rules,  235830 B = **230 KB**
- Trimmed application.css size: 855 rules,   104879 B = **102 KB**

If we were to minify this, we would achieve 64902 B = **63 KB** . 
This is far from the ideal < 14 KB but much better than current and helps the developer identify much more clearly what styles are affecting the pages

duanemay (approved):

PR #3767, opened 2026-03-10
Update commons-codec version to 1.21.0

strehle:


strehle:
maybe let dependabot do the job

PR #3784, opened 2026-03-16
Remove unused Kubernetes configurations and custom matchers

duanemay:


duanemay:
concourse-ci/status test fails since it is running the run-k8s-template-tests which no longer exist. We will have to remove those form concourse once we decide to merge

strehle (approved):
ok from my side we should remove because with this kind project we have now a k8s CF version which is really easy to use... e.g.
https://github.com/cloudfoundry/kind-deployment

strehle (dismissed):

adrianhoelzl-sap (approved):

PR #3789, opened 2026-03-17
Change userDatabase autowired type to UaaUserDatabase

gdgenchev:


strehle:
@gdgenchev is there issue or simply because you need it in your fork ?

gdgenchev:
No issue, but the current type is too restrictive, as it uses the impl instead of the interface, while the beans that are created accept the interface. And yes, this caused issues at our side when migrating to spring boot and consuming `SpringServletXml..` beans from cfuaa.

strehle (approved):
ok from my side @duanemay  colleagues overwrite this with a HANA DB in a fork they maintain

duanemay:
Works for me!

Issue #3692, opened 2025-12-03
[Feature Request] Configurable pre-login notice / consent page in UAA


duanemay:
## Summary

We have a requirement to show a **pre-login notice** on the UAA login flow, before the user can enter credentials or proceed to authenticate.

Today, UAA allows some limited customization of the login UI via branding options (banner text/logo, footer legal text, footer links such as Support / Terms of Service, and basic theming). However, there is no built-in way to show a **blocking pre-login notice** or **consent message** that must be acknowledged before login can proceed.

We would like UAA to support a configurable **pre-login notice / consent step** that can be used for things like:

- Terms of Service
- Acceptable Use Policy
- End-User License Agreement
- Service Terms & Conditions
- Privacy Notice
- Data Processing Notice
- Data Collection Consent
- Cookie & Tracking Disclosure (if applicable)
- Security Warning
- Authorized Use Notice
- Confidential Systems Notice
- Important Notice
- Before You Continue
- Usage Guidelines
- Organization Policy

This feature should be opt-in and preserve current login behavior by default.

PR #3792, opened 2026-03-18
Add an optional consent modal before login

duanemay:
Implements #3692


server/src/test/java/org/cloudfoundry/identity/uaa/login/LoginInfoEndpointTests.java
```diff
@@ -1807,4 +1809,80 @@ private IdentityProvider<AbstractExternalOAuthIdentityProviderDefinition> mockOa
...
+        loginConsent.setText("You are accessing a System that is provided for authorized use only.");
+        loginConsent.setAcceptButtonText("I Accept");
+        loginConsent.setDeclineLink("https://www.cloudfoundry.org");
+        loginConsent.setConsentValidDuration("12h");
+        branding.setLoginConsent(loginConsent);
+
+        IdentityZoneConfiguration config = IdentityZoneHolder.get().getConfig();
+        config.setBranding(branding);
+        IdentityZoneHolder.get().setConfig(config);
+
```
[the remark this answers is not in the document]
duanemay: Consistent with other tests

server/src/main/java/org/cloudfoundry/identity/uaa/login/LoginConsentHashUtil.java
```diff
@@ -0,0 +1,115 @@
...
+     */
+    public static String calculateConsentHash(LoginConsent consent) {
+        if (consent == null || !consent.isEnabled()) {
+            return null;
+        }
+
+        String title = StringUtils.hasText(consent.getTitle()) ? consent.getTitle() : "";
+        String text = StringUtils.hasText(consent.getText()) ? consent.getText() : "";
+        
+        String content = title + "|" + text;
```
adrianhoelzl-sap: In theory, there is the case where the values

title: 'abc|def'
content: 'ghi'

is changed to

title: 'abc'
content: 'def|ghi'

and would be considered equal.

(But I guess we can ignore that case)

uaa/src/main/resources/uaa.yml
```diff
@@ -516,6 +516,21 @@ login:
...
+#      text: |
+#        <p>You are accessing a site that is provided for authorized use only.</p>
+#        <p>By using this site, you consent to the following conditions:</p>
+#        <ul>
+#          <li>You are responsible for compliance with all applicable laws and regulations.</li>
+#          <li>You are responsible for obtaining all necessary permissions to access the site.</li>
+#          <li>All communications are subject to routine monitoring,
+#             interception, and search, and may be disclosed or used for any purpose.</li>
+#        </ul>
+#      acceptButtonText: "I Accept"
```
adrianhoelzl-sap, line 531: Do we also need a configurable text for the decline button?

adrianhoelzl-sap (approved):

PR #3815, opened 2026-04-07
Handle custom issuer configuration in `getTokenEndpointUrl` method

duanemay:
Inconsistent handling of the issuerUrl. constructTokenEndpointUrl expects issuer, but KeyInfoService was passing in the uaaBaseUrl

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/KeyInfoService.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/KeyInfoService.java
@@ -16,6 +16,7 @@ package org.cloudfoundry.identity.uaa.oauth;
 
 import org.cloudfoundry.identity.uaa.impl.config.LegacyTokenKey;
 import org.cloudfoundry.identity.uaa.util.UaaTokenUtils;
+import org.cloudfoundry.identity.uaa.zone.IdentityZone;
 import org.cloudfoundry.identity.uaa.zone.IdentityZoneConfiguration;
 import org.cloudfoundry.identity.uaa.zone.IdentityZoneHolder;
 import org.cloudfoundry.identity.uaa.zone.TokenPolicy;
@@ -24,6 +25,7 @@ import org.springframework.util.StringUtils;
 import java.net.URISyntaxException;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 
 import static org.cloudfoundry.identity.uaa.util.UaaUrlUtils.addSubdomainToUrl;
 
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/cloudfoundry/identity/uaa/oauth/KeyInfoService.java
```diff
@@ -95,6 +97,10 @@ private String getActiveKeyId() {
     }
 
     public String getTokenEndpointUrl() throws URISyntaxException {
-        return UaaTokenUtils.constructTokenEndpointUrl(uaaBaseURL, IdentityZoneHolder.get());
+        IdentityZone identityZone = IdentityZoneHolder.get();
+        String issuer = Optional.ofNullable(identityZone.getConfig())
+                .map(IdentityZoneConfiguration::getIssuer)
```
strehle, line 102: I am not sure, but I think issuer configuration can be without https, so I would ensure that https:// is set

duanemay: The default for the uaaUrl, which this falls back to, is not https:

public static final String DEFAULT_UAA_URL = "http://localhost:8080/uaa";


if you mean the protocol is not required, in constructTokenEndpointUrl it uses issuer already for the non-default zones, and it looks like both paths call validateIssuer which ensures it is a proper URL.

strehle: oh my god... ok...

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/token/KeyInfoServiceTests.java
```diff
@@ -152,12 +152,20 @@ void activeKeyFallsBackToLegacyKey() {
...
     @Test
-    void tokenEndpointUrl() throws URISyntaxException {
+    void tokenEndpointUrl_whenIssuerIsNull() throws URISyntaxException {
         configureDefaultZoneKeys(Collections.emptyMap());
 
         assertThat(keyInfoService.getTokenEndpointUrl()).isEqualTo("https://localhost/uaa/oauth/token");
     }
 
+    @Test
+    void tokenEndpointUrl_whenIssuerSet() throws URISyntaxException {
```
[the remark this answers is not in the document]
strehle: maybe check this

strehle (approved):
beside copilot ok for my side

server/src/test/java/org/cloudfoundry/identity/uaa/oauth/token/KeyInfoServiceTests.java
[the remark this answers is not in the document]
duanemay: Underscore seems common in the full codebase, just not within this test class. 
1618 test methods use underscores to separate the method name from the expected behavior or scenerio (e.g., methodName_expectedBehavior, methodName_scenerio)

PR #3827, opened 2026-04-10
Fix legacy SAML bootstrap with newer OpenSAML

strehle:
Fix legacy SAML bootstrap with newer OpenSAML

Depends on was before only on relyingPartyRegistrationRepository now pinned it to classes so that it should be done always before touching any SAML logic


duanemay (approved):

fhanik (approved):

strehle:
now reverted the revert and included this commit into PR: https://github.com/cloudfoundry/uaa/pull/3829
Close this because better go with https://github.com/cloudfoundry/uaa/pull/3829

PR #3829, opened 2026-04-11
Add OpenSaml work around for FIPS initialization

### LINKED ISSUE #3811: Decouple from opensaml 4.0.1
We PINNED the version to 4.0.1 because of hard depency to non FIPS BC, but UAA wants / needs FIPS complaint version

strehle:
Implement what suggested and ignore then the spring security exception. See

https://github.com/spring-projects/spring-security/issues/14656

other ref:
https://shibboleth.atlassian.net/wiki/spaces/DEV/pages/1159627167/FIPS

Finally. We can adopt and remove exception handling if https://github.com/spring-projects/spring-security/pull/19059 or a similar PR would be in spring security

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/IdentityZoneConfig.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/IdentityZoneConfig.java
@@ -4,29 +4,15 @@ import lombok.extern.slf4j.Slf4j;
 import org.bouncycastle.jcajce.provider.BouncyCastleFipsProvider;
 import org.cloudfoundry.identity.uaa.zone.IdentityZoneHolder;
 import org.cloudfoundry.identity.uaa.zone.IdentityZoneProvisioning;
-import org.opensaml.core.config.ConfigurationService;
-import org.opensaml.core.config.InitializationException;
-import org.opensaml.core.config.Initializer;
-import org.opensaml.security.config.GlobalNamedCurveRegistryInitializer;
-import org.opensaml.xmlsec.config.impl.DefaultSecurityConfigurationBootstrap;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.DependsOn;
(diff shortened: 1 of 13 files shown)


server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/IdentityZoneConfig.java
```diff
@@ -48,4 +62,28 @@ public IdentityZoneHolder.Initializer identityZoneHolderInitializer(IdentityZone
 
         return new IdentityZoneHolder.Initializer(provisioning, samlKeyManagerFactory);
     }
+
+    @Bean
+    public static Boolean setupOpenSaml() {
```
duanemay: Is there a better place to put setupOpenSaml than IdentityZoneConfig? SamlConfiguration seems more logical.

strehle: ok, done

duanemay (approved):

server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/SamlConfiguration.java
```diff
@@ -54,6 +69,31 @@ public class SamlConfiguration {
...
+    @Bean
+    public static Boolean setupOpenSaml() {
+        if (samlInitialized.compareAndSet(false, true)) {
+            Properties props = ConfigurationService.getConfigurationProperties();
+            props.put(CONFIG_PROPERTY_ECDH_DEFAULT_KDF, DefaultSecurityConfigurationBootstrap.PBKDF2);
+            Class<?> toSkip = GlobalNamedCurveRegistryInitializer.class;
+            ServiceLoader.load(Initializer.class).stream().filter((provider) -> provider.type() != toSkip).forEach((provider) -> init(provider));
+            try {
+                OpenSamlInitializationService.initialize();
+            } catch (NoClassDefFoundError | NoSuchMethodError e) {
```
fhanik, line 83: Do we need to log the warning here? Or can we always ignore it?

PR #3854, opened 2026-04-21
Migrate from AntPathRequestMatcher to PathPatternRequestMatcher

gdgenchev:
`AntPathRequestMatcher` was deprecated in Spring Security 6.x and later removed in Spring Security 7.x


duanemay (approved):

strehle (approved):
here some more context from gemeni
--- 
we use ** only at the end of pathes, thus ok
---
In Spring Security 6 and 7, PathPatternRequestMatcher is the intended modern successor to AntPathRequestMatcher, but it is not a 100% "drop-in" replacement due to architectural and syntax differences.
Here is the breakdown of why it isn't a direct replacement and what you need to watch out for:
1. Syntax Restrictions
The PathPattern parser used by the new matcher is more efficient but stricter than the old AntPathMatcher logic:
Wildcard placement: In AntPathRequestMatcher, you could use double wildcards anywhere (e.g., /api/**/details). In PathPatternRequestMatcher, the ** wildcard is only allowed at the end of a pattern.
Trailing Slashes: The new matcher is more rigid about how it handles trailing slashes depending on your Spring MVC configuration.

PR #3856, opened 2026-04-22
Migrate MediaType.sortByQualityValue() to local implementation

gdgenchev:
`MediaType.sortByQualityValue()` was deprecated in Spring 6 and later removed in Spring 7. Copy the sorting logic into a utility class to preserve content negotiation behavior that respects client quality value preferences from Accept headers.

Exact copy of Spring 6:

https://github.com/spring-projects/spring-framework/blob/9f431e2eac1b6d8d5ca385d0cc367bac94dd37e7/spring-web/src/main/java/org/springframework/http/MediaType.java#L906-L911

https://github.com/spring-projects/spring-framework/blob/9f431e2eac1b6d8d5ca385d0cc367bac94dd37e7/spring-web/src/main/java/org/springframework/http/MediaType.java#L927-L965

https://github.com/spring-projects/spring-framework/blob/9f431e2eac1b6d8d5ca385d0cc367bac94dd37e7/spring-web/src/test/java/org/springframework/http/MediaTypeTests.java#L390-L444 (+ some additional tests for the utility method)

This maintains backward compatibility.

---

The other approach would be to switch to https://github.com/spring-projects/spring-framework/blob/c6096ff6e5e7de92cf6a3466e211d72b401a0178/spring-core/src/main/java/org/springframework/util/MimeTypeUtils.java#L351-L358

but it will not be backward compatible and we will have to change some ITs that cover this.


fhanik (dismissed):

fhanik:
I'd backfill the test copilot is asking for, then we can merge

strehle:
alternative https://github.com/cloudfoundry/uaa/pull/3901

strehle (approved):
I have an alternative PR https://github.com/cloudfoundry/uaa/pull/3901 but not sure if this is 100% compatible, however others can decide which variant we should use. both allow spring 7

PR #3874, opened 2026-04-24
Replace deprecated check() with authorize() in AuthorizationManager

gdgenchev:
`AuthorizationManager.check()` was deprecated in Spring Security 6.5.x in favor of `authorize()` and later got removed in Spring Security 7.x.


gdgenchev:
Closing this, as it will anyway need migration in boot 4 - we have overridden the `check` in a few places and they cannot be removed pre-boot-4.

PR #3876, opened 2026-04-26
Replace deprecated APPLICATION_JSON_UTF8 import with local constant

gdgenchev:


strehle (approved):

duanemay:
Why not use the Spring value?

The official standard states that JSON must be encoded using UTF-8 for interoperability. Because of this, modern clients and servers are hardcoded to treat application/json as UTF-8. Adding ;charset=UTF-8 is now seen as "noise" because there is no other valid encoding for JSON.

[8.1](https://datatracker.ietf.org/doc/html/rfc8259#section-8.1).  Character Encoding

   JSON text exchanged between systems that are not part of a closed
   ecosystem MUST be encoded using UTF-8

strehle (approved):

PR #3878, opened 2026-04-28
Consolidate JWT keys usage stored in Clients for client authentication

fhanik:
# Pull request: Unify and persist client JWT config (`client_jwt_config` + `jwt_creds`)

## Summary

This change makes **private_key_jwt** trust for OAuth clients **consistent** across the client admin API, the in-memory `UaaClient` view, and the database. Trust can be expressed in more than one way (the `client_jwt_config` column, a top-level `jwt_creds` array, or nested / legacy shapes in `additional_information`). The UAA now **merges** these into a single `ClientJwtConfiguration` for create/update, **persists** the result on `PUT` (the `client_jwt_config` column is updated in the generic `updateClientDetails` path), and **normalizes** GET responses so `ClientDetailsModification` can expose a decomposed `jwks` / `jwks_uri` / `jwt_creds` view without duplicating raw keys in `additional_information`.

**Backward compatibility:** existing clients and payloads that still send `jwt_creds` (or related forms) at the top level or in `additional_information` continue to work; the server merges them. **New integrations** should **prefer a single top-level `client_jwt_config` JSON string** and avoid relying on parallel legacy fields.

## Motivation

- `UaaClient.getClientJwtConfiguration()` previously only read the `client_jwt_config` string and ignored trust carried only in `additional_information` / `jwt_creds`, so `private_key_jwt` could ignore configured trust.
- Create/update did not always fold and strip `jwt_creds` and nested `client_jwt_config` from the extra map, so responses and storage could disagree.
- `MultitenantJdbcClientDetailsService.updateClientDetails` did not include `client_jwt_config` in the UPDATE statement, so merged JWT could be **lost on PUT** while other columns updated.
- `ClientJwtConfiguration.merge` could yield `null` in edge cases and drop prior trust.

## What to review (suggested order)

1. **Behavior & API:** `UaaClient`, `ClientAdminEndpoints` + `ClientAdminEndpointsValidator`, `ClientDetailsModification`, `ClientJwtConfiguration`, `UaaClientDetails` copy, `MultitenantJdbcClientDetailsService`.
2. **Tests (unit / MockMvc):** validator, `UaaClientTest`, `ClientDetailsModificationTest`, `ClientJwtCredentialsEndpointMockMvcTests`, `JwtClientAuthenticationTest` (including multi-`jwt_cred` federated policy).
3. **Generated API docs (sources, not `build/`):** `ClientAdminEndpointDocs.java` (REST Docs field tables) and the **Clients** intro in `uaa/slateCustomizations/source/index.html.md.erb` (run `./gradlew generateDocs` locally; HTML is under `uaa/build/`, which is not committed).
4. **Design note:** [Optional] in-repo design notes under `ai/` (e.g. `CLIENT_JWT_CREDS.md`) if present, for before/after and `generateDocs` source locations.

**Integration tests under `uaa/.../integration/**` are intentionally not modified** (per project policy).

## Commits (typical structure)

- **test:** New/updated unit and MockMvc coverage for `jwt_creds` / `client_jwt` merge, admin API, and federated `private_key_jwt` (tests may be expected to fail without the next commit in a strict TDD workflow).
- **feat:** Production changes in `model/`, `server/…` (merge, validation, GET shape, JDBC update including `client_jwt_config`).
- **docs:** In-repo design notes, Slate + REST-Docs field descriptions (preferred `client_jwt_config` vs legacy `jwt_creds` / decomposed `jwks*`).

Here’s how it works in the code path that matters for **runtime** (`UaaClient.getClientJwtConfiguration()`), i.e. when building the single `ClientJwtConfiguration` the UAA uses (including for `private_key_jwt`).

## 1. Base: the stored column (`client_jwt_config`)

The process **always starts** from the formal field that backs the DB column: `UaaClientDetails.getClientJwtConfig()` (the `client_jwt` JSON string). It is parsed with `ClientJwtConfiguration.readValue(...)`. If that string is missing, you get an **empty** `ClientJwtConfiguration` and continue.

So the **first** source of truth is: **the `client_jwt_config` string on the client** (inline `jwks` / `jwks_uri` / embedded `jwt_creds` inside that JSON are all part of that one blob).

## 2. Then: top-level `jwt_creds` in `additionalInformation`

If `additionalInformation` contains the key `jwt_creds` (string or list), that is turned into a small `ClientJwtConfiguration` that **only** carries those credentials, then merged **on top of** the result from step 1:

```56:76:server/src/main/java/org/cloudfoundry/identity/uaa/client/UaaClient.java
    public ClientJwtConfiguration getClientJwtConfiguration() {
        // Start with configuration from clientJwtConfig field
        ClientJwtConfiguration result = Optional.ofNullable(clientJwtConfig)
                .map(ClientJwtConfiguration::readValue)
                .orElse(new ClientJwtConfiguration());
        // ...
            Object jwtCredsValue = additionalInformation.get(ClientJwtConfiguration.JWT_CREDS);
            if (jwtCredsValue != null) {
                // ...
                    result = ClientJwtConfiguration.merge(result, additionalConfig, false);
```

So **order of application** is: **column first**, then **top-level `jwt_creds` in additional**, with `overwrite = false` (union / add, not replace wholesale).

## 3. Then: nested `client_jwt_config` map + `jwt_creds` inside it

If `additionalInformation["client_jwt_config"]` is a **map** and has a nested `jwt_creds`, that is parsed the same way and merged again:

```81:94:server/src/main/java/org/cloudfoundry/identity/uaa/client/UaaClient.java
            Object clientJwtCredsValue = additionalInformation.get("client_jwt_config");
            if (clientJwtCredsValue instanceof Map clientJwtCredsMap) {
                Object nestedJwtCreds = clientJwtCredsMap.get(ClientJwtConfiguration.JWT_CREDS);
                if (nestedJwtCreds != null) {
                    // ...
                        result = ClientJwtConfiguration.merge(result, nestedConfig, false);
```

So the **sequence** is:

1. **`client_jwt_config` string (column)**  
2. **Top-level `jwt_creds` in `additionalInformation`**  
3. **Nested `jwt_creds` under `additionalInformation["client_jwt_config"]` (when that value is a map)**

All of those extra steps use `ClientJwtConfiguration.merge(..., false)`.

## 4. What `merge(..., false)` does (simplified)

With `overwrite == false`:

- **`jwks_uri`**: keeps the existing URI if both sides have ideas; otherwise fills in from the new side (see the `existingConfig.jwksUri != null ? existing : new` style branch in `merge`).
- **`jwks`**: **combines** key sets (union of keys, with rules if a key already exists).
- **`jwt_creds`**: **adds** credentials via `addJwtCredentials`, which **dedupes** by subject+issuer (not by your “which field came first” — so order of *fields* is above; order of *duplicate* creds is handled inside `addJwtCredentials`).

So you should not think of it as “try key A, then key B for signature” at the **storage** layer; it is “**build one merged `ClientJwtConfiguration`** in that order.” **Signature checking** for `private_key_jwt` then uses that merged object (e.g. federated path picks the `ClientJwtCredential` that matches the assertion’s `iss`/`sub`/`aud`).

## 5. Admin PUT path (different call site)

On **create/update**, `ClientAdminEndpoints` / `syncWithExisting` also merge **existing client from DB** with **incoming body** using `ClientJwtConfiguration.merge(..., false)` (and related helpers). That is the **request** merge (existing vs input), not the same three-step `UaaClient` list above, but the same `merge` semantics apply for combining two `ClientJwtConfiguration` instances.

**Bottom line:** The **column `client_jwt_config` is evaluated first** and forms the base; **additional** top-level `jwt_creds` and then **nested** `client_jwt_config`/`jwt_creds` are **layered on** in that order, each time with **non-destructive** merge (`overwrite = false`) unless a different code path explicitly uses `overwrite = true` (e.g. some update flows).

### DIFF
--- a/model/src/main/java/org/cloudfoundry/identity/uaa/client/UaaClientDetails.java
+++ b/model/src/main/java/org/cloudfoundry/identity/uaa/client/UaaClientDetails.java
@@ -99,6 +99,9 @@ public class UaaClientDetails implements ClientDetails {
         this.setScope(prototype.getScope());
         this.setResourceIds(prototype.getResourceIds());
         this.setAdditionalInformation(prototype.getAdditionalInformation());
+        if (prototype instanceof UaaClientDetails uaa) {
+            this.setClientJwtConfig(uaa.getClientJwtConfig());
+        }
     }
 
     public UaaClientDetails(String clientId, String resourceIds,
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/client/ClientAdminEndpoints.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/client/ClientAdminEndpoints.java
@@ -22,6 +22,7 @@ import org.cloudfoundry.identity.uaa.oauth.client.ClientConstants;
 import org.cloudfoundry.identity.uaa.oauth.client.ClientDetailsCreation;
 import org.cloudfoundry.identity.uaa.oauth.client.ClientDetailsModification;
 import org.cloudfoundry.identity.uaa.oauth.client.ClientJwtChangeRequest;
+import org.cloudfoundry.identity.uaa.oauth.client.ClientJwtCredential;
 import org.cloudfoundry.identity.uaa.oauth.client.SecretChangeRequest;
 import org.cloudfoundry.identity.uaa.provider.ClientAlreadyExistsException;
 import org.cloudfoundry.identity.uaa.provider.NoSuchClientException;
@@ -32,6 +33,7 @@ import org.cloudfoundry.identity.uaa.resources.ResourceMonitor;
 import org.cloudfoundry.identity.uaa.resources.SearchResults;
 import org.cloudfoundry.identity.uaa.resources.SearchResultsFactory;
 import org.cloudfoundry.identity.uaa.resources.SimpleAttributeNameMapper;
+import org.cloudfoundry.identity.uaa.util.JsonUtils;
 import org.cloudfoundry.identity.uaa.security.beans.SecurityContextAccessor;
 import org.cloudfoundry.identity.uaa.util.UaaPagingUtils;
 import org.cloudfoundry.identity.uaa.util.UaaStringUtils;
@@ -84,6 +86,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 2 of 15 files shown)


duanemay (dismissed):

duanemay (approved):

strehle (approved):
ok, my claude review create some improvements and I pushed them into UAA branch, so just see commit
https://github.com/cloudfoundry/uaa/commit/cbaea1a4e300124035e031224e8054cfe3f72a8a

PR #3889, opened 2026-04-30
Remove ApacheDS dependencies

duanemay:
* api-ldap-model:2.1.7 (actively used in production) has zero org.apache.directory.server transitive dependencies — the global excludes are not suppressing anything real
* apacheds-protocol-ldap is declared as testImplementation in uaa/build.gradle and simultaneously excluded by configurations.configureEach in build.gradle — a direct self-contradiction; it can never be resolved
* The global exclude lines for both apacheds-core and apacheds-protocol-ldap are themselves dead — nothing in the current dependency graph pulls them in


strehle (approved):
we never use LDAP thus ok

PR #3901, opened 2026-05-06
Migrate MediaType.sortByQualityValue

strehle:
MediaType.sortByQualityValue() was removed in Spring 7. Rather than
reimplementing the sorting logic locally, simplify the two exception
renderers to always write JSON responses directly. UAA is a JSON-only
API — the Accept-header content negotiation was legacy code from
Spring Security OAuth2 that never selected a non-JSON converter in
practice.

This removes the need for a custom MediaTypeUtils utility class and
200+ lines of reimplemented Spring internals. Clients sending non-JSON
Accept headers (e.g. text/html) now receive the JSON error body with
the correct HTTP status instead of a 406 with no body.

Alternative PR to https://github.com/cloudfoundry/uaa/pull/3856


strehle:
close it for now and use the other PR

PR #3904, opened 2026-05-06
refactor: allow multiple UaaTokenEnhancers

peterhaochen47:
- previously, only one is allowed
- add setUaaTokenEnhancers function to allow adding multiple UaaTokenEnhancers (this func will be implicitly called
when UaaTokenEnhancer bean/beans are supplied)
- refactor setUaaTokenEnhancer function (but keep same general business logic as before) for backward compatibility; mark as deprecated

---
incorporated feedback from earlier PR: https://github.com/cloudfoundry/uaa/pull/3899

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java
@@ -90,6 +90,7 @@ import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
@@ -195,7 +196,13 @@ public class UaaTokenServices implements AuthorizationServerTokenServices, Resou
 
     @Autowired(required = false)
     public void setUaaTokenEnhancers(List<UaaTokenEnhancer> uaaTokenEnhancers) {
-        this.uaaTokenEnhancers = new ArrayList<>(uaaTokenEnhancers == null ? emptyList() : uaaTokenEnhancers);
+        if (uaaTokenEnhancers != null) {
+            this.uaaTokenEnhancers = uaaTokenEnhancers.stream()
+                    .filter(Objects::nonNull)
+                    .toList();
+        } else {
+            this.uaaTokenEnhancers = new ArrayList<>();
+        }
     }
 
     @Deprecated
@@ -635,11 +642,9 @@ public class UaaTokenServices implements AuthorizationServerTokenServices, Resou
         if (!uaaTokenEnhancers.isEmpty()) {
             additionalRootClaims = new HashMap<>();
             for (UaaTokenEnhancer enhancer : uaaTokenEnhancers) {
-                if (enhancer != null) {
-                    Map<String, Object> claims = enhancer.enhance(additionalRootClaims, authentication);
-                    if (claims != null) {
-                        additionalRootClaims.putAll(claims);
-                    }
+                Map<String, Object> claims = enhancer.enhance(additionalRootClaims, authentication);
+                if (claims != null) {
+                    additionalRootClaims.putAll(claims);
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java
```diff
@@ -625,8 +634,16 @@ public OAuth2AccessToken createAccessToken(OAuth2Authentication authentication)
...
         Map<String, Object> additionalRootClaims = null;
-        if (uaaTokenEnhancer != null) {
-            additionalRootClaims = new HashMap<>(uaaTokenEnhancer.enhance(emptyMap(), authentication));
+        if (!uaaTokenEnhancers.isEmpty()) {
+            additionalRootClaims = new HashMap<>();
+            for (UaaTokenEnhancer enhancer : uaaTokenEnhancers) {
+                if (enhancer != null) {
+                    Map<String, Object> claims = enhancer.enhance(emptyMap(), authentication);
+                    if (claims != null) {
+                        additionalRootClaims.putAll(claims);
```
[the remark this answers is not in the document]
peterhaochen47: I am not sure merging the claim values from different token enhancers is a good idea; that would result in unpredictable/unintended final outcome. It seems more rigorous that an enhancer has deterministic power of the final outcome of the claim it touches, and if there is a conflict, the later enhancer wins. 

If we do wanna let two enhancers touch on the same claim (not a use case now), we could pass in the previously added claims resulted from prior enhancers to the current enhancer, like an assembly line, so the current enhancer knows what has already been added to what claims (and merge if needed):
```
enhancer.enhance(additionalRootClaims, authentication);
```
But we don't need this for now; we could always add this when we need to. But once we add this, we can't remove it without a breaking change. So I'm holding off on that for now.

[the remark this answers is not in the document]
duanemay: I like this idea, so each successive enhancer (ordered by @Order) can see previous changes and update as needed.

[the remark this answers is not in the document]
peterhaochen47: Yeah, we can add that if you think this is the time to do so. Added a new commit.

peterhaochen47:
@duanemay any other requests for change?

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java
```diff
@@ -625,8 +632,16 @@ public OAuth2AccessToken createAccessToken(OAuth2Authentication authentication)
         boolean isRefreshTokenRevocable = isAccessTokenRevocable || OPAQUE.getStringValue().equals(getActiveTokenPolicy().getRefreshTokenFormat());
 
         Map<String, Object> additionalRootClaims = null;
-        if (uaaTokenEnhancer != null) {
-            additionalRootClaims = new HashMap<>(uaaTokenEnhancer.enhance(emptyMap(), authentication));
+        if (!uaaTokenEnhancers.isEmpty()) {
+            additionalRootClaims = new HashMap<>();
+            for (UaaTokenEnhancer enhancer : uaaTokenEnhancers) {
+                if (enhancer != null) {
```
adrianhoelzl-sap: Should we filter the list for non-null values already in the setter?

uaa/src/test/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServicesTests.java
```diff
@@ -224,6 +224,110 @@ void ensureAnIdTokenIsNotReturned(String grantType) {
...
+                    assertThat(claims).containsEntry("claim2", "value2");
+                } finally {
+                    tokenServices.setUaaTokenEnhancers(new ArrayList<>());
+                }
+            }
+
+            @DisplayName("claims are passed to subsequent enhancers")
+            @ParameterizedTest
+            @ValueSource(strings = {GRANT_TYPE_PASSWORD, GRANT_TYPE_AUTHORIZATION_CODE})
+            void claimsArePassedToSubsequentEnhancers(String grantType) {
```
adrianhoelzl-sap, line 283: Would it make sense to add another test case where `enhancer2` returns a value for `claim1` and thereby overrides the value from `enhancer1`?

peterhaochen47: suggestion adopted

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java
peterhaochen47: suggestion adopted

adrianhoelzl-sap (approved):

PR #3739, opened 2026-02-09
Use entityBaseURL for URLs when generating SAML Metadata

georgikpetrov00:
GitHub issue #3738
Currently, SAML metadata generation derives all URLs (entity ID, ACS, SLO) from the incoming HTTP request domain.

This causes issues when:
- UAA is behind a reverse proxy / load balancer
- multiple external domains are used

### Proposed change

Use the configured `entityBaseURL` (from `login.yml`) as the base for all URLs
generated in the SAML metadata, instead of relying on the request domain.

This allows operators to explicitly define the canonical external URL
used in SAML metadata generation.

### Benefits

- Deterministic and stable SAML metadata URLs
- Correct behavior behind proxies and ingress controllers

### Backward compatibility

If `entityBaseURL` is not configured, the existing behavior
(basing URLs on the request domain) remains unchanged.

PR #3915, opened 2026-05-15
Fix: SAML metadata ACS URL ignores zone subdomain when entityBaseURL is set

fhanik:
### Background — the regression introduced by #3739

PR #3739 ("Use entityBaseURL for URLs when generating SAML Metadata", merged 2026-02-19, commit `4123092`) changed `UaaRelyingPartyRegistrationResolver` to use the operator-configured `login.entityBaseURL` as the base for all SAML SP metadata URLs.

The intent was correct — a static, canonical base URL produces stable metadata behind proxies and load balancers. However, the implementation applied `entityBaseURL` verbatim for **all** zones, including non-default identity zones accessed via subdomain (e.g. `testzone-jnakrm.localhost`). Because `entityBaseURL` is a single static value (`http://localhost:8080/uaa`), the generated metadata for any non-default zone contained the wrong host:

```xml
<!-- Regression: host is localhost, not the zone subdomain host -->
<md:AssertionConsumerService
    Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST"
    Location="http://localhost:8080/uaa/saml/SSO/alias/testzone-jnakrm.integration-saml-entity-id"
    index="0"
    isDefault="true"/>
```

The correct URL for a subdomain-based zone is:

```xml
<md:AssertionConsumerService
    Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST"
    Location="http://testzone-jnakrm.localhost:8080/uaa/saml/SSO/alias/testzone-jnakrm.integration-saml-entity-id"
    index="0"
    isDefault="true"/>
```

### Why the existing test did not catch this

`SamlMetadataEndpointMockMvcTests#nonDefaultZoneSamlMetadataXMLValidation` — the test that directly exercises non-default zone metadata — only asserted with `containsString` on the **path suffix**:

```java
// Before: only checks the alias path, not the full URL — passes even with the wrong host
xpath(".../AssertionConsumerService/@Location")
    .string(containsString("/saml/SSO/alias/testzone-jnakrm.integration-saml-entity-id"))
```

The path suffix was correct (`/saml/SSO/alias/testzone-jnakrm.integration-saml-entity-id`); only the host was wrong (`localhost` vs `testzone-jnakrm.localhost`). Because the assertion never checked the full URL, the test kept passing throughout the regression.

### What this PR fixes

**`UaaRelyingPartyRegistrationResolver`** — adds `resolveBaseUrl` which selects the correct base URL for each access pattern:

| Scenario | Behaviour |
|---|---|
| `entityBaseURL` not set | Derive base URL from the incoming `HttpServletRequest` |
| Default (UAA) zone, `entityBaseURL` set | Use `entityBaseURL` as-is |
| Non-default zone, subdomain access, `entityBaseURL` set | Prepend zone subdomain to `entityBaseURL` host via `UaaUrlUtils.addSubdomainToUrl` |
| Non-default zone, path access (`/z/{subdomain}/`), any `entityBaseURL` | Always derive from request — `entityBaseURL` is a static value that cannot encode the dynamic zone path; `ZonePathContextRewritingFilter` has already embedded the zone in the context path |

**`SamlMetadataEndpointMockMvcTests`** — three improvements:

1. The default-zone ACS Location assertion is tightened from `containsString` to `equalTo` with the full expected URL, so no future host regression can go undetected.
2. `nonDefaultZoneSamlMetadataXMLValidation` now asserts the **full** ACS Location URL (scheme + host + path), not just the path suffix.
3. A new `SamlMetadataZonePathMockMvcTests` nested class covers identity zones accessed via path (`/z/{subdomain}/`). It is placed in a `@Nested @TestPropertySource(properties = {"zones.paths.enabled=true"})` class — the same pattern the other config-variant nested classes use — so the test always runs unconditionally without relying on a Gradle system-property default.

**`UaaRelyingPartyRegistrationResolverTests`** — adds two unconditional unit tests (no `@EnabledIfZonePathsEnabled` gate) covering `resolveBaseUrl` for both zone access patterns:

- `resolveWhenEntityBaseUrlIsSetAndNonDefaultZonePathIgnoresEntityBaseUrlAndUsesRequestContextPath` — uses `ZONE_ORIGINAL_CONTEXT_PATH = ""` and `contextPath = "/z/myzone"`, which is the realistic value the filter sets in a no-servlet-context-path environment (MockMvc / embedded server). This test is the unconditional safety net: it fails immediately if `isZonePathRequest` is broken or removed, regardless of whether `zones.paths.enabled` is set as a system property.
- `resolveWhenEntityBaseUrlIsSetAndNonDefaultZonePathWithServletContextPathIgnoresEntityBaseUrl` — covers the production deployment variant (`contextPath = "/uaa/z/myzone"`, `ZONE_ORIGINAL_CONTEXT_PATH = "/uaa"`).

**`SamlMetadataEndpointKeyRotationTests`** — fixes a bug introduced by #3739 in the test itself: the resolver was constructed with `entityBaseURL = "http://zone-id.localhost:8080/uaa"` — the zone-specific URL with the subdomain already baked in. Because `resolveBaseUrl` now correctly prepends the zone subdomain, this produced a doubled subdomain (`http://zone-id.zone-id.localhost:8080/uaa`). The test is corrected to use `"http://localhost:8080/uaa"` — the UAA instance base URL, consistent with what operators configure in `login.entityBaseURL` and what `integration_test_properties.yml` contains.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/UaaRelyingPartyRegistrationResolver.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/UaaRelyingPartyRegistrationResolver.java
@@ -16,6 +16,10 @@
 package org.cloudfoundry.identity.uaa.provider.saml;
 
 import lombok.extern.slf4j.Slf4j;
+import org.cloudfoundry.identity.uaa.util.UaaUrlUtils;
+import org.cloudfoundry.identity.uaa.zone.IdentityZone;
+import org.cloudfoundry.identity.uaa.zone.IdentityZoneHolder;
+import org.cloudfoundry.identity.uaa.zone.ZonePathContextRewritingFilter;
 import org.springframework.core.convert.converter.Converter;
 import org.springframework.security.saml2.provider.service.registration.RelyingPartyRegistration;
 import org.springframework.security.saml2.provider.service.registration.RelyingPartyRegistrationRepository;
@@ -85,7 +89,7 @@ public final class UaaRelyingPartyRegistrationResolver implements Converter<Http
             if (relyingPartyRegistration == null) {
                 return null;
             } else {
-                String baseUrl = StringUtils.hasText(entityBaseURL) ? entityBaseURL : getApplicationUri(request);
+                String baseUrl = resolveBaseUrl(entityBaseURL, request);
                 Function<String, String> templateResolver = this.templateResolver(baseUrl, relyingPartyRegistration);
                 String relyingPartyEntityId = templateResolver.apply(relyingPartyRegistration.getEntityId());
                 String assertionConsumerServiceLocation = templateResolver.apply(relyingPartyRegistration.getAssertionConsumerServiceLocation());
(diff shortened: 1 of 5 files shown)


fhanik:
@georgikpetrov00 - can you review?

duanemay:
I think we should add a section to UAA-Configuration-Reference.md that has a summary of the table from the initial comment for how entityBaseURL works in different scenarios.

server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/UaaRelyingPartyRegistrationResolver.java
```diff
@@ -152,6 +156,63 @@ private static Map<String, String> constructUriVariables(String baseUrl, Relying
...
+     * </ul>
+     *
+     * When {@code entityBaseURL} is not configured the base URL is always derived from
+     * the incoming HTTP request, which already carries the correct zone information
+     * (subdomain in the {@code Host} header, or zone path in the rewritten context path).
+     */
+    private static String resolveBaseUrl(String entityBaseURL, HttpServletRequest request) {
+        // Path-based zone: ZonePathContextRewritingFilter has embedded /z/{subdomain}
+        // into the context path. entityBaseURL is a static value that cannot reflect this
+        // dynamically, so always derive from the request for this access pattern.
```
georgikpetrov00, line 183: Suggestion: This is already explained in the javadoc above and can be deleted.

server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/UaaRelyingPartyRegistrationResolver.java
```diff
@@ -152,6 +156,63 @@ private static Map<String, String> constructUriVariables(String baseUrl, Relying
...
+        if (isZonePathRequest(request)) {
+            return getApplicationUri(request);
+        }
+
+        if (!StringUtils.hasText(entityBaseURL)) {
+            return getApplicationUri(request);
+        }
+
+        // Subdomain-based zone with entityBaseURL configured: prepend the zone
+        // subdomain to the host so ACS/SLO endpoints resolve to the right virtual host.
```
georgikpetrov00, line 193: this one as well

georgikpetrov00 (approved):
My only thoughts are on the 2 comments mentioned. Everything else looks great.

strehle (approved):

PR #3933, opened 2026-06-04
fix: resolve SAML entity ID from metadata when external_key is null

fhanik:
#### SAML login fails for IDPs bootstrapped from `uaa.yml`

---

### Problem

A SAML IDP configured in `uaa.yml` with a URL-type metadata location (`metadataTrustCheck: true` + `metaDataLocation: https://...`) cannot authenticate users. Two distinct failure modes are observed depending on which ACS URL the IDP targets:

**Failure 1 — `relying_party_registration_not_found`** (primary reported symptom)

When the IDP posts its SAML response to the IDP-alias ACS URL (`/saml/SSO/alias/{idpAlias}`), `UaaRelyingPartyRegistrationResolver.resolveFromRequest` performs an `endsWith` check against the SP entity-ID alias. Because the IDP alias does not end with the SP alias string, the resolver returns `null`, and Spring Security throws:

```
Saml2AuthenticationException{error=[relying_party_registration_not_found]
  No relying party registration found}
```

**Failure 2 — `invalid_issuer`** (secondary; same root cause)

When the IDP posts to the canonical SP ACS URL (`/saml/SSO/alias/{spAlias}`), the issuer is extracted from the SAML response and used to look up the IDP by `external_key`. Because `external_key` is `null` for URL-type bootstrap IDPs (see root cause below), the lookup returns nothing. The default-stub registration is returned instead, and Spring Security throws:

```
Saml2AuthenticationException{error=[invalid_issuer]
  Invalid issuer [https://idp.example.com/...]}
```

---

### Root Cause

`BootstrapSamlIdentityProviderData.setIdentityProviders` only calls `def.setIdpEntityId(...)` for `DATA` (inline XML) metadata type, never for `URL` type. `JdbcIdentityProviderProvisioning` stores `external_key = saml.getIdpEntityId()`, so all URL-type bootstrap IDPs are persisted with `external_key = null`. Both lookup paths that depend on `external_key` therefore silently fail.

---

#### 1. `UaaRelyingPartyRegistrationResolver` — fallback to URL alias as registration ID

When `resolveFromRequest` finds that the URL path segment does not end with the SP alias but a `SAMLResponse` parameter is present, it now falls back to using the URL path segment (the IDP alias) directly as the `relyingPartyRegistrationId`. `ConfiguratorRelyingPartyRegistrationRepository` can then find the IDP by origin key.

```java
// Fallback for IDP-initiated SSO: when the ACS URL carries the IDP alias instead of
// the SP alias, the endsWith check does not fire and relyingPartyRegistrationId stays null.
if (relyingPartyRegistrationId == null && resolvedEntityId != null && samlResponseParameter != null) {
    relyingPartyRegistrationId = resolvedEntityId;
}
```

#### 2. `ConfiguratorRelyingPartyRegistrationRepository` — resolve entity ID from metadata when `external_key` is null

In the loop that matches IDPs by entity ID, if `idpEntityId` (sourced from `external_key`) is `null`, the repository now resolves the entity ID on-the-fly from the stored metadata via `SamlIdentityProviderConfigurator.getExtendedMetadataDelegate`.

```java
String resolvedEntityId = identityProviderDefinition.getIdpEntityId();
if (resolvedEntityId == null) {
    resolvedEntityId = configurator.getExtendedMetadataDelegate(identityProviderDefinition)
            .getAssertingPartyMetadata().getEntityId();
}
if (registrationId.equals(resolvedEntityId)) {
    return createRelyingPartyRegistration(...);
}
```

---

### Changes

- `server/.../saml/UaaRelyingPartyRegistrationResolver.java` — fallback registration ID resolution from URL alias
- `server/.../saml/ConfiguratorRelyingPartyRegistrationRepository.java` — on-the-fly entity ID resolution from metadata when `external_key` is null
- `server/.../saml/Saml2TestUtils.java` — new `responseWithAssertions(issuer, pemKey, pemCert)` overload for signing assertions with custom credentials
- `uaa/.../mock/saml/BootstrapSamlIdpSsoMockMvcTests.java` — new regression test covering both failure modes

---

### Test Plan

- [x] `BootstrapSamlIdpSsoMockMvcTests#samlResponse_viaIdpAliasUrl_authenticatesSuccessfully` — reproduces `relying_party_registration_not_found`; passes with fix in `UaaRelyingPartyRegistrationResolver`
- [x] `BootstrapSamlIdpSsoMockMvcTests#samlResponse_viaSpAliasUrl_authenticatesSuccessfully` — reproduces `invalid_issuer`; passes with fix in `ConfiguratorRelyingPartyRegistrationRepository`
- [ ] Run full `./gradlew integrationTest` to verify no regression in existing SAML flows

Made with [Cursor](https://cursor.com)

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/BootstrapSamlIdentityProviderData.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/BootstrapSamlIdentityProviderData.java
@@ -184,8 +184,22 @@ public class BootstrapSamlIdentityProviderData implements InitializingBean {
 
             IdentityProvider<SamlIdentityProviderDefinition> provider = parseSamlProvider(def);
             if (def.getType() == SamlIdentityProviderDefinition.MetadataLocation.DATA) {
+                // Inline XML: parse is synchronous; let exceptions propagate to fail fast on bad config.
                 RelyingPartyRegistration metadataDelegate = samlConfigurator.getExtendedMetadataDelegate(def);
                 def.setIdpEntityId(metadataDelegate.getAssertingPartyMetadata().getEntityId());
+            } else {
+                // URL-type: remote fetch may fail transiently at startup (IdP unreachable, DNS not
+                // ready, etc.). Log an error so operators know the entity ID could not be stored in
+                // external_key, but continue so the server does not refuse to start.
+                try {
+                    RelyingPartyRegistration metadataDelegate = samlConfigurator.getExtendedMetadataDelegate(def);
+                    def.setIdpEntityId(metadataDelegate.getAssertingPartyMetadata().getEntityId());
+                } catch (Exception e) {
+                    log.error("Could not resolve entity ID for SAML IDP '{}' at startup from '{}'; " +
+                                    "external_key will be null. SAML logins via SP-alias ACS URL may fail " +
+                                    "until the metadata URL becomes reachable: {}",
+                            alias, metaDataLocation, e.getMessage());
+                }
             }
             IdentityProviderWrapper<SamlIdentityProviderDefinition> wrapper = new IdentityProviderWrapper<>(provider);
             wrapper.setOverride(override == null || override);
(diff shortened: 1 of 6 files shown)


fhanik:
We need to decide if we should ingest SAML metadata when bootstrapping SAML IDP providers from uaa.yml

duanemay (dismissed):

duanemay (approved):

PR #3962, opened 2026-06-24
Fix: client secret re-encoding

jochenehret:
Spring Security 7's BCryptPasswordEncoder now extends AbstractValidatingPasswordEncoder, which returns false from matches() when rawPassword is empty. ClientAdminBootstrap calls nonCachingPasswordEncoder.matches("", storedHash) to check whether a client's secret has changed; for clients with an empty secret (e.g. the CF CLI `cf` client) this now always returns false, causing a new BCrypt hash to be written to the DB on every UAA startup.

**Note:** This PR has been AI-generated. It resolves the issue we currently have with the BOSH disaster-recovery-acceptance-tests ("failed to verify token with uaa"):
https://concourse.wg-ard.ci.cloudfoundry.org/teams/main/pipelines/cf-deployment/jobs/bbr-run-drats/builds/2453

I cannot judge however if the fix is conceptually correct. Please review carefully.


duanemay:
I am also out of Copilot quota

strehle:
FYI.
https://github.com/cloudfoundry/uaa/pull/3965

duanemay:
We are going with https://github.com/cloudfoundry/uaa/pull/3965

PR #3964, opened 2026-06-24
feat: implement configurable concurrent session limits per user and client

mshipkovenski:
Introduce numeric concurrent session limits for refresh tokens, replacing the previous binary refreshTokenUnique flag with a numeric cap.

Key changes:

- TokenPolicy: the internal `refreshTokenUnique` field is now an `int` (positive value = max sessions, -1 = unlimited). The original boolean getters/setters (`isRefreshTokenUnique` / `setRefreshTokenUnique`) are preserved with @JsonIgnore for backwards compatibility. New `getMaxSessionLimit` / `setMaxSessionLimit` accessors operate on the integer value. The @JsonSetter accepts boolean or integer input so existing configurations continue to work.

- ClientConstants: adds a REFRESH_TOKEN_UNIQUE constant for the additionalInformation key used for per-client overrides.

- OauthEndpointBeanConfiguration: the `jwt.token.refresh.unique` property now accepts both boolean strings ("true"/"false") and positive integers, parsed at startup.

- UaaTokenServices: replaces the all-or-nothing `deleteRefreshTokensForClientAndUserId` call with `enforceConcurrentSessionLimit`, which revokes only the oldest tokens when the active count would exceed the configured limit. A per-client override (stored in additionalInformation) takes precedence over the zone-level policy. Rotation-aware: token reuse and token replacement are detected to avoid counting a refresh as a new session.

- SampleIdentityZone.json: `refreshTokenUnique` serialises as -1 (integer) instead of false to reflect the new type.

- Tests: DeprecatedUaaTokenServicesTests, RefreshRotationTest, TokenPolicyTest, and the IdentityZone endpoint tests are updated to cover the new behaviour, including per-user isolation, public-client rotation, and confidential-client JTI-reuse scenarios.

### DIFF
--- a/docs/UAA-Configuration-Reference.md
+++ b/docs/UAA-Configuration-Reference.md
@@ -111,7 +111,7 @@ or `$CLOUDFOUNDRY_CONFIG_PATH/uaa.yml`.
 | <a href="#jwttokenpolicyglobalaccesstokenvalidityseconds"><img src="images/click-me.png" width="14" height="14"/></a> `jwt.token.policy.global.accessTokenValiditySeconds` | `43200`| Global access token validity (s)|
 | <a href="#jwttokenpolicyglobalrefreshtokenvalidityseconds"><img src="images/click-me.png" width="14" height="14"/></a> `jwt.token.policy.global.refreshTokenValiditySeconds` | `2592000`| Global refresh token validity (s)|
 | <a href="#jwttokenrefreshformat"><img src="images/click-me.png" width="14" height="14"/></a> `jwt.token.refresh.format` | `opaque`| Refresh token format|
-| <a href="#jwttokenrefreshunique"><img src="images/click-me.png" width="14" height="14"/></a> `jwt.token.refresh.unique` | `false`| Unique refresh tokens|
+| <a href="#jwttokenrefreshunique"><img src="images/click-me.png" width="14" height="14"/></a> `jwt.token.refresh.unique` | `false`| Max concurrent refresh-token sessions per user/client|
 | <a href="#jwttokenrefreshrotate"><img src="images/click-me.png" width="14" height="14"/></a> `jwt.token.refresh.rotate` | `false`| Rotate refresh tokens|
 | <a href="#jwttokenrefreshrestrict_grant"><img src="images/click-me.png" width="14" height="14"/></a> `jwt.token.refresh.restrict_grant` | —| Restrict refresh token grant|
 | <a href="#jwttokenclaimsexclude"><img src="images/click-me.png" width="14" height="14"/></a> `jwt.token.claims.exclude` | `[]`| Claims excluded from tokens|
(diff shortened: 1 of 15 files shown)


fhanik:
Can you please update the configuration reference too: https://github.com/cloudfoundry/uaa/blob/develop/docs/UAA-Configuration-Reference.md#jwttokenrefreshunique

mshipkovenski:
> Can you please update the configuration reference too: https://github.com/cloudfoundry/uaa/blob/develop/docs/UAA-Configuration-Reference.md#jwttokenrefreshunique

Done

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java
```diff
@@ -780,23 +784,21 @@ CompositeToken persistRevocableToken(String tokenId,
...
                     .setIssuedAt(now)
                     .setFormat(isRefreshTokenOpaque ? OPAQUE.getStringValue() : JWT.getStringValue())
                     .setResponseType(REFRESH_TOKEN)
-                    .setZoneId(IdentityZoneHolder.get().getId())
+                    .setZoneId(zoneId)
                     .setUserId(userId)
                     .setScope(scope)
                     .setValue(refreshToken.getValue());
-            if (refreshTokenUnique) {
-                tokenProvisioning.deleteRefreshTokensForClientAndUserId(clientId, userId, IdentityZoneHolder.get().getId());
```
strehle, line 797: this method is now orphan, means no longer in use so you can clean it

strehle (commented):
see here the review from my local AI 

https://github.com/cloudfoundry/uaa/commit/8c60a252d6e16b6a7fd8d8e11c92da0e28eeddd3

Mainly request. check the remove of the no longer needed method to delete refresh token entries

server/src/main/java/org/cloudfoundry/identity/uaa/oauth/UaaTokenServices.java
mshipkovenski: Ack

docs/UAA-Configuration-Reference.md
```diff
@@ -111,7 +111,7 @@ or `$CLOUDFOUNDRY_CONFIG_PATH/uaa.yml`.
 | <a href="#jwttokenpolicyglobalaccesstokenvalidityseconds"><img src="images/click-me.png" width="14" height="14"/></a> `jwt.token.policy.global.accessTokenValiditySeconds` | `43200`| Global access token validity (s)|
 | <a href="#jwttokenpolicyglobalrefreshtokenvalidityseconds"><img src="images/click-me.png" width="14" height="14"/></a> `jwt.token.policy.global.refreshTokenValiditySeconds` | `2592000`| Global refresh token validity (s)|
 | <a href="#jwttokenrefreshformat"><img src="images/click-me.png" width="14" height="14"/></a> `jwt.token.refresh.format` | `opaque`| Refresh token format|
-| <a href="#jwttokenrefreshunique"><img src="images/click-me.png" width="14" height="14"/></a> `jwt.token.refresh.unique` | `false`| Unique refresh tokens|
+| <a href="#jwttokenrefreshunique"><img src="images/click-me.png" width="14" height="14"/></a> `jwt.token.refresh.unique` | `false`| Max concurrent refresh-token sessions per user/client|
```
strehle: fix the false please ,

strehle (commented):
Please make clear in the documentation , that we have changed the format but that we allow the legacy values false and true, but explain the new behaviour

duanemay:
@mshipkovenski can you pull in the commit that @strehle sent 8c60a252d6e16b6a7fd8d8e11c92da0e28eeddd3

mshipkovenski:
> @mshipkovenski can you pull in the commit that @strehle sent [8c60a25](https://github.com/cloudfoundry/uaa/commit/8c60a252d6e16b6a7fd8d8e11c92da0e28eeddd3)

Done

duanemay (approved):

strehle (approved):
thnx

PR #3965, opened 2026-06-25
alternative fix: client secret re-encoding

gdgenchev:
This basically restores pre spring security 7 behaviour across all components that use a caching/non-caching password encoder bean. It reuses the empty aware delegating password encoder that was introduced in spring boot 4 migration, but it was used just in one place.

Alternative for https://github.com/cloudfoundry/uaa/pull/3962


strehle:
thanks, there is already another PR https://github.com/cloudfoundry/uaa/pull/3962

we will handle the fix for it with prio

jochenehret:
Confirmed: This fix resolves the issue. The bbr-run-drats jobs was green:
https://concourse.wg-ard.ci.cloudfoundry.org/teams/main/pipelines/cf-deployment/jobs/bbr-run-drats/builds/2456

strehle (dismissed):

jochenehret:
Is there a regression test for this? For a client with empty secret, the test should verify that the secret hash doesn't change after a uaa restart.

strehle:
@gdgenchev Ok, thank you please fix the tests. we want your PR to get merged :-)

duanemay (dismissed):

gdgenchev:
@duanemay, sick! This was bugging me for hours. I was trying to fix it only test side, but the shadowing `PasswordEncoderConfig` in tests was causing a lot of trouble.

duanemay:
@gdgenchev after a review I can merge

gdgenchev:
I am checking again this `@DefaultTestContext` - seems with the current change, the encoder bean in those tests will not get overridden to the noop one from test config (but the prod one will be used), which might affect performance. I am trying to add the `PasswordEncoderTestConfig` to the `DefaultTestContext`, but I need to remove `@DefaultTestContext` from some `@Nested` test classes as it causes issues. Will try.

strehle (dismissed):

gdgenchev:
I tested and indeed before the rename of the shadowed test class `PasswordEncoderConfig`, it was using the test one and now it uses the prod one (I checked the info log that both bean methods have, seems they are there for this purpose). I am struggling to make this work. Seems indeed unit tests take 22m in this PR and in other 12m, so we might have hit the performance.

gdgenchev:
This is the best I could achieve with smallest change. Unit tests are back to 10 min.

Other approaches caused a lot of issues - i guess it could be possible to somehow rename the shadowing test config class and adapt, but that led to a lot of failing tests.

Thus, decided to keep the shadowing class and just focus on a way to achieve the new tests - and this was possible through gradle configuration and introducing new source set with prod classpath (so the shadowing class in tests does not overwrite the prod one). Seems a bit weird, but it is the simplest and best-working approach I found - now the new regression tests run as part of the integration tests - I have seen them red without the empty aware wrapper.

oliver-heinrich:
Hi @gdgenchev & @strehle the fix is reviewed and tested. Can you merge it and create a new UAA release as the issue is currently blocking the new CF release.

strehle (dismissed):

strehle (dismissed):

strehle (approved):

strehle:
> Hi @gdgenchev & @strehle the fix is reviewed and tested. Can you merge it and create a new UAA release as the issue is currently blocking the new CF release.

New UAA should appear in cf deployment next hours

oliver-heinrich:
UAA release: https://bosh.io/d/github.com/cloudfoundry/uaa-release?v=79.3.0 already passed the update-release pipeline: https://concourse.wg-ard.ci.cloudfoundry.org/teams/main/pipelines/update-releases/jobs/update-uaa/builds/102

Starting at 9 AM today the release will run through the remaining validation pipelines: https://concourse.wg-ard.ci.cloudfoundry.org/teams/main/pipelines/cf-deployment

PR #3968, opened 2026-06-29
feat(spiffe): JWT-SVID signing endpoint (RFC UAA-SPIFFE-001)

rkoster:
## Summary

Adds a **JWT-SVID signing endpoint** to UAA so it can act as a SPIFFE identity
server for Cloud Foundry workloads (offline-OIDC federation).

`POST /jwt-svid/sign` authenticates a SPIFFE Agent (client credentials), verifies
the caller's CF **instance-identity certificate** against the configured CA,
parses org/space/app GUIDs from the cert OUs, checks a **proof-of-possession**
signature for freshness, and mints an RS256 JWT-SVID signed with UAA's active key
and OIDC issuer (so it verifies **offline** against `/token_keys`).

## Key changes (all under `org.cloudfoundry.identity.uaa.spiffe`)
- `JwtSvidController`, `JwtSvidRequest` / `JwtSvidResponse`, `JwtSvidSigner`
- `InstanceIdentityVerifier`, `CertificateOuParser`, `ProofOfPossessionVerifier`
- `SpiffeId`, `CfInstanceIdentity`, `SpiffeProperties`, `SpiffeConfiguration`,
  `SpiffeSecurityConfiguration`
- Full unit-test coverage for each component (9 focused commits).

## Context
Part of **RFC UAA-SPIFFE-001** — *UAA as a SPIFFE Identity Server for Cloud Foundry*.

## Cross-repo dependencies
- **cloudfoundry/uaa-release** `spiffe-signer` — exposes `uaa.spiffe.*` (trust
  domain, instance-identity CA) and the signer client.
- Consumed by the SPIFFE Agent in **cloudfoundry/diego-release** `spiffe-agent`.

## Status
Draft / POC. End-to-end verified on bosh-lite: a CF app received a valid JWT-SVID
with `cf` org/space/app/process_type claims.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/spiffe/CertificateOuParser.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/spiffe/CertificateOuParser.java
@@ -1,0 +1,63 @@
+package org.cloudfoundry.identity.uaa.spiffe;
+
+import org.bouncycastle.asn1.x500.AttributeTypeAndValue;
+import org.bouncycastle.asn1.x500.RDN;
+import org.bouncycastle.asn1.x500.X500Name;
+import org.bouncycastle.asn1.x500.style.BCStyle;
+import org.bouncycastle.asn1.x500.style.IETFUtils;
+import org.bouncycastle.cert.jcajce.JcaX509CertificateHolder;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.stereotype.Component;
+
+import java.security.cert.CertificateEncodingException;
+import java.security.cert.X509Certificate;
+
+/** Extracts CF org/space/app GUIDs from the OU attributes of an instance cert subject. */
+@Component
+@ConditionalOnProperty(prefix = "uaa.spiffe", name = "instance-identity-ca")
+public class CertificateOuParser {
+
+    private static final String ORG_PREFIX = "organization:";
+    private static final String SPACE_PREFIX = "space:";
+    private static final String APP_PREFIX = "app:";
+
+    public CfInstanceIdentity parse(X509Certificate certificate) {
+        String org = null;
+        String space = null;
+        String app = null;
+
+        X500Name subject;
+        try {
+            subject = new JcaX509CertificateHolder(certificate).getSubject();
+        } catch (CertificateEncodingException e) {
+            throw new IllegalArgumentException("Unable to read certificate subject", e);
+        }
+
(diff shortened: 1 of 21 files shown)


strehle:
@rkoster @beyhan which RFC references this feature ? SPIFFE not equal to mTLS, but for us not clear in which context UAA should do what

marcohelmerich:
@strehle @rkoster — a concrete use case from the operator side, in case it helps...

**What we need:** CF apps accessing Google Cloud without service-account keys. Today the keys come from a service broker, sit long-lived in `VCAP_SERVICES`, and have to be rotated by hand. GCP Workload Identity Federation (WIF) solves this if the workload can present a short-lived, signed JWT with a stable subject and an `aud` equal to the WIF provider URI.

**What we tried:**

1. **UAA as the issuer** (per-app client, `client_credentials`). It fails because `aud` can't be controlled: `resource_ids` is always overwritten with `["none"]`, and `aud` is derived from `client_id` plus scopes. As far as we can tell that's by design. It just doesn't fit OIDC federation, where `aud` has to match a pre-registered provider URI exactly.
2. **Per-cell daemon on the host network.** It fails because containers can't reach their cell's IP, which is the intended isolation. And we didnt't want to change that behaviour.
3. **What we found working (PoC):** a small JWT issuer deployed as a CF app and reached over C2C (`apps.internal`). It authenticates callers via mTLS with the Diego instance identity cert and parses the org/space/app GUIDs from the OUs. It signs its own RS256 JWT (`sub=cf-app-<app-guid>`, `aud=<WIF provider>`). The JWKS is uploaded statically to the GCP WIF provider, so GCP never calls back into the foundation. A supply buildpack adds a sidecar that refreshes the token file, plus an `external_account` config, so the Google SDKs pick it up with zero app changes. This works end-to-end on a real foundation.

It works, but it's a second issuer next to UAA, with its own key management, JWKS lifecycle and HA story. This PR would make that redundant: the instance-identity check is the same one we do, and it fixes the one thing that blocked us in (1), a caller-chosen `aud`. With `/jwt-svid/sign` plus the agent's socket, CF could offer GCP WIF out of the box, and likely other OIDC-federating providers too (AWS, Azure, Vault's JWT auth, …).

Happy to test this on our foundations against GCP WIF once there's something deployable.

rkoster:
@marcohelmerich thanks for the feedback. The focus is now on getting https://github.com/cloudfoundry/uaa/pull/3972 across the finish line. Which adds support for [RFC 8705](https://www.rfc-editor.org/info/rfc8705/) to the UAA. This should also allow doing what you are describing.

PR #3971, opened 2026-07-02
Fix: skipSslValidation ignored when fetching SAML IDP metadata during login                              

duanemay:
## Problem

SAML login fails with:

org.springframework.security.saml2.Saml2Exception: Unable to build SAML relying party                    
registration for: <registrationId>                                                                       
Caused by: javax.net.ssl.SSLHandshakeException: (certificate_unknown) PKIX path building                 
failed: unable to find valid certification path to requested target

...for SAML IDPs whose metadata endpoint uses a certificate the JVM's default trust                      
store doesn't trust (self-signed, internal CA, etc.) — **even when the IDP is                            
configured with `skipSslValidation: true`.**

## Root cause

UAA has two separate SAML metadata-fetch code paths:

- **Admin validate path** (`SamlIdentityProviderConfigurator.configureURLMetadata` →                     
  `FixedHttpMetaDataProvider`) — correctly honors the per-IDP `skipSslValidation` flag                   
  and caches results.
- **Live login path** (`ConfiguratorRelyingPartyRegistrationRepository` →                                
  `RelyingPartyRegistrationBuilder` → Spring Security's                                                  
  `RelyingPartyRegistrations.fromMetadataLocation()`) — does a raw `HttpsURLConnection`                  
  fetch on **every SSO redirect**, using only the JVM's default trust store, and                         
  completely ignoring `skipSslValidation`.

This means an IDP can validate/save successfully in the admin UI/API but fail on every                   
actual login.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/ConfiguratorRelyingPartyRegistrationRepository.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/saml/ConfiguratorRelyingPartyRegistrationRepository.java
@@ -89,7 +89,7 @@ public class ConfiguratorRelyingPartyRegistrationRepository extends BaseUaaRelyi
                 .samlEntityID(zonedSamlEntityID)
                 .samlSpNameId(nameID)
                 .keys(keyWithCerts)
-                .metadataLocation(identityProviderDefinition.getMetaDataLocation())
+                .metadataLocation(configurator.resolveMetadataXml(identityProviderDefinition))
                 .rpRegistrationId(registrationId)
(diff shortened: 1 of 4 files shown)


strehle (commented):
this skipssl for SAML metadata was alway a problem to me. In our landsapes we dont allow the automatic update but provision SAML metadata with API ... in the rare use cases. 
We may allow the custom trust achor - as discussed - so do you really want it back ?

duanemay:
I agree with providing the metadata XML approach, however the issue here is URLs are inconsistently wired. With skipSslValidation=true, UAA resolves the entity ID correctly at startup but login attempts fail, because request time doesn't honor the flag.

strehle (approved):
ok

PR #3984, opened 2026-07-16
Add Java 26 to Gradle workflow matrix

duanemay:
Not an LTS, but good compatibility check. Otherwise this standalone test is not adding anything to the already run hsqldb tests

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [ '25' ]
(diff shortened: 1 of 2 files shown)


duanemay:
java.lang.UnsupportedOperationException: Cannot define class using reflection: As of Java 26, using Unsafe is disabled by default, set net.bytebuddy.safe to true if you want to use the JVM's internal unsafe API even though it will become unsupported in the future and should be replaced by injection using method handles: net.bytebuddy.dynamic.loading.ClassInjector$UsingLookup

strehle (approved):

PR #3985, opened 2026-07-17
Fix additionalConfiguration not mapped during YAML IdP bootstrap

duanemay:
## Summary
- `OauthIDPWrapperFactoryBean#setCommonProperties()` maps nearly every field from a YAML-defined OAuth/OIDC identity provider onto the provider definition object (`attributeMappings`, `scopes`, `issuer`, etc.), but silently skipped `additionalConfiguration`.
- Any OAuth/OIDC identity provider bootstrapped from YAML therefore always ends up with `additionalConfiguration == null`, regardless of what the manifest sets. Providers created/updated via the REST API are unaffected, since Jackson deserializes the field correctly there.
- Added the missing one-line mapping, following the exact pattern used for the other fields in this method.

## Test plan
- [x] Added `additionalConfigurationIsMapped` — verifies a YAML map containing `additionalConfiguration` is reflected on the definition object after `setCommonProperties`.
- [x] Added `additionalConfigurationIsNullWhenNotProvided` — verifies null passthrough when the key is absent.
- [x] Ran `OauthIdentityProviderDefinitionFactoryBeanTest` locally — all 26 tests pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


selzoc (approved):
Thanks!

strehle (approved):

PR #4000, opened 2026-07-25
Fix SSRF via jwks_uri

strehle:
Problem: UAA allows operators and clients to configure a jwks_uri pointing to an external JWKS endpoint (used for private_key_jwt client authentication). There was no enforcement preventing a jwks_uri from resolving to internal/private network addresses, making UAA a vector for Server-Side Request Forgery (SSRF) and DNS rebinding
  attacks — an attacker could point the URI at cloud instance-metadata endpoints (e.g. 169.254.169.254), internal services, or loopback addresses.

  Fix — two layers of defense:

  1. Validation-time check (PrivateNetworkGuard): When a client registers or updates a jwks_uri, ClientJwtConfiguration now resolves the hostname at validation time and rejects any URI whose host resolves to a private, loopback, link-local, multicast, or cloud-metadata address (including 169.254.169.254). Unresolvable hosts are allowed through at validation time — the fetch will simply fail later.
  2. Fetch-time check (PrivateNetworkBlockingDnsResolver + safeRestTemplate): A new PrivateNetworkBlockingDnsResolver wraps the system DNS resolver and blocks private addresses at connection time. UaaHttpRequestUtils.createSafeRequestFactory builds a dedicated RestTemplate using this resolver. OidcMetadataFetcher uses this safeRestTemplate when fetching JWKS content for client JWT authentication, providing defense-in-depth against DNS rebinding (where a hostname resolves to a public IP at validation time but rebinds to a private IP at fetch time).

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/OidcMetadataFetcher.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/OidcMetadataFetcher.java
@@ -83,7 +83,9 @@ public class OidcMetadataFetcher {
         if (clientJwtConfiguration.getJwkSet() != null) {
             return clientJwtConfiguration.getJwkSet();
         } else if (clientJwtConfiguration.getJwksUri() != null) {
-            byte[] rawContents = getJsonBody(clientJwtConfiguration.getJwksUri(), false, true, null, safeRestTemplate);
+            String jwksUri = clientJwtConfiguration.getJwksUri();
+            RestTemplate template = isLocalhost(jwksUri) ? nonTrustingRestTemplate : safeRestTemplate;
+            byte[] rawContents = getJsonBody(jwksUri, false, true, null, template);
             if (rawContents != null && rawContents.length > 0) {
                 ClientJwtConfiguration clientKeys = ClientJwtConfiguration.parse(null, new String(rawContents, StandardCharsets.UTF_8));
                 if (clientKeys != null && clientKeys.getJwkSet() != null) {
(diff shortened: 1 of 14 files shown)


server/src/main/java/org/cloudfoundry/identity/uaa/provider/oauth/OidcMetadataFetcher.java
```diff
@@ -130,12 +143,8 @@ private String getClientAuthHeader(AbstractExternalOAuthIdentityProviderDefiniti
     }
 
     private OidcMetadata fetchMetadata(URL discoveryUrl, boolean shouldDoSslValidation) throws OidcMetadataFetchingException {
-        byte[] rawContents;
-        if (shouldDoSslValidation) {
-            rawContents = contentCache.getUrlContent(discoveryUrl.toString(), trustingRestTemplate);
-        } else {
-            rawContents = contentCache.getUrlContent(discoveryUrl.toString(), nonTrustingRestTemplate);
-        }
+        RestTemplate restTemplate = shouldDoSslValidation ? trustingRestTemplate : nonTrustingRestTemplate;
```
duanemay, line 146: Does the pattern like fetchWebKeySet, apply for metadata too? nonTrustingRestTemplate or safeRestTemplate

strehle: no :-) not yet,... but the finding was only for jwks_uri.... but yes both have the same issue, means... they fetch keys in order to vlidate JWT, thus this fetch should not allow to break out...

duanemay:
Should this have a flag for cases where we might need to use a private network?

strehle:
> Should this have a flag for cases where we might need to use a private network?

From standard the jwks_uri and tokenKeyUrl is public... in case the address is not reachable, because the system is no real cloud system , then you can also configure always the JWKSet manually... the URI is mainly for public cloud systems to allow the automatic trust update... but then there should no option to break out -  into internals of the cloud provider ...

duanemay (approved):

PR #4001, opened 2026-07-25
fix: block spoofable IP headers (X-Client-IP, X-Real-IP) from rate limiting

strehle:
fix: block spoofable IP headers (X-Client-IP, X-Real-IP) from rate limiting

Add X-Client-IP and X-Real-IP to the default filteredHeaders list so these headers are stripped from incoming requests before rate-limiting logic reads the client IP, preventing spoofing via untrusted proxy headers.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/SpringServletXmlSecurityConfiguration.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/SpringServletXmlSecurityConfiguration.java
@@ -201,6 +201,8 @@ public class SpringServletXmlSecurityConfiguration {
         bean.setIgnore(Arrays.asList("secFilterOpen05Healthz"));
         int filterPos = 0;
         Map<SecurityFilterChainPostProcessor.FilterPosition, Filter> additionalFilters = new LinkedHashMap<>();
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/cloudfoundry/identity/uaa/SpringServletXmlSecurityConfiguration.java
```diff
@@ -201,12 +201,12 @@ SecurityFilterChainPostProcessor securityFilterChainPostProcessor(
         bean.setIgnore(Arrays.asList("secFilterOpen05Healthz"));
         int filterPos = 0;
         Map<SecurityFilterChainPostProcessor.FilterPosition, Filter> additionalFilters = new LinkedHashMap<>();
+        additionalFilters.put(SecurityFilterChainPostProcessor.FilterPosition.position(filterPos++), headerFilter.getFilter());
```
duanemay, line 206: comment or test explaining why headerFilter must run before rateLimitingFilter

strehle: done

duanemay (approved):

PR #4061, opened 2026-08-31
Don't bootstrap a dummy LDAP identity provider when unconfigured

### LINKED ISSUE #4062: Bootstrap unconditionally creates dummy identity providers
## Problem

`IdentityProviderBootstrap.addLdapProvider()` creates an `identity_provider` row for
origin `ldap` in the `uaa` zone on every startup, even when no real/functioning `ldap:` config is
supplied. That row's mere existence collides with the unique `(identity_zone_id, origin_key)` index,
so a real LDAP identity provider can never be created fresh via `POST /identity-providers` - only
`PUT` against the existing dummy row is possible.

Separately, `ldap`, `login-server`, `keystone`, and `uaa` are getting an unconditional
placeholder row from the 2014 migration `V2_0_2__BootstrapIdentityZones`, regardless
of whether that provider type is ever configured.

## Reproduce

1. Start UAA with the `ldap` profile active and no `ldap:` config (or only
   `ldap.override: false`, which doesn't help - it only affects updates to an
   already-configured row, not creation).
2. `GET /identity-providers` shows an `ldap`-origin row: `active: false`, empty
   config.
3. `POST /identity-providers` with `type=ldap` fails with `idp_exists` - only
   `PUT` against the dummy row works.

## Impact
Blocks provisioning a new real, LDAP identity provider via the REST API for the `uaa` zone. Pollutes the configuration with unnecessary records.

egorpak528:
## Problem

`IdentityProviderBootstrap.addLdapProvider()` unconditionally creates an `IdentityProvider` row with `origin_key="ldap"` in each zone on every startup, regardless of whether real LDAP configuration is supplied. Only the `active` flag depends on real config (`ldapProfile && json.isConfigured()`) - the row itself is always persisted.

Since `identity_provider` enforces a unique `(identity_zone_id, origin_key)` index, this causes two problems for anyone trying to manage LDAP identity providers via the REST API:

1. `POST /identity-providers` with `type=ldap` collides with the existing dummy row - only `PUT` against it is possible, so a real, independently-managed LDAP IDP can never be created fresh.
2. A `ldap:` config supplied with no real connection details (e.g. only `override: false`) does **not** prevent the dummy row from being created on a fresh zone/install - `override` only controls whether an *already-existing* row gets updated on a later boot, not whether one gets created in the first place. So `ldap.override: false` is not a usable workaround.

There's a second, deeper layer to this: even with the app-level fix below, `identity_provider` rows for `uaa`/`login-server`/`ldap`/`keystone` are *also* unconditionally inserted by a much older migration, `V2_0_2__BootstrapIdentityZones` (2014) - on *every* database, fresh or upgraded, since Flyway replays the full migration history regardless. That insert predates any application code and can't be prevented by fixing `IdentityProviderBootstrap` alone.

## Fix

Fixes #4062

Two parts, addressing both layers:

**1. `IdentityProviderBootstrap.addLdapProvider()`** - skip creating the bootstrap provider when there's no real configuration (`LdapIdentityProviderDefinition.isConfigured()`, i.e. no `baseUrl` - so a `ldap:` block with only control flags like `override` still counts as unconfigured) **and** no provider already exists for this zone. A provider that already exists - from a prior real config, the REST API, or the legacy Flyway migration - is left untouched; it's only new, never-configured zones that no longer get a dummy row on boot.

**2. A new migration, `V4_114__DeleteUnconfiguredBootstrapIdentityProviders`** - deletes the `login-server`/`ldap`/`keystone` rows `V2_0_2__BootstrapIdentityZones` left behind, but only if they're still genuinely unconfigured (`config IS NULL`, or for `ldap` specifically, a parsed config with no `baseUrl`). `uaa` is never touched - it's the mandatory internal auth source. Because Flyway replays every migration on every database, this closes the loop for both timeframes at once: a fresh install inserts-then-immediately-deletes these rows within the same initial `migrate()` run (never persisting), while an already-provisioned database gets them cleaned up retroactively on its next upgrade. A real, actively-configured provider is left alone either way.

Verified no blast radius concerns before adding the migration: no FK constraints anywhere reference `identity_provider(id)` (the columns that once did were dropped back in `V2_0_4`), and the two real callers that look up these origins already handle total absence gracefully (`DynamicZoneAwareAuthenticationManager#getProvider()` falls back to a synthetic inactive stub; so does `InvitationsController#acceptLdapInvitation()`).

## Testing

- `IdentityProviderBootstrapTest` - added tests for the profile-alone and override-only-config cases; updated `upgradeLDAPProvider` and `getGenericLdapConfig()` (pre-existing helpers whose assumptions no longer held once dummy rows stopped being free - see commit messages for details).
- `DeleteUnconfiguredBootstrapIdentityProvidersTest` - new, covers null-config deletion, semantic ldap/keystone "still unconfigured" detection, preservation of real configs, and defensive handling of unparseable JSON.
- `TestUtils.cleanAndSeedDb()` updated to match the new post-migration baseline; re-verified against all 10 test files that depend on it (160 tests).

ai-assisted=yes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/server/src/main/java/org/cloudfoundry/identity/uaa/db/DeleteUnconfiguredBootstrapIdentityProviders.java
+++ b/server/src/main/java/org/cloudfoundry/identity/uaa/db/DeleteUnconfiguredBootstrapIdentityProviders.java
@@ -1,0 +1,95 @@
+package org.cloudfoundry.identity.uaa.db;
+
+import org.cloudfoundry.identity.uaa.constants.OriginKeys;
+import org.cloudfoundry.identity.uaa.provider.KeystoneIdentityProviderDefinition;
+import org.cloudfoundry.identity.uaa.provider.LdapIdentityProviderDefinition;
+import org.cloudfoundry.identity.uaa.util.JsonUtils;
+import org.cloudfoundry.identity.uaa.zone.IdentityZone;
+import org.flywaydb.core.api.migration.BaseJavaMigration;
+import org.flywaydb.core.api.migration.Context;
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.jdbc.datasource.SingleConnectionDataSource;
+import org.springframework.util.StringUtils;
+
+import java.util.List;
+import java.util.Map;
+import java.util.function.Predicate;
+
+/**
+ * V2.0.2 ({@link BootstrapIdentityZones}) unconditionally inserted a placeholder
+ * {@code identity_provider} row for each of "uaa", "login-server", "ldap", and
+ * "keystone" in the {@code uaa} zone, with {@code config = null} - regardless
+ * of whether anything was ever actually configured for those origins. That
+ * row's mere existence blocks provisioning a real identity provider for that
+ * origin via {@code POST /identity-providers} (unique
+ * {@code (identity_zone_id, origin_key)} index), and serves no operational
+ * purpose if nothing has ever configured it.
+ * <p>
+ * This migration removes exactly those rows that are STILL in that
+ * never-configured state. It never touches a row that has ever had real
+ * configuration applied to it - whether by {@code IdentityProviderBootstrap},
+ * the REST API, or by hand. "ldap" and "keystone" rows can carry a non-null
+ * but still-empty JSON config (bootstrap code overwrites the literal
+ * {@code NULL} with an unconfigured definition on every boot), so this checks
+ * the parsed content, not just {@code config IS NULL}.
+ * <p>
+ * "uaa" is never touched: it's the mandatory internal auth source and is kept
+ * actively managed by {@code IdentityProviderBootstrap#updateDefaultZoneUaaIDP}.
+ * "login-server" has no bootstrap code that ever populates its config, so a
+ * plain {@code NULL} check is sufficient there.
+ */
+public class DeleteUnconfiguredBootstrapIdentityProviders extends BaseJavaMigration {
+
+    @Override
+    public void migrate(Context context) {
+        JdbcTemplate jdbcTemplate = new JdbcTemplate(new SingleConnectionDataSource(context.getConnection(), true));
+        String uaaZoneId = IdentityZone.getUaaZoneId();
+
+        deleteIfConfigIsNull(jdbcTemplate, uaaZoneId, OriginKeys.LOGIN_SERVER);
+        deleteIfUnconfigured(jdbcTemplate, uaaZoneId, OriginKeys.KEYSTONE,
+                DeleteUnconfiguredBootstrapIdentityProviders::keystoneIsUnconfigured);
+        deleteIfUnconfigured(jdbcTemplate, uaaZoneId, OriginKeys.LDAP,
+                DeleteUnconfiguredBootstrapIdentityProviders::ldapIsUnconfigured);
+    }
+
+    private void deleteIfConfigIsNull(JdbcTemplate jdbcTemplate, String zoneId, String origin) {
+        jdbcTemplate.update(
+                "delete from identity_provider where identity_zone_id = ? and origin_key = ? and config is null",
+                zoneId, origin);
+    }
+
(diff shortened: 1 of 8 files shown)


server/src/main/java/org/cloudfoundry/identity/uaa/impl/config/IdentityProviderBootstrap.java
```diff
@@ -140,6 +140,23 @@ protected void addLdapProvider() {
...
+          bootstrap: either real connection details were actually supplied (baseUrl set -
+          see LdapIdentityProviderDefinition.isConfigured()), or a provider already exists
+          for this zone (e.g. from a prior real config, the REST API, or the legacy Flyway
+          migration that used to insert one unconditionally). A `ldap:` block containing
+          only control flags like `override: false` - with no baseUrl/bind/search settings
+          - is not "configured" and must be treated the same as no `ldap:` block at all.
+          Merely activating the `ldap` Spring profile (with or without such a block) must
+          not create a dummy, unconfigured identity provider - that row blocks a real LDAP
+          IDP from being created via the REST API (unique origin/zone constraint) and has
+          no operational purpose. See TNZ-125736.
```
gdgenchev: `See TNZ-125736.` - replace with github issue (create one). Check also other places where this is written in comments (`TestUtils` and `IndentityProviderBootstrapTest`)

server/src/main/java/org/cloudfoundry/identity/uaa/db/DeleteUnconfiguredBootstrapIdentityProviders.java
```diff
@@ -0,0 +1,95 @@
...
+                zoneId, origin);
+        for (Map<String, Object> row : rows) {
+            String config = (String) row.get("config");
+            if (config == null || isUnconfigured.test(config)) {
+                jdbcTemplate.update("delete from identity_provider where id = ?", row.get("id"));
+            }
+        }
+    }
+
+    private static boolean ldapIsUnconfigured(String config) {
```
gdgenchev: applies to both predicate methods:

1. static not needed
2. exception can be logged as warning

egorpak528: Thanks, addressed in 6f659a3

server/src/main/java/org/cloudfoundry/identity/uaa/impl/config/IdentityProviderBootstrap.java
egorpak528: Thanks, addressed in 2101828.

server/src/main/java/org/cloudfoundry/identity/uaa/db/DeleteUnconfiguredBootstrapIdentityProviders.java
```diff
@@ -0,0 +1,95 @@
...
+import org.cloudfoundry.identity.uaa.provider.LdapIdentityProviderDefinition;
+import org.cloudfoundry.identity.uaa.util.JsonUtils;
+import org.cloudfoundry.identity.uaa.zone.IdentityZone;
+import org.flywaydb.core.api.migration.BaseJavaMigration;
+import org.flywaydb.core.api.migration.Context;
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.jdbc.datasource.SingleConnectionDataSource;
+import org.springframework.util.StringUtils;
+
+import java.util.List;
```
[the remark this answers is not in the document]
egorpak528: Fixed in [6f659a3](https://github.com/cloudfoundry/uaa/pull/4061/commits/6f659a3bf0cdff281c375df9665e2c2a94708e93)

egorpak528:
Thanks for the prompt review, gdgenchev!

adrianhoelzl-sap (approved):

PR #4065, opened 2026-09-04
Pass through unrecognized claims in check_token and introspect responses

duanemay:
## Summary
- `CheckTokenEndpoint` and `IntrospectEndpoint` both decode a JWT into a `Claims`/`IntrospectionClaims` POJO annotated `@JsonIgnoreProperties(ignoreUnknown = true)`, so any top-level claim without a matching field is silently dropped from the response.
- `Claims` captures unrecognized claims via `@JsonAnySetter` and re-flattens them back to top level on serialization via `@JsonAnyGetter`, instead of discarding them.
- RFC 7662 (token introspection) explicitly allows this — section 2.2 says the response "MAY" include extension
  members beyond the standard set, and clients "SHOULD" ignore unrecognized ones. 
- Since the catch-all lives on the shared `Claims` class, this change applies to **both** `/check_token` and `/introspect` — any claim added to a token (e.g. by a custom `UaaTokenEnhancer`) that isn't one of the explicitly modeled fields now shows up in both endpoints' responses.

## Test plan
- [x] `IntrospectionClaimsTest` — verifies an unrecognized claim round-trips through deserialize/serialize and appears at the top level of the JSON output.
- [x] `CheckTokenEndpointTests#additionalClaimsInResult` — verifies custom top-level claims added by a `UaaTokenEnhancer` (`ex_groups`, `ex_prop`) are no longer dropped from the `/check_token` response.
- [x] Existing `IntrospectionClaimsTest`, `IntrospectEndpointTest`, and `CheckTokenEndpointTests` suites pass unchanged.

### DIFF
--- a/model/src/main/java/org/cloudfoundry/identity/uaa/oauth/token/Claims.java
+++ b/model/src/main/java/org/cloudfoundry/identity/uaa/oauth/token/Claims.java
@@ -14,6 +14,8 @@
 
 package org.cloudfoundry.identity.uaa.oauth.token;
 
+import com.fasterxml.jackson.annotation.JsonAnyGetter;
+import com.fasterxml.jackson.annotation.JsonAnySetter;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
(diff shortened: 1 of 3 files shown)


fhanik (dismissed):

model/src/main/java/org/cloudfoundry/identity/uaa/oauth/token/Claims.java
```diff
@@ -379,4 +382,14 @@ public Map<String, Object> getActorClaims() {
     public Map<String, Object> getClaimMap() {
         return JsonUtils.convertValue(this, HashMap.class);
     }
+
+    @JsonAnyGetter
+    public Map<String, Object> getAdditionalClaims() {
+        return java.util.Collections.unmodifiableMap(additionalClaims);
```
adrianhoelzl-sap: Can we add an import?

adrianhoelzl-sap (approved):

PR #4080, opened 2026-09-21
introduce @Value wiring for allowPublicClients in ClientAdminBootstrap

robl-sap:
Problem

When ClientAdminBootstrap was converted to a @Component (commit a1459a300), the @Value SpEL annotation was accidentally dropped from the allowPublicClients constructor parameter. Without it, Spring performs type-based injection and injects all String-typed beans from the context into that Set instead of reading oauth.client.allowpublic from the YAML config.

Fix

Restore the @Value expression on the allowPublicClients parameter with null-guards for the case where oauth or oauth.client are absent from the config:

@Value("#{@config['oauth']==null ? null : @config['oauth']['client']==null ? null : @config['oauth']['client']['allowpublic']}")

Tests

Added ClientAdminBootstrapWiringTest — a minimal Spring context wiring test that covers three scenarios:

- oauth.client.allowpublic is configured → correct client ID is picked up
- oauth key is absent from config → resolves to empty set (no NPE)
- oauth.client subkey is absent → resolves to empty set (no NPE)


server/src/main/java/org/cloudfoundry/identity/uaa/client/ClientAdminBootstrap.java
```diff
@@ -90,7 +90,7 @@ public class ClientAdminBootstrap implements
             @Value("#{@applicationProperties.containsKey('oauth.client.autoapprove') ? @config['oauth']['client']['autoapprove'] : 'cf'}") final Collection<String> autoApproveClients,
             @Value("#{@config['delete']==null ? null : @config['delete']['clients']}") final Collection<String> clientsToDelete,
             final JdbcTemplate jdbcTemplate,
-            final Set<String> allowPublicClients) {
+            @Value("#{@config['oauth']==null ? null : @config['oauth']['client']==null ? null : @config['oauth']['client']['allowpublic']}") final Set<String> allowPublicClients) {
```
duanemay: Can we make this more like the other case above?

`@Value("#{@applicationProperties.containsKey('oauth.client.autoapprove') ? @config['oauth']['client']['autoapprove'] : 'cf'}")`

strehle:
please. check unit tests

strehle (approved):

